PR #13, opened 2023-08-15
feat: get node info

okjodom:
implement get node info for lnd

### DIFF
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -24,7 +24,11 @@ async fn main() -> anyhow::Result<()> {
 
     for node in nodes {
         let lnd = LndNode::new(node.address, node.macaroon, node.cert).await?;
-        clients.insert(node.id, Arc::new(lnd));
+
+        let node_info = lnd.get_info().await?;
+        println!("Node info {:?}", node_info);
+
+        clients.insert(node_info.pubkey, Arc::new(lnd));
     }
 
     let sim = Simulation::new(clients, activity);
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -40,11 +40,21 @@ pub struct ActivityDefinition {
 }
 
 #[derive(Debug, Error)]
-pub enum PaymentError {
-    #[error("Get info failed {0}")]
-    GetInfoFailed(String),
+pub enum SimulationError {
+    #[error("Lightning Error: {0:?}")]
+    LightningError(#[from] LightningError),
     #[error("Other: {0:?}")]
     Error(#[from] anyhow::Error),
+}
+
+#[derive(Debug, Error)]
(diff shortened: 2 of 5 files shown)


sim-cli/src/main.rs
```diff
@@ -24,7 +24,12 @@ async fn main() -> anyhow::Result<()> {
 
     for node in nodes {
         let lnd = LndNode::new(node.address, node.macaroon, node.cert).await?;
-        clients.insert(node.id, Arc::new(lnd));
+
+        let node_info = lnd.get_info().await?;
+        println!("Node info {:?}", node_info);
+
+        // TODO: validate node has keysend feature
+        clients.insert(node_info.pubkey, Arc::new(lnd));
```
carlaKC, line 32: We only need this for destinations so I'm not sure it needs to live here?

sim-lib/src/lib.rs
```diff
@@ -35,15 +40,37 @@ pub struct ActivityDefinition {
     pub amt_msat: u64,
 }
 
-// Phase 2: Event Queue
+#[derive(Debug, Error)]
+pub enum PaymentError {
```
carlaKC, line 44: I think we should just change this to `Error`

sim-lib/src/lib.rs
```diff
@@ -35,15 +40,37 @@ pub struct ActivityDefinition {
...
+}
 
-#[allow(dead_code)]
-pub enum PaymentError {}
+#[derive(Debug, Clone)]
+pub struct NodeInfo {
+    pub pubkey: PublicKey,
+    pub alias: String,
+    // TODO: This should be generic for any node impl
+    pub features: HashMap<u32, Feature>,
```
carlaKC, line 56: I think this will fit most impls, since `Feature` is just a number?

carlaKC: Does this need to map to feature or just a map where we can lookup the u32?

sim-cli/src/main.rs
okjodom: move to validation

sim-lib/src/lib.rs
okjodom: I suppose a vec of the `u32`s from get info is sufficient. And we'd have to get this for the `destination` nodes rather than the `origin`s?

carlaKC (commented):
LGTM pending CI fixing up! 
Are you going to take over #10 in this PR?

sim-lib/src/lib.rs
okjodom: fleshed out error def more so we can report ln errors better

okjodom:
> LGTM pending CI fixing up! Are you going to take over #10 in this PR?

Picking it up as a follow up to this PR

carlaKC (approved):

PR #15, opened 2023-08-16
feat: track payment

okjodom:


sim-lib/src/lnd.rs
```diff
@@ -60,7 +62,27 @@ impl LightningNode for LndNode {
...
-    async fn track_payment(&self, _hash: PaymentHash) -> Result<(), LightningError> {
-        unimplemented!()
+    async fn track_payment(&self, hash: PaymentHash) -> Result<PaymentOutcome, LightningError> {
+        let mut client = self.client.clone();
+        let router_client = client.router();
+
+        let response = router_client
+            .track_payment_v2(TrackPaymentRequest {
+                payment_hash: hash.0.to_vec(),
+                no_inflight_updates: false,
```
carlaKC, line 72: I think we want this to be true, in-flight can be a little spammy and we generally only care about success/failure here (for now).

sim-lib/src/lnd.rs
```diff
@@ -60,7 +62,27 @@ impl LightningNode for LndNode {
...
+            })
+            .await?;
+
+        let mut stream = response.into_inner();
+
+        while let Some(payment) = stream.message().await? {
+            info!("Payment: {payment:?}");
+            if payment.status == 2 {
```
carlaKC, line 80: Does `tonic_lnd` define an enum for the status? Would make this a little more readable

sim-lib/src/lnd.rs
```diff
@@ -60,7 +62,27 @@ impl LightningNode for LndNode {
         unimplemented!()
     }
 
-    async fn track_payment(&self, _hash: PaymentHash) -> Result<(), LightningError> {
-        unimplemented!()
+    async fn track_payment(&self, hash: PaymentHash) -> Result<PaymentOutcome, LightningError> {
```
carlaKC, line 65: Thinking about how we're going to use this, do you think it would make sense to pass a receiving channel in here which we report the payment outcome to? 

That way we can: 
1. Send the payment
2. Spin up a thread to track its outcome
3. Report the outcome to a single monitoring thread that writes them wherever we need them

carlaKC (commented):
Looking good, just one question about the API we want here.

sim-lib/src/lnd.rs
sr-gi: Did you mean false @carlaKC? Otherwise this hasn't change

sim-lib/src/lnd.rs
```diff
@@ -101,12 +104,72 @@ impl LightningNode for LndNode {
...
+                payment_hash: hash.0.to_vec(),
+                no_inflight_updates: false,
+            })
+            .await
+        {
+            Ok(response) => response,
+            Err(err) => {
+                return Err(LightningError::TrackPaymentError(err.to_string()));
+            }
+        };
```
sr-gi, line 125:
```suggestion
        let response = self
            .client
            .router()
            .track_payment_v2(TrackPaymentRequest {
                payment_hash: hash.0.to_vec(),
                no_inflight_updates: false,
            })
            .await
            .map_err(|e| LightningError::TrackPaymentError(e.to_string()))?;
```

sim-lib/src/lnd.rs
```diff
@@ -101,12 +104,72 @@ impl LightningNode for LndNode {
...
+                let payment  = match stream {
+                    Ok(payment) => payment,
+                    Err(err) => {
+                        return Err(LightningError::TrackPaymentError(err.to_string()));
+                    }
+                };
+
+                match payment {
+                    Some(payment) => {
+
```
sr-gi, line 141: nit: remove empty line

sim-lib/src/lnd.rs
```diff
@@ -101,12 +104,72 @@ impl LightningNode for LndNode {
...
+                    },
+                }
+            },
+            _ = shutdown => { Err(LightningError::TrackPaymentError("Shutdown before tracking results".to_string())) }
+        }
     }
 }
 
-fn string_to_payment_hash(hash: &str) -> Result<PaymentHash, LightningError> {
+pub fn string_to_payment_hash(hash: &str) -> Result<PaymentHash, LightningError> {
```
sr-gi, line 172: Why does this need to be pub?

sim-lib/src/lnd.rs
```diff
@@ -101,12 +104,72 @@ impl LightningNode for LndNode {
...
+                            3 => Some(PaymentFailureReason::UnexpectedError),
+                            4 => Some(PaymentFailureReason::UnexpectedError), // TODO: Capture failure reasons for incorrect payment
+                            5 => Some(PaymentFailureReason::UnexpectedError), // TODO: Capture failure reasons for insufficient balance
+                            _ => None,
+                        };
+                        return Ok(PaymentOutcome::Resolved(PaymentResult {
+                            settled,
+                            htlc_count,
+                            failure_reason,
+                        }));
```
sr-gi, line 158: I think all of this can be flattened to the `PaymentResult` constructor. There is no need of temp variables

sim-lib/src/lib.rs
```diff
@@ -95,16 +100,17 @@ enum NodeAction {
     SendPayment(PublicKey, u64),
 }
 
-// Phase 3: CSV output
+#[derive(Debug, Clone)]
+pub enum PaymentOutcome {
+    Failed(PaymentFailureReason),
```
sr-gi, line 105: This is not being used

sr-gi (changes requested):
Some nits, code formatting, and simplifications.

Looks good functionality-wise

sim-lib/src/lnd.rs
okjodom: should be `false`

sim-lib/src/lnd.rs
okjodom: opted to skip streaming the result through a channel and instead refer a direct `PaymentResult`

sim-lib/src/lib.rs
okjodom: removed!

sim-lib/src/lnd.rs
okjodom: I was using it from the cli to test track_payment. doesn't need to be pub

sim-lib/src/lnd.rs
```diff
@@ -101,8 +104,51 @@ impl LightningNode for LndNode {
...
+        let mut stream = response.into_inner();
+
+        tokio::select! {
+            biased;
+            stream = stream.message() => {
+                let payment = stream.map_err(|err| LightningError::TrackPaymentError(err.to_string()))?;
+                match payment {
+                    Some(payment) => {
+                        return Ok(PaymentResult {
+                            settled: payment.status == 2 || payment.status == 3,
```
sr-gi, line 131: What about something like this?

```
settled: PaymentStatus::from_i32(payment.status).unwrap() == PaymentStatus::Succeeded,
```

sr-gi (approved):
LGTM

sim-lib/src/lnd.rs
okjodom: works!

PR #19, opened 2023-08-16
send payment

okjodom:


sim-lib/src/lnd.rs
```diff
@@ -54,13 +60,61 @@ impl LightningNode for LndNode {
...
+            .try_into()
+            .map_err(|_| LightningError::SendPaymentError("Invalid send amount".to_string()))?;
+
+        let preimage = PaymentPreimage(rand::random());
+
+        let mut dest_custom_records = HashMap::new();
+        dest_custom_records.insert(KEYSEND_KEY, preimage.0.to_vec());
+
+        let response = router_client
+            .send_payment_v2(SendPaymentRequest {
```
carlaKC, line 79: Needs timeout (5 mins?) + hash of `preimage`

okjodom: fixed

carlaKC (approved):
ackackack, picked this onto main and payment are firing 🔥

PR #23, opened 2023-08-16
Split CI jobs into test and formatting

sr-gi:
Also gives priority to formatting checks so CI fails early if needed


okjodom (approved):
lgtm

PR #34, opened 2023-08-17
Reworks LndNode so we persist NodeInfo within it

sr-gi:
This modifies the LightningNode trait with respect to get_info. Now instead of implementing the get_info rpc command on LightningNode::get_info we set that as a getter to NodeInfo, and NodeInfo is built calling the rpc command on LndNode::new.


carlaKC (approved):
LGTM!

PR #35, opened 2023-08-17
Deleted binaries

bigmarh:


carlaKC (changes requested):
This PR has a whole bunch of commits squashed into it - I think it can just be a single commit deleting the binaries.

sr-gi:
> This PR has a whole bunch of commits squashed into it - I think it can just be a single commit deleting the binaries.

Agreed

Issue #31, opened 2023-08-17
nit: Rename frequency to interval for the activities


sr-gi:
Given the use we are giving to "frequencies" parsed from the config file, we may be better off calling them intervals

PR #46, opened 2023-08-17
Replaces frequency for interval, changes the order or wait/execution for produce_events

### LINKED ISSUE #30: Swap the interval waits/action order in produce_events
Currently, we are waiting before performing any actions for `produce_events`. This means that we will start waiting when the simulation is launched. This is somewhat annoying, it'd be better to swap the order so the payment flow starts right away.

sr-gi:
Close #30 #31


sim-lib/src/lib.rs
```diff
@@ -37,8 +37,8 @@ pub struct ActivityDefinition {
     pub source: PublicKey,
     // The destination of the action.
     pub destination: PublicKey,
-    // The frequency of the action, as in number of times per minute.
-    pub frequency: u16,
+    // The interval of the action, as in every how many seconds the action is performed.
+    pub interval: u16,
```
carlaKC, line 41: `interval_seconds` in config

carlaKC:
cc: @bigmarh  we're changing `frequency` -> `interval_secs`

sr-gi:
@carlaKC I've added annotations for `interval_secs` and included the `tokio::select` logic

carlaKC (approved):
🔥

PR #64, opened 2023-08-18
hotfix: Fixes select for produce_events

sr-gi:
`tokio::select` for `produce_events` was not working correctly. The way it was structured, `send` will always be ready. Therefore payments were not respecting the intervals set by `time::timeout`.

This was introduced by #46 

#Opsie :/


carlaKC (approved):
tACK ✅

PR #66, opened 2023-08-18
feat: clean shutdown simulation on sigterm

### LINKED ISSUE #58: Capture sigterm and trigger shutdown

okjodom:
capture sigterm from `sim-cli` and issue simulation shutdown. this allows us to complete tasks like report writing, before simulation exit

fixes #58

### DIFF
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -44,7 +44,14 @@ async fn main() -> anyhow::Result<()> {
         clients.insert(node.id, Arc::new(Mutex::new(lnd)));
     }
 
-    let sim = Simulation::new(clients, activity);
+    let sim = Arc::new(Simulation::new(clients, activity));
+    let sim2 = sim.clone();
+
+    ctrlc::set_handler(move || {
+        log::info!("Shutting down simulation...");
+        sim2.shutdown();
+    })?;
+
     sim.run().await?;
 
     Ok(())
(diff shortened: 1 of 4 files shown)


sim-lib/src/lib.rs
```diff
@@ -292,10 +302,11 @@ impl Simulation {
     async fn generate_activity(
         &self,
         executed_actions: Sender<ActionOutcome>,
-        shutdown: Trigger,
-        listener: Listener,
     ) -> Result<JoinSet<()>, SimulationError> {
         let mut set = JoinSet::new();
+        let shutdown = self.shutdown_trigger.clone();
+        let listener = self.shutdown_listener.clone();
```
carlaKC, line 308: These aren't inline with `spawn` because of `self` escaping the borrow, right?

sim-cli/src/main.rs
```diff
@@ -44,7 +44,14 @@ async fn main() -> anyhow::Result<()> {
         clients.insert(node.id, Arc::new(Mutex::new(lnd)));
     }
 
-    let sim = Simulation::new(clients, activity);
+    let sim = Arc::new(Simulation::new(clients, activity));
+    let sim2 = sim.clone();
```
carlaKC, line 48: Not my _favorite_ var name 🙈

carlaKC (approved):
tACK ✅

sim-lib/src/lib.rs
okjodom: correct. we can never move `self` into the spawn

PR #69, opened 2023-08-18
Adds a global timeout

### LINKED ISSUE #38: Feature: Add total runtime for simulation
We've made the decision to express payment frequency in intervals, which means that we have no end time for the simulation. 

We should add an option to the simulation to either run forever or stop after some duration X.
This should be surfaced on the CLI.

sr-gi:
Simplifies the JoinSet management to have a single global set of tasks

Also adds a cli param to control the total simulation runtime (`--total-time`)

Close #38


carlaKC (approved):
Already reviewed code in dependent PR #70, LGTM!

PR #70, opened 2023-08-18
fix: show failure reason in csv report

### LINKED ISSUE #63: Show failure reason in csv report

okjodom:
uses serializable `FailureReason` that we can then show in the generated report
```
source,destination,hash,amount_msat,dispatch_time,settled,htlc_count,failure_reason
02dd7ca90bac8579b7c5aa8b6eece032327b12f14314c59323d9ff9a5b65e1a63c,0378f52c4a103734795040a7d563df24a59c091a1d25d7acb3041a99bde8afbd13,43ebabbb538d9619042a066ad647d6ff80beeb6f5010113d4e87ff75f505f186,10000,1692366457128,true,0,RouteNotFound
02dd7ca90bac8579b7c5aa8b6eece032327b12f14314c59323d9ff9a5b65e1a63c,0378f52c4a103734795040a7d563df24a59c091a1d25d7acb3041a99bde8afbd13,bce62bc0b782904190d9a9dded1aabbb4c652f361531d28b184921b3dac23122,10000,1692366459132,true,0,RouteNotFound
```

fixes #63

### DIFF
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -1,7 +1,6 @@
 use async_trait::async_trait;
 use bitcoin::secp256k1::PublicKey;
 use csv::WriterBuilder;
-use lightning::events::PaymentFailureReason;
 use lightning::ln::PaymentHash;
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
@@ -16,6 +15,7 @@ use tokio::time;
 use triggered::{Listener, Trigger};
 
 pub mod lnd;
+mod serializers;
 
 const KEYSEND_OPTIONAL: u32 = 55;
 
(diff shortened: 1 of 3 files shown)


carlaKC:
Since we're here and no longer under time pressure, I think we should go for the "final" format that we want. Right now we'll produce a settled bool + a failure reason: 
Screenshot 2023-08-21 at 9 32 28 AM

I think that we should replace those two columns with a single `payment_outcome`:
* Successful payments: `Success`.
* Failed payments: Failure reason string as implemented here. 

I know we'll need to add custom serialization for that, but it seems like a much more natural way to represent the data.

carlaKC (approved):
tACK!

PR #71, opened 2023-08-18
Adds debug arg to the CLI

sr-gi:
Debug logs can be turned in using `--debug`.

Also fixes pending logs.

### DIFF
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -14,6 +14,8 @@ use simple_logger::SimpleLogger;
 struct Cli {
     #[clap(long, short)]
     config: PathBuf,
+    #[clap(long, short)]
+    total_time: Option<u32>,
 }
 
 #[tokio::main]
@@ -44,7 +46,7 @@ async fn main() -> anyhow::Result<()> {
         clients.insert(node.id, Arc::new(Mutex::new(lnd)));
     }
 
-    let sim = Arc::new(Simulation::new(clients, activity));
+    let sim = Simulation::new(clients, activity, cli.total_time);
     let sim2 = sim.clone();
 
     ctrlc::set_handler(move || {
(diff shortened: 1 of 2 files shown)


sim-lib/src/lib.rs
```diff
@@ -233,6 +233,7 @@ impl Simulation {
             self.activity.len(),
             self.nodes.len()
         );
+        let mut tasks = JoinSet::new();
```
carlaKC, line 245: noice 🥇

sim-cli/src/main.rs
```diff
@@ -16,19 +16,35 @@ struct Cli {
     config: PathBuf,
     #[clap(long, short)]
     total_time: Option<u32>,
+    #[clap(long, short)]
+    debug: bool,
```
carlaKC, line 20: Bool config option vs passing in a string that allows enabling other values like `trace`?

carlaKC (approved):
LGTM - possibly want to update log config to something like `--log-level=value` so that we can set any logging level but already a big improvement!

sim-cli/src/main.rs
```diff
@@ -14,19 +14,37 @@ use simple_logger::SimpleLogger;
...
     config: PathBuf,
+    #[clap(long, short)]
+    total_time: Option<u32>,
+    #[clap(long, short)]
+    debug: bool,
 }
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    let cli = Cli::parse();
```
okjodom, line 25: would we also want to read in `RUST_LOG` environment at this point?

okjodom (approved):

sim-cli/src/main.rs
sr-gi: I don't think I follow

PR #74, opened 2023-08-23
Sets the basic structure for NodeConnections so other lightning nodes can be parsed from the config file

sr-gi:
Currently we only accept one type of node, LndNode.

This generalizes the type of connection so other nodes can be added, and templates CLN.


sr-gi:
It is worth noting that we've replaced `get_node_info` for `get_node_features`. For LND, we are currently only returning `NodeFeatures` signaling wether keysend is enabled or not. This is a bit nasty, so we should add a method to build `NodeFeatures` to consider every single signaled feature

okjodom:
tACK

okjodom (approved):

PR #75, opened 2023-08-23
feat: track payment on cln nodes

okjodom:
Polls `lightning-listpays` until payment completes or fails

### DIFF
--- a/sim-lib/src/cln.rs
+++ b/sim-lib/src/cln.rs
@@ -1,18 +1,22 @@
 use async_trait::async_trait;
 use bitcoin::secp256k1::PublicKey;
 use cln_grpc::pb::{
-    node_client::NodeClient, Amount, GetinfoRequest, GetinfoResponse, KeysendRequest,
-    KeysendResponse, ListnodesRequest,
+    listpays_pays::ListpaysPaysStatus, node_client::NodeClient, Amount, GetinfoRequest,
+    GetinfoResponse, KeysendRequest, KeysendResponse, ListnodesRequest, ListpaysRequest,
+    ListpaysResponse,
 };
 use lightning::ln::features::NodeFeatures;
 use lightning::ln::PaymentHash;
 
 use tokio::fs::File;
 use tokio::io::{AsyncReadExt, Error};
+use tokio::time::{self, Duration};
 use tonic::transport::{Certificate, Channel, ClientTlsConfig, Identity};
 use triggered::Listener;
 
-use crate::{ClnConnection, LightningError, LightningNode, NodeInfo, PaymentResult};
+use crate::{
+    ClnConnection, LightningError, LightningNode, NodeInfo, PaymentOutcome, PaymentResult,
+};
 
 pub struct ClnNode {
     pub client: NodeClient<Channel>,
@@ -101,10 +105,45 @@ impl LightningNode for ClnNode {
 
     async fn track_payment(
         &mut self,
-        _hash: PaymentHash,
-        _shutdown: Listener,
+        hash: PaymentHash,
+        shutdown: Listener,
     ) -> Result<PaymentResult, LightningError> {
-        unimplemented!()
+        loop {
+            tokio::select! {
+                biased;
+                _ = shutdown.clone() => {
+                    return Err(LightningError::TrackPaymentError("Shutdown before tracking results".to_string()));
+                },
+                _ = time::sleep(Duration::from_millis(500)) => {
+                    let ListpaysResponse { pays } = self
+                        .client
+                        .list_pays(ListpaysRequest {
+                            payment_hash: Some(hash.0.to_vec()),
+                            ..Default::default()
+                        })
+                        .await
+                        .map_err(|err| LightningError::TrackPaymentError(err.to_string()))?
+                        .into_inner();
+
+                    if let Some(pay) = pays.first() {
+                        let payment_status = ListpaysPaysStatus::from_i32(pay.status)
+                            .ok_or(LightningError::TrackPaymentError("Invalid payment status".to_string()))?;
+
+                        let payment_outcome = match payment_status {
+                            ListpaysPaysStatus::Pending => continue,
+                            ListpaysPaysStatus::Complete => PaymentOutcome::Success,
+                            // Task: https://github.com/bitcoin-dev-project/sim-ln/issues/26#issuecomment-1691780018
+                            ListpaysPaysStatus::Failed => PaymentOutcome::UnexpectedError,
+                        };
+                        let htlc_count = pay.number_of_parts.unwrap_or(1).try_into().map_err(|_| LightningError::TrackPaymentError("Invalid number of parts".to_string()))?;
+                        return Ok(PaymentResult {
+                            htlc_count,
+                            payment_outcome,
+                        });
+                    }
+                },
+            }
+        }
     }
 
     async fn get_node_features(&mut self, node: PublicKey) -> Result<NodeFeatures, LightningError> {


sim-lib/src/cln.rs
```diff
@@ -101,10 +107,44 @@ impl LightningNode for ClnNode {
...
+                        break Ok(PaymentResult {
+                            // Task: https://github.com/bitcoin-dev-project/sim-ln/issues/26#issuecomment-1691780018
+                            htlc_count: 1,
+                            payment_outcome,
+                        });
+                    }
+                }
+            }
+
+            time::sleep(Duration::from_millis(100)).await;
```
carlaKC, line 146: Can we also select on `sleep` + `shutdown` here? That lets us exit a little faster if we shut down while tracking the payment. 

If we do this we can probably bump it up to polling once a second so that we don't hammer CLN too much.

carlaKC (commented):
🏄

sim-lib/src/cln.rs
```diff
@@ -101,10 +107,44 @@ impl LightningNode for ClnNode {
...
+
+                        let payment_outcome = match payment_status {
+                            ListpaysPaysStatus::Pending => continue,
+                            ListpaysPaysStatus::Complete => PaymentOutcome::Success,
+                            // Task: https://github.com/bitcoin-dev-project/sim-ln/issues/26#issuecomment-1691780018
+                            ListpaysPaysStatus::Failed => PaymentOutcome::UnexpectedError,
+                        };
+
+                        break Ok(PaymentResult {
+                            // Task: https://github.com/bitcoin-dev-project/sim-ln/issues/26#issuecomment-1691780018
```
sr-gi, line 138: I think this is the amount of attempts that LND required in order to complete the payment. CLN does not report that number on successful payments AFAIK.

I don't think we should care about this, and if we do, it should be an optional field in the outcome.

sim-lib/src/cln.rs
okjodom: select kind of fun

sr-gi: Aren't we sleeping twice now?

Feels like this sleep should be removed and the duration of the sleep in `select` should be reduced.

sim-lib/src/cln.rs
```diff
@@ -101,10 +107,47 @@ impl LightningNode for ClnNode {
...
+                            .ok_or(LightningError::TrackPaymentError("Invalid payment status".to_string()))?;
+
+                        let payment_outcome = match payment_status {
+                            ListpaysPaysStatus::Pending => continue,
+                            ListpaysPaysStatus::Complete => PaymentOutcome::Success,
+                            // Task: https://github.com/bitcoin-dev-project/sim-ln/issues/26#issuecomment-1691780018
+                            ListpaysPaysStatus::Failed => PaymentOutcome::UnexpectedError,
+                        };
+                        let htlc_count = pay.number_of_parts.unwrap_or(1).try_into().map_err(|_| LightningError::TrackPaymentError("Invalid number of parts".to_string()))?;
+                        break Ok(PaymentResult {
```
sr-gi, line 141: Is that break returning? Please don't. This is literally the return of the method.

sim-lib/src/cln.rs
```diff
@@ -101,10 +107,47 @@ impl LightningNode for ClnNode {
...
-        _shutdown: Listener,
+        hash: PaymentHash,
+        shutdown: Listener,
     ) -> Result<PaymentResult, LightningError> {
-        unimplemented!()
+        loop {
+            tokio::select! {
+                biased;
+                _ = shutdown.clone() => {
+                    break Err(LightningError::TrackPaymentError("Shutdown before tracking results".to_string()));
```
sr-gi, line 117: Same here, replace by `return`

sr-gi: I also think it is worth considering that, if a shutdown signal is raised, we may want to wait until the next tick before stopping the simulation.

This would certainly need a followup, but I think it is worth we discuss it (and we open an issue for it if it makes sense)

sim-lib/src/cln.rs
```diff
@@ -101,10 +107,47 @@ impl LightningNode for ClnNode {
...
+                        .into_inner();
+
+                    if let Some(pay) = pays.first() {
+                        let payment_status = ListpaysPaysStatus::from_i32(pay.status)
+                            .ok_or(LightningError::TrackPaymentError("Invalid payment status".to_string()))?;
+
+                        let payment_outcome = match payment_status {
+                            ListpaysPaysStatus::Pending => continue,
+                            ListpaysPaysStatus::Complete => PaymentOutcome::Success,
+                            // Task: https://github.com/bitcoin-dev-project/sim-ln/issues/26#issuecomment-1691780018
```
sr-gi, line 137: I'll look into this.

sim-lib/src/cln.rs
```diff
@@ -1,18 +1,24 @@
+use std::time::Duration;
```
sr-gi, line 1: You can import `Duration` from `tokio::time` given we are already adding that dependency.

sim-lib/src/cln.rs
okjodom: return works, but imo `break val` here better communicates the exit condition of the outer `loop`. its really easy to miss the loop.

sim-lib/src/cln.rs
okjodom: good point

sim-lib/src/cln.rs
okjodom: > wait until the next tick before stopping the simulation

would that be a tick in the current track payment task, i.e 1 second later? how would we "harmonize waiting" for all the other timers and timeouts?

sim-lib/src/cln.rs
sr-gi: Sure, but there is nothing outside the outer loop, nor are you using the value returned by the loop to anything but returning from the method, hence what I think returning from there makes more sense

sim-lib/src/cln.rs
sr-gi: What I had in mind is, when a shutdown is triggered:

- First, you stop the producers, so no new payments are passed down the pipeline
- Then you wait for whatever we feel is reasonable for inflight payments to complete

okjodom: concept ack. I created a tracking issue here: https://github.com/bitcoin-dev-project/sim-ln/issues/82

sim-lib/src/cln.rs
okjodom: makes sense

sim-lib/src/cln.rs
okjodom: ack. I was inattentive!

carlaKC:
Looks good! Going to find some time to test this in the wild this week, should be good to go after that 🥇

sr-gi:
LGTM too. There is not much we can do regarding tracking without polling atm, and there is an open issue at CLN for the error reporting in `list_pays` on payment failure

carlaKC (approved):
tACK! Sorry this took me so long to get to. 

Tested LND <-> CLN keysend payments, everything as expected + tracked in results. 

We do however need to take another look at error handling for _sending_ CLN payments, when I tried to test out failures (just closed the only channel) I get an error that shuts down the simulation: 
```
ERROR [sim_lib] Error while sending payment 0366dc9e42d6cb2a89d3ba9a347a0995c88e62949618db6f62bd86fd1da0364868 
-> 02c0c6b8be1a367c2fce4efdd9ea075b1e2b15d2566c6eab845513a88e313ee0ab. 

Terminating consumer. Send payment error status: 
Unknown, message: "Error calling method KeySend: RpcError { 
code: Some(210), message: \"Ran out of routes to try after 1 attempt: see `paystatus`\" }", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 11 Sep 2023 15:48:33 GMT", "content-length": "0"} 
}
```

Created #96 to track!

PR #76, opened 2023-08-24
batch print results to csv

### LINKED ISSUE #59: Robustness: Batched export of CSV Results
We're going with a MVP-ey flush to CSV - we should batch this to only write every X results.

okjodom:
closes #59

### DIFF
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -6,7 +6,6 @@ use lightning::ln::PaymentHash;
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
 use std::marker::Send;
-use std::mem::size_of;
 use std::{collections::HashMap, sync::Arc, time::SystemTime};
 use thiserror::Error;
 use tokio::sync::mpsc::{channel, Receiver, Sender};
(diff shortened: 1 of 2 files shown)


sim-lib/src/lib.rs
```diff
@@ -501,18 +500,19 @@ async fn write_payment_results(
...
+        "simulation_{:?}.csv",
+        SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap()
+            .as_secs()
+    ))?;
 
     let mut result_logger = PaymentResultLogger::new();
 
+    let batch_print = 5;
```
carlaKC: nit: `batch_size`?

sim-lib/src/lib.rs
```diff
@@ -182,13 +182,18 @@ pub struct Simulation {
     shutdown_listener: Listener,
     // Total simulation time. The simulation will run forever if undefined.
     total_time: Option<time::Duration>,
+    /// The number of activity results to batch before printing in CSV.
+    batch_print: u32,
 }
 
+const DEFAULT_BATCH_PRINT: u32 = 5;
```
carlaKC, line 189: Lazy back of envelope for picking a value: 
- We don't want too many events in memory, let's be conservative and say no more than 128kb 
- We don't want to hit disk _too_ often, in case somebody is running on a pi

Each CSV row is ~sorta 130 bytes, which means we can have a buffer of about 1000 events. That seems a little high for "toy" scenarios so let's say handwave handwave 500?

carlaKC (commented):
Very nice! My nitty soul wants to call this `batch_size` across the board (since that's a somewhat known term in lightning land) but non-blocking .

Would like to update the default then this can sheep 🚢

sim-lib/src/lib.rs
okjodom: yess

sim-lib/src/lib.rs
okjodom: > We don't want to hit disk too often, in case somebody is running on a pi

Directly influenced by the frequency of events, so 500 might _feel_ like a large batch size for a user watching the file.
Good thing we have a cli config of the batch size :)

sim-lib/src/lib.rs
okjodom: `print_batch_size` just so we remember what we are batching?..

okjodom:
latest update: I dropped the csv patch because I was wrong and didn't need it

carlaKC (approved):
tACK, tested batch of 15 with 30 payments and manually checked that it writes in 2 batches 🔥

carlaKC:
@okjodom hit the button on this if you're ready!

PR #88, opened 2023-09-11
multi: update workspace resolver and fix cln alias

### LINKED ISSUE #87: Compile error when building cli
When running `cargo install --path sim-cli/`

I got:

```
error[E0308]: mismatched types
  --> sim-lib/src/cln.rs:78:17
   |
78 |                 alias,
   |                 ^^^^^ expected `String`, found `Option<String>`
   |
   = note: expected struct `std::string::String`
                found enum `std::option::Option<std::string::String>`
```

carlaKC:
Fixes #87.

We've ignored a warning: 
```
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"
```

Which means that `cargo install` breaks: 
```
error[E0308]: mismatched types
  --> sim-lib/src/cln.rs:78:17
   |
78 |                 alias,
   |                 ^^^^^ expected `String`, found `Option<String>`
   |
   = note: expected struct `std::string::String`
                found enum `std::option::Option<std::string::String>`

For more information about this error, try `rustc --explain E0308`.
```

I _think_ that this happened when we bumped minimum rust version via some dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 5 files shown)


carlaKC:
Merging this without review so that install isn't broken on main.

sr-gi (commented):
Post-merge review.

I've been digging into this because I didn't get how the resolver had anything to do with this. Turns out it doesn't, at least AFAIK.

The issue comes from our own [manifest](https://github.com/bitcoin-dev-project/sim-ln/blob/main/sim-lib/Cargo.toml#L11), sort of. 

```
cln-grpc = "0.1.3"
```

This implies that we are allowing minimum package version updates, that is, any 0.1.X, but not 0.2.X.

[cln-grpc v 1.0.4](https://crates.io/crates/cln-grpc/0.1.4) was released 16 days ago, with a breaking change for the alias (`Option<String>` instead of `String`), hence breaking our code.

carlaKC:
Sequence of events here: 

- 23 August: be166381 added `cln-grpc` dependency at `v0.1.3` on 23 August. 
- 28 August: `cln-grpc` [v0.1.4](https://crates.io/crates/cln-grpc/0.1.4) was released with a breaking change (changing `alias` to an `Option`)
- Cargo build and install worked because they bind to the version that we provide in our `Cargo.lock`.
- Cargo install breaks because without `--locked` it'll accept any latest minor version of a dependency, so it upgraded itself to the breaking change `cln-grpc`.

🤯  

h/t @sr-gi for figuring out how rust builds work :') resolver stuff is unrelated and opened #100 to get rid of unnecessary resolver lines in the lib/cli.

PR #90, opened 2023-09-11
Update README add log level to run command

maxwedwards:
Add log level flag to run example. Cli doesn't run for me without it.


carlaKC (approved):
Thanks for the PR! 

Agree that we definitely shouldn't require this, but merging in the meantime so that we don't have broken readme instructions :(

PR #97, opened 2023-09-11
multi: Set default log level to info and note in readme

### LINKED ISSUE #93: On the CLI log level is required
I've created a PR (#90) if this is the desired behaviour to update the README.

Personally, I would not accept the PR and would instead make log level have a default value.

Also, I tried info and it didn't seem to like it. The help on the cli could perhaps suggest what the valid log levels are.

carlaKC:
Fixes #93.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,13 +105,13 @@ cargo install --path sim-cli/
 
 Run Simulation with Config: 
 ```
-sim-cli --log-level debug --config config.json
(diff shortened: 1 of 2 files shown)


sr-gi:
I was fixing this here: https://github.com/sr-gi/sim-ln/tree/fix-loglevel-required

But your solution is cleaner.

LGTM.

PR #102, opened 2023-09-15
lnd: set no fee limit on payments

carlaKC:
When we don't set a fee limit for payments, LND will set the limit to the amount of the payment. This can make smaller payments fail in scenarios where the default base fee is more than the payment amount.

We update our sending code to have no fee limit, because we're more concerned with payment activity than fees. If we'd like to think about fees, we'll surface this on an activity and pass it through.


okjodom (approved):
lgtm

PR #103, opened 2023-09-15
Allows using aliases within the config file

sr-gi:
The goal of this PR is to allow aliases to be used throughout the config file to identify nodes, so the file is more human-readable. This involves:

- [x] Allow aliases to be defined as node identifiers
- [x] Allow aliases to be used as `source` of activities
- [x] Allow aliases to be used as `destination` of activities **as long as we control those nodes***
- [x] Include both aliases and short versions of public keys when logging activities 

\* We cannot use aliases as `destination` for nodes we don't control, given we have no way of mapping the aliases to the nodes' public keys. Notice aliases here do not have to match node aliases, they are just user-defined identifiers, and even if they did, node aliases don't have to be unique.

### DIFF
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -6,7 +6,9 @@ use tokio::sync::Mutex;
 
 use clap::Parser;
 use log::LevelFilter;
-use sim_lib::{cln::ClnNode, lnd::LndNode, Config, LightningNode, NodeConnection, Simulation};
+use sim_lib::{
+    cln::ClnNode, lnd::LndNode, Config, LightningError, LightningNode, NodeConnection, Simulation,
+};
 use simple_logger::SimpleLogger;
 
 #[derive(Parser)]
@@ -38,6 +40,7 @@ async fn main() -> anyhow::Result<()> {
     let Config { nodes, activity } = serde_json::from_str(&config_str)?;
 
     let mut clients: HashMap<PublicKey, Arc<Mutex<dyn LightningNode + Send>>> = HashMap::new();
+    let mut alias_node_map = HashMap::new();
 
     for connection in nodes {
         // TODO: Feels like there should be a better way of doing this without having to Arc<Mutex<T>>> it at this time.
@@ -56,6 +59,21 @@ async fn main() -> anyhow::Result<()> {
             node_info.pubkey
         );
 
+        if clients.contains_key(&node_info.pubkey) {
+            anyhow::bail!(LightningError::ValidationError(format!(
+                "duplicated node: {}.",
+                node_info.pubkey
+            )));
+        }
+
+        if alias_node_map.contains_key(&node_info.alias) {
+            anyhow::bail!(LightningError::ValidationError(format!(
+                "duplicated node: {}.",
+                node_info.alias
+            )));
+        }
+
+        alias_node_map.insert(node_info.alias.clone(), node_info.pubkey);
         clients.insert(node_info.pubkey, node);
     }
 
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -40,33 +40,23 @@ async fn main() -> anyhow::Result<()> {
     let mut clients: HashMap<PublicKey, Arc<Mutex<dyn LightningNode + Send>>> = HashMap::new();
 
     for connection in nodes {
-        // TODO: We should simplify this into two minimal branches plus shared logging and inserting into the list
-        match connection {
-            NodeConnection::LND(c) => {
-                let node_id = c.id;
-                let lnd = LndNode::new(c).await?;
+        // TODO: Feels like there should be a better way of doing this without having to Arc<Mutex<T>>> it at this time.
+        // Box sort of works, but we won't know the size of the dyn LightningNode at compile time so the compiler will
+        // scream at us when trying to create the Arc<Mutex>> later on while adding the node to the clients map
+        let node: Arc<Mutex<dyn LightningNode + Send>> = match connection {
+            NodeConnection::LND(c) => Arc::new(Mutex::new(LndNode::new(c).await?)),
+            NodeConnection::CLN(c) => Arc::new(Mutex::new(ClnNode::new(c).await?)),
+        };
 
-                log::info!(
-                    "Connected to {} - Node ID: {}.",
-                    lnd.get_info().alias,
-                    lnd.get_info().pubkey
-                );
+        let node_info = node.lock().await.get_info().clone();
 
-                clients.insert(node_id, Arc::new(Mutex::new(lnd)));
-            }
-            NodeConnection::CLN(c) => {
-                let node_id = c.id;
-                let cln = ClnNode::new(c).await?;
+        log::info!(
+            "Connected to {} - Node ID: {}.",
+            node_info.alias,
+            node_info.pubkey
+        );
 
-                log::info!(
-                    "Connected to {} - Node ID: {}.",
-                    cln.get_info().alias,
-                    cln.get_info().pubkey
-                );
-
-                clients.insert(node_id, Arc::new(Mutex::new(cln)));
-            }
-        }
+        clients.insert(node_info.pubkey, node);
     }
 
     let sim = Simulation::new(clients, activity, cli.total_time, cli.print_batch_size);
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -38,12 +38,10 @@ async fn main() -> anyhow::Result<()> {
         .unwrap();
 
     let config_str = std::fs::read_to_string(cli.config)?;
-    let Config {
-        nodes,
-        mut activity,
-    } = serde_json::from_str(&config_str)?;
+    let Config { nodes, activity } = serde_json::from_str(&config_str)?;
 
     let mut clients: HashMap<PublicKey, Arc<Mutex<dyn LightningNode + Send>>> = HashMap::new();
+    let mut pk_node_map = HashMap::new();
     let mut alias_node_map = HashMap::new();
 
     for connection in nodes {
@@ -77,36 +75,67 @@ async fn main() -> anyhow::Result<()> {
             )));
         }
 
-        alias_node_map.insert(node_info.alias.clone(), node_info.pubkey);
         clients.insert(node_info.pubkey, node);
+        pk_node_map.insert(node_info.pubkey, node_info.clone());
+        alias_node_map.insert(node_info.alias.clone(), node_info);
     }
 
     let mut validated_activities = vec![];
     // Make all the activities identifiable by PK internally
-    for act in activity.iter_mut() {
+    for act in activity.into_iter() {
         // We can only map aliases to nodes we control, so if either the source or destination alias
         // is not in alias_node_map, we fail
-        if let NodeId::Alias(a) = &act.source {
-            if let Some(pk) = alias_node_map.get(a) {
-                act.source = NodeId::PublicKey(*pk);
-            } else {
-                anyhow::bail!(LightningError::ValidationError(format!(
-                    "activity source alias {} not found in nodes.",
-                    act.source
-                )));
+        let source = if let Some(source) = match &act.source {
+            NodeId::PublicKey(pk) => pk_node_map.get(pk),
+            NodeId::Alias(a) => alias_node_map.get(a),
+        } {
+            source.clone()
+        } else {
+            anyhow::bail!(LightningError::ValidationError(format!(
+                "activity source {} not found in nodes.",
+                act.source
+            )));
+        };
+
+        let destination = match &act.destination {
+            NodeId::Alias(a) => {
+                if let Some(info) = alias_node_map.get(a) {
+                    info.clone()
+                } else {
+                    anyhow::bail!(LightningError::ValidationError(format!(
+                        "unknown activity destination: {}.",
+                        act.destination
+                    )));
+                }
             }
-        }
-        if let NodeId::Alias(a) = &act.destination {
-            if let Some(pk) = alias_node_map.get(a) {
-                act.destination = NodeId::PublicKey(*pk);
-            } else {
-                anyhow::bail!(LightningError::ValidationError(format!(
-                    "unknown activity destination: {}.",
-                    act.destination
-                )));
+            NodeId::PublicKey(pk) => {
+                if let Some(info) = pk_node_map.get(pk) {
+                    info.clone()
+                } else {
+                    clients
+                        .get(&source.pubkey)
+                        .unwrap()
+                        .lock()
+                        .await
+                        .get_node_info(pk)
+                        .await
+                        .map_err(|e| {
+                            log::debug!("{}", e);
+                            LightningError::ValidationError(format!(
+                                "Destination node unknown or invalid: {}.",
+                                pk,
+                            ))
+                        })?
+                }
             }
-        }
-        validated_activities.push(ActivityDefinition::try_from(act)?);
+        };
+
+        validated_activities.push(ActivityDefinition {
+            source,
+            destination,
+            interval_secs: act.interval_secs,
+            amount_msat: act.amount_msat,
+        });
     }
 
     let sim = Simulation::new(
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -7,7 +7,8 @@ use tokio::sync::Mutex;
 use clap::Parser;
 use log::LevelFilter;
 use sim_lib::{
-    cln::ClnNode, lnd::LndNode, Config, LightningError, LightningNode, NodeConnection, Simulation,
+    cln::ClnNode, lnd::LndNode, ActivityDefinition, Config, LightningError, LightningNode,
+    NodeConnection, NodeId, Simulation,
 };
 use simple_logger::SimpleLogger;
 
@@ -37,7 +38,10 @@ async fn main() -> anyhow::Result<()> {
         .unwrap();
 
     let config_str = std::fs::read_to_string(cli.config)?;
-    let Config { nodes, activity } = serde_json::from_str(&config_str)?;
+    let Config {
+        nodes,
+        mut activity,
+    } = serde_json::from_str(&config_str)?;
 
     let mut clients: HashMap<PublicKey, Arc<Mutex<dyn LightningNode + Send>>> = HashMap::new();
     let mut alias_node_map = HashMap::new();
@@ -77,7 +81,40 @@ async fn main() -> anyhow::Result<()> {
         clients.insert(node_info.pubkey, node);
     }
 
-    let sim = Simulation::new(clients, activity, cli.total_time, cli.print_batch_size);
+    let mut validated_activities = vec![];
+    // Make all the activities identifiable by PK internally
+    for act in activity.iter_mut() {
+        // We can only map aliases to nodes we control, so if either the source or destination alias
+        // is not in alias_node_map, we fail
+        if let NodeId::Alias(a) = &act.source {
+            if let Some(pk) = alias_node_map.get(a) {
+                act.source = NodeId::PublicKey(*pk);
+            } else {
+                anyhow::bail!(LightningError::ValidationError(format!(
+                    "activity source alias {} not found in nodes.",
+                    act.source
+                )));
+            }
+        }
+        if let NodeId::Alias(a) = &act.destination {
+            if let Some(pk) = alias_node_map.get(a) {
+                act.destination = NodeId::PublicKey(*pk);
+            } else {
+                anyhow::bail!(LightningError::ValidationError(format!(
+                    "unknown activity destination: {}.",
+                    act.destination
+                )));
+            }
+        }
+        validated_activities.push(ActivityDefinition::try_from(act)?);
+    }
+
+    let sim = Simulation::new(
+        clients,
+        validated_activities,
+        cli.total_time,
+        cli.print_batch_size,
+    );
     let sim2 = sim.clone();
 
     ctrlc::set_handler(move || {
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -33,31 +33,9 @@ pub const ACTIVITY_MULTIPLIER: f64 = 2.0;
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum NodeConnection {
     #[serde(alias = "lnd", alias = "Lnd")]
-    LND(LndConnection),
+    LND(lnd::LndConnection),
     #[serde(alias = "cln", alias = "Cln")]
-    CLN(ClnConnection),
-}
-
-#[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct LndConnection {
-    pub id: PublicKey,
-    pub address: String,
-    #[serde(deserialize_with = "serializers::deserialize_path")]
-    pub macaroon: String,
-    #[serde(deserialize_with = "serializers::deserialize_path")]
-    pub cert: String,
-}
-
-#[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct ClnConnection {
-    pub id: PublicKey,
-    pub address: String,
-    #[serde(deserialize_with = "serializers::deserialize_path")]
-    pub ca_cert: String,
-    #[serde(deserialize_with = "serializers::deserialize_path")]
-    pub client_cert: String,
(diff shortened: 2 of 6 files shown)


sr-gi:
@carlaKC @okjodom I've added the changes to allow aliases to be used as ids in the config file (check [8b65670](https://github.com/bitcoin-dev-project/sim-ln/pull/103/commits/8b6567070d529440ca3b569abc5155555298990e) for the code and rationale).

Looking for an Approach ACK/ tACK of this before moving along with the testing, given the data passed to the simulator to be able to handle logging both PKs and aliases is dependant on how we approach this.

sim-lib/src/lnd.rs
```diff
@@ -59,6 +60,19 @@ impl LndNode {
...
+        match connection.id {
+            crate::NodeId::PublicKey(pk) => {
+                if pk != node_id {
+                    return Err(LightningError::ValidationError(format!(
+                    "The provided node id does not match the one returned by the backend ({} != {})",
+                    pk, node_id,
+                )));
+                }
+            }
+            crate::NodeId::Alias(a) => alias = a,
```
okjodom, line 73: Why do we assert that  pubkey used as ID must match the node pubkey, but for the alias, we prefer the user provided value, ignoring the actual node alias

sim-lib/src/lib.rs
```diff
@@ -247,27 +273,36 @@ impl Simulation {
...
-                self.nodes
-                    .get(&payment_flow.source)
-                    .ok_or(LightningError::ValidationError(format!(
-                        "source node not found {}",
-                        payment_flow.source,
-                    )))?;
+            let src_id = payment_flow
+                .source
+                .get_pk()
+                .map_err(LightningError::ValidationError)?;
```
okjodom, line 279: just realized these should be `SimulationError::ValidationError`

sim-lib/src/lib.rs
```diff
@@ -35,6 +35,29 @@ pub enum NodeId {
     Alias(String),
 }
 
+impl NodeId {
+    pub fn get_pk(&self) -> Result<&PublicKey, String> {
+        if let NodeId::PublicKey(pk) = self {
+            Ok(pk)
+        } else {
+            panic!("NodeId is not a PublicKey. This should never happen, please report")
```
okjodom, line 43: maybe we could just return a `SimulationError`. No need to panic :)

sim-lib/src/lib.rs
```diff
@@ -383,8 +418,8 @@ impl Simulation {
         for (id, node) in self.nodes.iter().filter(|(pk, _)| {
             self.activity
                 .iter()
-                .map(|a| a.source)
-                .collect::<HashSet<PublicKey>>()
+                .map(|a| a.source.get_pk().unwrap())
```
okjodom, line 421: okay but can we avoid the assumption of having called `validate_activity` before this point?
Same sentiment for the other `get_pk.unwrap()` points

sim-lib/src/lib.rs
okjodom: We can message this as `"NodeId is not a PublicKey. Please prefer to use actual node pubkey for the node aliased as {}"`. 

Any dependent on `sim-lib` who passes in activity definition with aliases knows what to change, _instead of reporting (?)_

okjodom:
cACK on e2d5fb3badc39ce29a704b00ef2e15635406b61f allowing aliases as source or destination.

> we internally identify activities using PK, so we do an additional mapping
step once the data is loaded to go from NodeId{PK, Alias} to NodeId(PK), hence, once data
is passed to the simulator it is always identified by NodeId(PK)

Sounds to me like a specific feature offering of `sim-cli`, or any other UI we build on top of `sim-lib`

> after `Simulation::validate_activity` it is safe to call `NodeId::get_pk::unwrap()`,
given no NodeId::Alias should have passed validation.

I have strong reactions against this. Had suggestions to better communicate the error case and avoid the panic, but that doesn't seem sufficient.

sim-lib/src/lib.rs
```diff
@@ -85,19 +93,19 @@ pub enum SimulationError {
 // Phase 2: Event Queue
 #[derive(Debug, Error)]
 pub enum LightningError {
-    #[error("Node connection error {0}")]
+    #[error("Node connection error: {0}")]
```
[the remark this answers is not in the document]
sr-gi: I'm happy to roll it back so there's no conflict with yours

sim-lib/src/lib.rs
sr-gi: Agreed, tho that does not exist, we may want to move it.

sim-lib/src/lib.rs
sr-gi: Ups, I forgot to change that to an error.

Regarding encouraging one or the other, I don't think we should. I'm happy with getting rid of the reporting stuff if we think of this as something that can be used by a third party, but we actually don't care whether they use one variant or the other, we cast everything to PK internally

sim-lib/src/lib.rs
sr-gi: Can we? Sure, but I don't think we should. `validate_activity` is our sanity checker, we could modify all these calls to account for an alias being here, but that's exactly what is not supposed to happen. If any `NodeId` contains an alias at this point, it means the code if wrong (hence why there was a "report message".

I could patch the code so it does not unwrap and it fails more "elegantly", gut all in all we want it to fail if that happens

sr-gi: Alternatively, we could have two different classes to account for node definition and activity definition, one that interfaces with the config file, and where we allow aliases and pks (data is parsed as `NodeId`), and an internal one that gets constructed from the former, and only accepts pks. In that way, we can be sure at compile time that everything is identified by pk internally.

sim-lib/src/lib.rs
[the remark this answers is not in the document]
sr-gi: I rolled this back

sim-lib/src/lib.rs
sr-gi: Fixed

carlaKC (commented):
Concept ACK, with a slight preference to shift some handling to the cli as mentioned [here](https://github.com/bitcoin-dev-project/sim-ln/pull/103/files#r1332865062).

sim-lib/src/lib.rs
sr-gi: I've split this into two sets of data structures in [2546166](https://github.com/bitcoin-dev-project/sim-ln/pull/103/commits/2546166018c9dbcaf51b6995b099b315e4070919). The `Parsers` that interact directly with the config file and allow `NodeId`s to be loaded, and the regular ones which roll back to use `PublicKey`s and can be built from the former.

This commit can be squashed with the previous one. Also, I'm happy to replace `Parser` with any other better term we can come up with (`Loader`, `Reader`, ...)

sim-lib/src/lnd.rs
sr-gi: I've rebased [bf6f31b](https://github.com/bitcoin-dev-project/sim-ln/pull/103/commits/bf6f31bdecd293b1d4a06bb7006d7792df699788) and added the warning

carlaKC (commented):
Just done a high level read through of the new approach, cACK!

sr-gi:
I've added the last bit here. Notice how the last commit is pretty opinionated. I'm happy to re-visit the approach or even approach it in another way as long as it does not end up being super messy (which has been my experience when trying to go for some of the alternatives).

sim-lib/src/lib.rs
```diff
@@ -149,6 +130,18 @@ pub struct NodeInfo {
     pub features: NodeFeatures,
 }
 
+impl Display for NodeInfo {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let pk = self.pubkey.to_string();
+        let pk_summary = format!("{}...{}", &pk[..6], &pk[pk.len() - 6..]);
```
carlaKC, line 136: 😻

carlaKC (commented):
Tested out a few variations (source as alias, description as alias of known node etc), such a nice improvement on just vanilla pubkeys! 

Submitting what I have so far - main Q about the `LNDConnection` in conjunction with the `Parser` I can delete it and it seems like everything works.

sim-lib/src/lnd.rs
```diff
@@ -214,7 +214,10 @@ impl LightningNode for LndNode {
         }
     }
 
-    async fn get_node_features(&mut self, node: PublicKey) -> Result<NodeFeatures, LightningError> {
+    async fn get_node_features(
+        &mut self,
+        node: &PublicKey,
+    ) -> Result<NodeFeatures, LightningError> {
```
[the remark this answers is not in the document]
sr-gi: I think this actually happened because one character was enough to trigger the line split (notice we're passing `node` as a reference now)

sim-lib/src/lib.rs
```diff
@@ -35,23 +35,78 @@ pub enum NodeId {
...
     // The source of the payment.
     pub source: PublicKey,
     // The destination of the payment.
     pub destination: PublicKey,
     // The interval of the event, as in every how many seconds the payment is performed.
-    #[serde(alias = "interval_secs")]
     pub interval: u16,
     // The amount of m_sat to used in this payment.
-    pub amount_msat: u64,
+    pub amount: u64,
```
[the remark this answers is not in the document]
sr-gi: Does it?

The only place we're constructing an `ActivityDefinition` instance is in main, and we need to parse it from `act` anyway:

```
activities.push(ActivityDefinition {
        source,
        destination,
        interval: act.interval_secs,
        amount: act.amount_msat,
});
```

Having said that, I don't really have a strong opinion for this, so I'm happy to change it if you do

sim-cli/src/main.rs
```diff
@@ -40,33 +40,23 @@ async fn main() -> anyhow::Result<()> {
...
+            NodeConnection::LND(c) => Arc::new(Mutex::new(LndNode::new(c).await?)),
+            NodeConnection::CLN(c) => Arc::new(Mutex::new(ClnNode::new(c).await?)),
+        };
 
-                log::info!(
-                    "Connected to {} - Node ID: {}.",
-                    lnd.get_info().alias,
-                    lnd.get_info().pubkey
-                );
+        let node_info = node.lock().await.get_info().clone();
```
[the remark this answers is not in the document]
sr-gi: Yeah, I think it's not that bad, I just feel a higher-ranked wizard may find a nicer way to do so. Looks like I'll have to go back to Hogwarts, or leave it as is until someone more talented nitpicks about it

sim-cli/src/main.rs
```diff
@@ -77,7 +81,35 @@ async fn main() -> anyhow::Result<()> {
...
+        if let NodeId::Alias(a) = &act.source {
+            if let Some(pk) = r_alias_map.get(a) {
+                act.source = NodeId::PublicKey(*pk);
+            } else {
+                anyhow::bail!(LightningError::ValidationError(format!(
+                    "unknown activity source: {}.",
+                    act.source
+                )));
+            }
+        }
```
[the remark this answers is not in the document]
sr-gi: This changes after that commit, is it still duplicated?

sim-cli/src/main.rs
```diff
@@ -77,7 +81,35 @@ async fn main() -> anyhow::Result<()> {
         clients.insert(node_info.pubkey, node);
     }
 
-    let sim = Simulation::new(clients, activity, cli.total_time, cli.print_batch_size);
+    let mut activities = vec![];
```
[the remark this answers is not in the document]
sr-gi: ACK `validated_activities`

sim-lib/src/lnd.rs
```diff
@@ -18,15 +20,24 @@ const KEYSEND_KEY: u64 = 5482373484;
...
+    pub id: NodeId,
     pub address: String,
     #[serde(deserialize_with = "serializers::deserialize_path")]
     pub macaroon: String,
     #[serde(deserialize_with = "serializers::deserialize_path")]
     pub cert: String,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct LndConnection {
```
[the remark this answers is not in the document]
sr-gi: You're right! I wanted to create a converter from the `Parser` to the `Connection`, but I ended up implementing the logic to handle the parsing there, so this is indeed unnecessary.

Will get rid of the `Parser` classes (i.e. rename them to `Connection` so we have a smaller diff)

Nice catch!

sr-gi:
Addressed comments by @carlaKC

sim-lib/src/lib.rs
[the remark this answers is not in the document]
carlaKC: I think it makes sense to explicitly note when we're using msat/sat because both are represented with u64 and it's very easy to run into "off by 1000" errors in LN. 

>  Saves us a few line changes lower down as well.

This acts as a rename in `lib.rs`, so we have to change the logging in `produce_events` to just use `amount` ([around here](https://github.com/bitcoin-dev-project/sim-ln/pull/103/commits/47b3f5a2a62c0fe6cde0b4640eeee1de88a1bd42#diff-f891364aff98268ac6d94109bd61594c23af50e37e547fec9b0ea44a816459a0R566)).

sim-cli/src/main.rs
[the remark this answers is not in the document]
carlaKC: Ah yeah, was going by commit reviewing.

sim-cli/src/main.rs
```diff
@@ -70,43 +71,74 @@ async fn main() -> anyhow::Result<()> {
...
-        if let NodeId::Alias(a) = &act.destination {
-            if let Some(pk) = r_alias_map.get(a) {
-                act.destination = NodeId::PublicKey(*pk);
-            } else {
-                anyhow::bail!(LightningError::ValidationError(format!(
-                    "unknown activity destination: {}.",
-                    act.destination
-                )));
+            NodeId::PublicKey(pk) => {
+                if let Some(info) = pk_node_map.remove(pk) {
```
carlaKC, line 115: Do we need to remove here? As is, config with multiple activity descriptions for a single source node will fail.

I think that we want to be able to support activity of this nature: 
```
Alice -> Dave / 10 sat every 5 mins
Alice -> Bob / 1000 sat every 30 mins
```

carlaKC (commented):
Looking great ✨ 

Only major comment is allowing multiple activity descriptions for the same source (last commit). Really happy with the way everything else turned out.

carlaKC:
> Only major comment is allowing multiple activity descriptions for the same source (last commit). Really happy with the way everything else turned out.

Tested with the following config to sanity check that we do actually fail with repeated sources: 
```
{
    "nodes": [
        {
            "LND": {
                "id": "alice",
                "address": "https://localhost:10011",
                "macaroon": "~/alice/admin.macaroon",
                "cert": "~/alice/tls.cert"
            }
        },
		{
            "LND": {
                "id": "dave",
                "address": "https://localhost:10014",
                "macaroon": "~/dave/admin.macaroon",
                "cert": "~/dave/tls.cert"
            }
        }
],
	"activity": [
		{
			"source": "alice",
			"destination": "03fd1854561cfd7e3b58714f4f8c597bba29428100352b207c0d2baef832bdf390",
			"interval_secs": 2,
			"amount_msat": 100
		},
		{
			"source": "alice",
			"destination": "0374f63bfb1a0e9607b05ac6a7fb5393216e6ffcdf80700ec6da7131815249881c",
			"interval_secs": 5,
			"amount_msat": 1000
		},
		{
			"source": "dave",
			"destination": "alice",
			"interval_secs": 2,
			"amount_msat": 100
		}

	]
}
```

-> `Error: Config validation failed: unknown activity source: alice.`

sim-lib/src/lib.rs
```diff
@@ -245,7 +299,7 @@ impl Simulation {
         for payment_flow in self.activity.iter() {
             // We need every source node that is configured to execute some activity to be included in our set of
             // nodes so that we can execute events on it.
-            let source_node =
+            let src_node =
```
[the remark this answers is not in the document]
sr-gi: We don't actually. That's even dropped later on. Let me clean it.

sim-cli/src/main.rs
sr-gi: Oh damn, that was me trying to be clever by avoiding copying the `NodeInfo` :D

This actually applies to both repeated sources and destinations. My bad.

sim-cli/src/main.rs
```diff
@@ -77,7 +81,40 @@ async fn main() -> anyhow::Result<()> {
...
+    // Make all the activities identifiable by PK internally
+    for act in activity.iter_mut() {
+        // We can only map aliases to nodes we control, so if either the source or destination alias
+        // is not in r_alias_map, we fail
+        if let NodeId::Alias(a) = &act.source {
+            if let Some(pk) = r_alias_map.get(a) {
+                act.source = NodeId::PublicKey(*pk);
+            } else {
+                anyhow::bail!(LightningError::ValidationError(format!(
+                    "unknown activity source: {}.",
```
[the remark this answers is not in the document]
sr-gi: We'll end up with `"activity source {} not found in nodes."`, given we reuse the same error for aliases and pks, but sure, that way clearer

sim-lib/src/lib.rs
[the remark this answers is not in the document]
sr-gi: I also changed back `interval_secs` for consistency

sr-gi:
Addressed the latest comments from @carlaKC plus the pending ones from the last review.

The issue with the repeated source/destination in an activity should be fixed now

sim-cli/src/main.rs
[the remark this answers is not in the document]
okjodom: I'd avoid wizardry by waiting to create the mutex when inserting to clients map?

okjodom (commented):
change read through, lgtm

sim-cli/src/main.rs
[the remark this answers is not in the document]
sr-gi: That won't work since that's where the actual wizardry comes from.

The issue lies in that the size of `dyn LightningNode + Send` cannot be known at compilation time, and that is required to build an `Arc<Mutex<T>>`

sr-gi:
> > Only major comment is allowing multiple activity descriptions for the same source (last commit). Really happy with the way everything else turned out.
> 
> Tested with the following config to sanity check that we do actually fail with repeated sources:
> 
> ```
> {
>     "nodes": [
>         {
>             "LND": {
>                 "id": "alice",
>                 "address": "https://localhost:10011",
>                 "macaroon": "~/alice/admin.macaroon",
>                 "cert": "~/alice/tls.cert"
>             }
>         },
> 		{
>             "LND": {
>                 "id": "dave",
>                 "address": "https://localhost:10014",
>                 "macaroon": "~/dave/admin.macaroon",
>                 "cert": "~/dave/tls.cert"
>             }
>         }
> ],
> 	"activity": [
> 		{
> 			"source": "alice",
> 			"destination": "03fd1854561cfd7e3b58714f4f8c597bba29428100352b207c0d2baef832bdf390",
> 			"interval_secs": 2,
> 			"amount_msat": 100
> 		},
> 		{
> 			"source": "alice",
> 			"destination": "0374f63bfb1a0e9607b05ac6a7fb5393216e6ffcdf80700ec6da7131815249881c",
> 			"interval_secs": 5,
> 			"amount_msat": 1000
> 		},
> 		{
> 			"source": "dave",
> 			"destination": "alice",
> 			"interval_secs": 2,
> 			"amount_msat": 100
> 		}
> 
> 	]
> }
> ```
> 
> -> `Error: Config validation failed: unknown activity source: alice.`

tested with a similar setup to this one, can confirm it does not crash anymore

sr-gi:
FYI #120 needs to go in before this, given otherwise we cannot rebase main without setting the network field for nodes we don't control to something meaningful.

carlaKC:
tACK https://github.com/bitcoin-dev-project/sim-ln/pull/103/commits/f17662d07ff29133ac0755bc0d41ffa6bb332404 pending rebase on #120.

Tested out various scenarios and everything is looking great. When I test this PR the easy config makes me _so happy_ 🎆 

- Duplicate alias: `Error: Config validation failed: duplicated node: alice.`
- Alias mismatch warning:  `WARN [sim_lib] The provided alias does not match the one returned by the backend (cln != BLUESQUIRREL-v23.05-175-g910630c)`
-  Bad pubkey: `Error: Config validation failed: the provided node id does not match the one returned by the backend (0255b34764dbe7a9d5ad4e1fd7fb069bb7992128cd55467e4db50796b318a9bc9e != 037359ab1279636979218125289e5abf832c1809cb80bbb33d8982e95e597d8127).`
- Unknown destination alias: `Error: Config validation failed: unknown activity destination: loser`

Also tested out positive paths of: 
* Alias/Pubkey -> Alias/Pubkey.
* Multiple activities for a single source.

sr-gi:
Sorry, rebasing this took me longer than I wanted to. I was trying to move the `validated_activities` creation inside the `Simulator` so the validation happens internally but it has proven to be a pretty nasty change that involves passing along multiple maps and making `validate_activity` pretty messy.

I think keeping the `Simulator` interface simple is one of the design goals, and the way we personally parse information from out config in the `CLI` should not interfere with it, so I decided not to go for the change, at least for now.

I've squeezed a fix for #126 in [5c5497d](https://github.com/bitcoin-dev-project/sim-ln/pull/103/commits/5c5497d8a59b83ed7465af5e388beb758eccdee6) since being able to check that depended on having `NodeInfo` around.

sim-cli/src/main.rs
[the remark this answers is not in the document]
okjodom: you're right

sim-lib/src/lib.rs
```diff
@@ -62,27 +62,81 @@ impl NodeId {
...
 pub struct Config {
     pub nodes: Vec<NodeConnection>,
-    pub activity: Option<Vec<ActivityDefinition>>,
+    #[serde(default)]
+    pub activity: Vec<ActivityParser>,
 }
 
-#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ActivityParser {
```
carlaKC, line 96: nit: would be worthwhile to have docs explaining why we have these two structs, I can see this taking some time to figure out as a newcomer to the codebase.

sim-lib/src/lib.rs
```diff
@@ -815,32 +815,37 @@ async fn produce_random_events<N: NetworkGenerator, A: PaymentGenerator + Displa
             // Wait until our time to next payment has elapsed then execute a random amount payment to a random
             // destination.
             _ = time::sleep(wait) => {
-                let destination = network_generator.lock().await.sample_node_by_capacity(source.pubkey);
+                let (destination, capacity) = network_generator.lock().await.sample_node_by_capacity(source.pubkey);
```
carlaKC, line 818: I would have thought we want to do this validation much higher up? 

Like nodes that don't have enough capacity, if a node doesn't support keysend when we start running it never will so we should just exclude it from the start (and log a warning so the user knows why).

If we do this validation here, we could also _theoretically_ end up effectively running with one node (in the case of two nodes and one doesn't have keysend) and then just run without ever sending a payment.

carlaKC (commented):
Only comment is on the final commit adding validation. 

Would be happy to drop that off of this and open it in a separate PR so that we can hit the button on this one.

sim-lib/src/lib.rs
sr-gi: We can run this much higher up, as long as we are ok with compromising with `--accept-keysend` being enabled for both source and destination nodes. Otherwise, we cannot know if `keysend` receiving is really going to be necessary until a node is picked as a payment destination, and that happens on the fly.

sim-lib/src/lib.rs
sr-gi: Makes sense

sim-lib/src/lib.rs
carlaKC: > compromising with --accept-keysend being enabled for both source and destination nodes.

In the case of random activity I think that's okay because every sending node can reasonably be expected to also receive?

sr-gi: Fair enough. https://github.com/bitcoin-dev-project/sim-ln/pull/103/commits/98c15d076e28c96416b6151ef9ac6fd68cef2715 should have fixed that

sr-gi:
Let's go 👶 !

PR #104, opened 2023-09-15
sim-lib: continue simulation on send payment errors

### LINKED ISSUE #96: Bug: Handle Send Errors for CLN
When CLN fails a payment (eg, because there's no route), they'll fail the _entire_ API call rather than sending back an okay response _containing_ an error code. This will shut down the simulation, so we'll need to add some error handling. 

My instinct would be to take a look at the error code they're giving us (hopefully that helps classify pathfinding errors vs terminal errors).

```
ERROR [sim_lib] Error while sending payment 0366dc9e42d6cb2a89d3ba9a347a0995c88e62949618db6f62bd86fd1da0364868 
-> 02c0c6b8be1a367c2fce4efdd9ea075b1e2b15d2566c6eab845513a88e313ee0ab. 

Terminating consumer. Send payment error status: 
Unknown, message: "Error calling method KeySend: RpcError { 
code: Some(210), message: \"Ran out of routes to try after 1 attempt: see `paystatus`\" }", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 11 Sep 2023 15:48:33 GMT", "content-length": "0"} 
}
```

okjodom:
Instead of stopping the simulation on the first send payment rpc error, we gracefully handle the error and report a default send payment result. Since send payment rpc might fail before dispatching a payment (i.e in CLN case), we allow activity result reporting to skip tracking of payment if there's no payment hash

Sample results with failed keysends. The simulation continues until stopped
```csv
source,destination,hash,amount_msat,dispatch_time,htlc_count,payment_outcome
023d45dc57ed99e6d793e2fed8dcf79d2f6ad2f32ec31b48102de445e245e08336,03b73874dde2c0af24619f03b6fc5d36adbf00ad6d25ce6479d4f5b60f648c8cb6,83d340ab8a7f7b383294dd06a8fc02bcbff6c77c73debaa8e590aad550216f98,1000,1695160575397,1,Success
02c0c854f511082f9d174f9b0e51ca14fc299c25de9d65dddadebe9ac1fd529758,03b73874dde2c0af24619f03b6fc5d36adbf00ad6d25ce6479d4f5b60f648c8cb6,Unknown,10000000000,1695160576400,0,Unknown
```

TODO:
- [x] Continue simulation on send payment errors
- [x] Formalize terminal error handling
- [x] Parse CLN `send_payment` rpc errors to determine terminal failures
- [x] Parse LND `send_payment` rpc errors to determine terminal failures

fixes #96

### DIFF
--- a/sim-lib/src/cln.rs
+++ b/sim-lib/src/cln.rs
@@ -104,7 +104,18 @@ impl LightningNode for ClnNode {
                 ..Default::default()
             })
             .await
-            .map_err(|err| LightningError::SendPaymentError(err.to_string()))?
+            .map_err(|s| {
+                let message = s.message();
+                // REF: https://docs.corelightning.org/reference/lightning-keysend#return-value
+
+                if message.contains("Some(-1") | message.contains("Some(203") {
+                    // Error codes -1 and 203 indicate permanent errors
+                    LightningError::PermanentError(format!("{:?}", message))
+                } else {
+                    // Error codes, 205, 206 and 210 indicate temporary errors that can be retried
+                    LightningError::SendPaymentError(format!("{:?}", message))
+                }
+            })?
             .into_inner();
         let slice: [u8; 32] = payment_hash
             .as_slice()
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -99,8 +99,8 @@ pub enum LightningError {
     GetNodeInfoError(String),
     #[error("Config validation failed {0}")]
     ValidationError(String),
-    #[error("RPC error: {0:?}")]
-    RpcError(#[from] tonic_lnd::tonic::Status),
+    #[error("Permanent error: {0:?}")]
+    PermanentError(String),
 }
 
 #[derive(Debug, Clone)]
@@ -131,18 +131,26 @@ pub trait LightningNode {
     async fn get_node_features(&mut self, node: PublicKey) -> Result<NodeFeatures, LightningError>;
 }
 
-/// SimulationEvent describes the set of actions that the simulator can run on nodes that it has execution permissions
-/// on.
-#[derive(Clone, Copy)]
-enum SimulationEvent {
-    // Dispatch a payment of the specified amount to the public key provided.
-    SendPayment(PublicKey, u64),
-}
-
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct PaymentResult {
     pub htlc_count: usize,
     pub payment_outcome: PaymentOutcome,
+}
+
+impl PaymentResult {
+    pub fn not_dispatched() -> Self {
+        PaymentResult {
+            htlc_count: 0,
+            payment_outcome: PaymentOutcome::NotDispatched,
+        }
+    }
+
+    pub fn track_payment_failed() -> Self {
+        PaymentResult {
+            htlc_count: 0,
+            payment_outcome: PaymentOutcome::TrackPaymentFailed,
+        }
+    }
 }
 
 impl Display for PaymentResult {
@@ -167,25 +175,33 @@ pub enum PaymentOutcome {
     IncorrectPaymentDetails,
     InsufficientBalance,
     Unknown,
+    NotDispatched,
+    TrackPaymentFailed,
 }
 
-#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
-struct DispatchedPayment {
+/// Describes a payment from a source node to a destination node.
+#[derive(Debug, Clone, Copy, Serialize)]
+struct Payment {
+    /// Pubkey of the source node dispatching the payment.
     source: PublicKey,
+    /// Pubkey of the destination node receiving the payment.
     destination: PublicKey,
-    #[serde(with = "serializers::serde_payment_hash")]
-    hash: PaymentHash,
+    /// Amount of the payment in msat.
     amount_msat: u64,
+    /// Hash of the payment if it has been successfully dispatched.
+    #[serde(with = "serializers::serde_option_payment_hash")]
+    hash: Option<PaymentHash>,
+    /// Time at which the payment was dispatched.
     #[serde(with = "serde_millis")]
     dispatch_time: SystemTime,
 }
 
-impl Display for DispatchedPayment {
+impl Display for Payment {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
             "Payment {} dispatched at {:?} sending {} msat from {} -> {}",
-            hex::encode(self.hash.0),
+            self.hash.map(|h| hex::encode(h.0)).unwrap_or(String::new()),
             self.dispatch_time.duration_since(UNIX_EPOCH).unwrap(),
             self.amount_msat,
             self.source,
@@ -194,10 +210,23 @@ impl Display for DispatchedPayment {
     }
 }
 
+/// SimulationEvent describes the set of actions that the simulator can run on nodes that it has execution permissions
+/// on.
+#[derive(Clone, Copy)]
+enum SimulationEvent {
+    /// Dispatch a payment of the specified amount to the public key provided.
+    /// Results in `SimulationOutput::SendPaymentSuccess` or `SimulationOutput::SendPaymentFailure`.
+    SendPayment(PublicKey, u64),
+}
+
 /// SimulationOutput provides the output of a simulation event.
 enum SimulationOutput {
-    // The payment hash that results from a SendPayment SimulationEvent being triggered.
-    PaymentSent(DispatchedPayment),
+    /// Intermediate output for when simulator has successfully dispatched a payment.
+    /// We need to track the result of the payment to report on it.
+    SendPaymentSuccess(Payment),
+    /// Final output for when simulator has failed to dispatch a payment.
+    /// Report this as the final result of simulation event.
+    SendPaymentFailure(Payment, PaymentResult),
 }
 
 #[derive(Clone)]
@@ -391,6 +420,7 @@ impl Simulation {
                 node.clone(),
                 receiver,
                 output_sender.clone(),
+                shutdown.clone(),
             ));
 
             // Add the producer channel to our map so that various activity descriptions can use it. We may have multiple
@@ -421,6 +451,7 @@ async fn consume_events(
     node: Arc<Mutex<dyn LightningNode + Send>>,
     mut receiver: Receiver<SimulationEvent>,
     sender: Sender<SimulationOutput>,
+    shutdown: Trigger,
 ) {
     let node_id = node.lock().await.get_info().pubkey;
     log::debug!("Started consumer for {}.", node_id);
@@ -429,9 +460,16 @@ async fn consume_events(
         match event {
             SimulationEvent::SendPayment(dest, amt_msat) => {
                 let mut node = node.lock().await;
-                let payment = node.send_payment(dest, amt_msat);
 
-                match payment.await {
+                let mut payment = Payment {
+                    source: node.get_info().pubkey,
+                    hash: None,
+                    amount_msat: amt_msat,
+                    destination: dest,
+                    dispatch_time: SystemTime::now(),
+                };
+
+                let outcome = match node.send_payment(dest, amt_msat).await {
                     Ok(payment_hash) => {
                         log::debug!(
                             "Send payment: {} -> {}: ({}).",
@@ -439,34 +477,34 @@ async fn consume_events(
                             dest,
                             hex::encode(payment_hash.0)
                         );
-
-                        log::debug!("Sending output for {}.", hex::encode(payment_hash.0));
-                        let output = SimulationOutput::PaymentSent(DispatchedPayment {
-                            source: node.get_info().pubkey,
-                            hash: payment_hash,
-                            amount_msat: amt_msat,
-                            destination: dest,
-                            dispatch_time: SystemTime::now(),
-                        });
-
-                        match sender.send(output).await {
-                            Ok(_) => {}
-                            Err(e) => {
-                                log::error!("Error sending simulation output: {:?}.", e);
-                                break;
-                            }
-                        }
+                        // We need to track the payment outcome using the payment hash that we have received.
+                        payment.hash = Some(payment_hash);
+                        SimulationOutput::SendPaymentSuccess(payment)
                     }
                     Err(e) => {
-                        log::error!(
-                            "Error while sending payment {} -> {}. Terminating consumer. {}.",
-                            node_id,
-                            dest,
-                            e
-                        );
-                        break;
+                        log::error!("Error while sending payment {} -> {}.", node_id, dest);
+
+                        match e {
+                            LightningError::PermanentError(s) => {
+                                log::error!("Simulation terminated with error: {s}.");
+                                shutdown.trigger();
+                                break;
+                            }
+                            _ => SimulationOutput::SendPaymentFailure(
+                                payment,
+                                PaymentResult::not_dispatched(),
+                            ),
+                        }
                     }
                 };
+
+                match sender.send(outcome).await {
+                    Ok(_) => {}
+                    Err(e) => {
+                        log::error!("Error sending action outcome: {:?}.", e);
+                        break;
+                    }
+                }
             }
         };
     }
@@ -524,7 +562,7 @@ async fn produce_events(
 }
 
 async fn consume_simulation_results(
-    receiver: Receiver<(DispatchedPayment, PaymentResult)>,
+    receiver: Receiver<(Payment, PaymentResult)>,
     listener: Listener,
     print_batch_size: u32,
 ) {
@@ -538,7 +576,7 @@ async fn consume_simulation_results(
 }
 
 async fn write_payment_results(
-    mut receiver: Receiver<(DispatchedPayment, PaymentResult)>,
+    mut receiver: Receiver<(Payment, PaymentResult)>,
     listener: Listener,
     print_batch_size: u32,
 ) -> Result<(), SimulationError> {
@@ -608,7 +646,7 @@ impl PaymentResultLogger {
         }
     }
 
-    fn report_result(&mut self, details: &DispatchedPayment, result: &PaymentResult) {
+    fn report_result(&mut self, details: &Payment, result: &PaymentResult) {
         match result.payment_outcome {
             PaymentOutcome::Success => self.success_payment += 1,
             _ => self.failed_payment += 1,
@@ -641,7 +679,7 @@ impl PaymentResultLogger {
 async fn produce_simulation_results(
     nodes: HashMap<PublicKey, Arc<Mutex<dyn LightningNode + Send>>>,
     mut output_receiver: Receiver<SimulationOutput>,
-    results: Sender<(DispatchedPayment, PaymentResult)>,
+    results: Sender<(Payment, PaymentResult)>,
     shutdown: Listener,
 ) {
     log::debug!("Simulation results producer started.");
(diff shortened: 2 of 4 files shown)


sim-lib/src/cln.rs
```diff
@@ -96,16 +96,15 @@ impl LightningNode for ClnNode {
...
             .client
             .key_send(KeysendRequest {
                 destination: dest.serialize().to_vec(),
                 amount_msat: Some(Amount { msat: amount_msat }),
                 ..Default::default()
             })
-            .await
-            .map_err(|err| LightningError::SendPaymentError(err.to_string()))?
-            .into_inner();
+            .await?;
```
carlaKC, line 106: So, this is where the different lightning implementations behavior starts to bite us. 
I believe that these two are equivalent: 
* LND API call succeeds, then `TrackPayment` gives us `RouteNotFound`. 
* CLN API call fails with error code 210 "Ran out of routes to try"

While the approach of a zero payment hash _works_, I think that it makes the meaning of a `Ok` result kind of confusing here. 

An alternative could be something like: 
* Define a `LightningError` variant `SendPaymentFailure` that has a `PaymentOutcome` associated with it. 
* Add a new `ActionOutcome` (now `SimulationOutcome` `PaymentFailed` and send it along to `produce_simulation_results` like we do for payments that we could dispatch. 
* Rather than dispatch tracking in `produce_simulation_results`, we can just pipe the result right into the channel because we already know the outcome. 

This is a lot more code, but it does take a more "standard" path for handling the results. Do you think it's worth it?

carlaKC (commented):
Thanks for picking this up! Sorry that I gave you rebase conflicts :')

sim-lib/src/cln.rs
okjodom: > Add a new `ActionOutcome` (now `SimulationOutput`) PaymentFailed and send it along to `produce_simulation_results` like we do for payments that we could dispatch.

We'd need variants of `SimulationOutput` (formerly  `ActionOutcome`), like you propose, but for each `ActivityDefinition` type we add. I expect we'd have variants of `produce_simulation_results` and `consume_simulation_results` that service each `ActivityDefinition` and should confine result reporting to the channels defined by the pair.

> Rather than dispatch tracking in `produce_simulation_results`, we can just pipe the result right into the channel because we already know the outcome.

Looks like a complement to zero payment hash. Since the channel within which we need to stick the result is of type `Sender<(DispatchedPayment, PaymentResult)>`, we still need to have an optional payment hash within the sender, else we'd have to differentiate the channel in which we pipe the results. See https://github.com/bitcoin-dev-project/sim-ln/pull/104/commits/0615f4bc2f7fe4c25e16cf650ec19f61b0100dc0

> Define a `LightningError` variant `SendPaymentFailure` that has a `PaymentOutcome` associated with it.

`PaymentOutcome` is not an error and it's probably no good to carry non-error payloads within errors. Besides, we don't need it

sim-lib/src/lib.rs
```diff
@@ -670,12 +684,17 @@ async fn produce_simulation_results(
...
                         match simulation_output{
-                            SimulationOutput::PaymentSent(dispatched_payment) => {
+                            SimulationOutput::SendPaymentSuccess(dispatched_payment) => {
                                 let source_node = nodes.get(&dispatched_payment.source).unwrap().clone();
                                 set.spawn(track_payment_result(
-                                    source_node,results.clone(),simulation_output, shutdown.clone(),
+                                    source_node,results.clone(),dispatched_payment, shutdown.clone(),
                                 ));
                             },
+                            SimulationOutput::SendPaymentFailure(payment, result) => {
```
carlaKC, line 693: Nice, this is definitely a better way of going about it!

carlaKC (commented):
Approach looks good! 

I do think that we need to take a slightly more subtle approach to handling errors (right now we'll just create failed events for _any_ `send_payment` failure). In my mind there are two types of errors: 
* Terminal errors: the lightning node has shut down -> we can't recover and should also shut down. 
* Payment errors: the API has returned an error, but it's actually just a pathfinding error (eg: no route) -> we can recover from this, and should record the failure and try again.

Right now, if one of the lightning nodes shuts down we'll just continue to track failed payments when we should really just shut down. To address #96, we need to examine the error from CLN and figure out whether it's terminal or now. So basically all of this code as-is, with a little some more precise error handling in the CLN impl.

sim-lib/src/cln.rs
carlaKC: > Looks like a complement to zero payment hash

Yeah indeed, didn't think that one all the way through. All for the zero hash wrapped in the success/failure enum, just don't like having it as the value we based logic on.

okjodom:
> I do think that we need to take a slightly more subtle approach to handling errors (right now we'll just create failed events for _any_ `send_payment` failure). In my mind there are two types of errors:
> 
> * Terminal errors: the lightning node has shut down -> we can't recover and should also shut down.
> * Payment errors: the API has returned an error, but it's actually just a pathfinding error (eg: no route) -> we can recover from this, and should record the failure and try again.

@carlaKC I agree with your sentiment above, but have a slightly different proposal regarding interpretation of Terminal errors

> RE: Terminal errors: the lightning node has shut down -> we can't recover and should also shut down.

Instead of shutting down the entire simulation, I propose we stop simulating the activity or activity set affected by nodes we have seen terminal errors on. This shuts down parts of the simulation, but continues the _still good parts._

> Right now, if one of the lightning nodes shuts down we'll just continue to track failed payments when we should really just shut down. To address #96, we need to examine the error from CLN and figure out whether it's terminal or now. So basically all of this code as-is, with a little some more precise error handling in the CLN impl.

So building on this PR, my proposal is we stop simulating activity related to the CLN node if we observe terminal failure on the node.

okjodom:
> > I do think that we need to take a slightly more subtle approach to handling errors (right now we'll just create failed events for _any_ `send_payment` failure). In my mind there are two types of errors:
> > 
> > * Terminal errors: the lightning node has shut down -> we can't recover and should also shut down.
> > * Payment errors: the API has returned an error, but it's actually just a pathfinding error (eg: no route) -> we can recover from this, and should record the failure and try again.
> 
> @carlaKC I agree with your sentiment above, but have a slightly different proposal regarding interpretation of Terminal errors
> 
> > RE: Terminal errors: the lightning node has shut down -> we can't recover and should also shut down.
> 
> Instead of shutting down the entire simulation, I propose we stop simulating the activity or activity set affected by nodes we have seen terminal errors on. This shuts down parts of the simulation, but continues the _still good parts._
> 
> > Right now, if one of the lightning nodes shuts down we'll just continue to track failed payments when we should really just shut down. To address #96, we need to examine the error from CLN and figure out whether it's terminal or now. So basically all of this code as-is, with a little some more precise error handling in the CLN impl.
> 
> So building on this PR, my proposal is we stop simulating activity related to the CLN node if we observe terminal failure on the node.

Oh, and we terminate the simulation if there's no more activities to simulate.

sim-lib/src/lib.rs
```diff
@@ -685,38 +716,44 @@ async fn produce_simulation_results(
...
+                    log::error!("Track payment failed for {}: {e}.", hex::encode(hash.0));
+                    PaymentResult::default()
                 }
-                Err(e) => log::error!(
-                    "Track payment failed for {}: {e}.",
-                    hex::encode(payment.hash.0)
-                ),
             }
         }
+        None => PaymentResult::default(),
```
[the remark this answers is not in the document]
okjodom: Calling `track_payment_result` this with a None payment hash is a bug! Should never happen
Added a loud complaint :p

> ignore the result

Think reporting some canned result is a good default. Similar to what we do when we fail to track payment (L:743), or when we fail to dispatch a payment and still want to report a result.

sim-lib/src/lib.rs
```diff
@@ -219,6 +219,8 @@ enum SimulationEvent {
 
 /// SimulationOutput provides the output of a simulation event.
 enum SimulationOutput {
+    /// Terminal error to signal simulation exit
+    TerminalError(SimulationError),
```
okjodom, line 223: Implication:
- any task or method call should send this `SimulationOutput` to cleanly terminate the simulation
- in case sending this specific output to the channel is not possible, `shutdown` trigger should be used, to the same effect but without the final log of why the simulation terminated.

sim-lib/src/lib.rs
```diff
@@ -478,13 +482,22 @@ async fn consume_events(
...
+                        log::error!("Error while sending payment {} -> {}", node_id, dest);
+
+                        match e {
+                            LightningError::LndRpcError(_) => {
+                                // TODO: Parse LND rpc status to determine if it's a terminal error or not.
+                                SimulationOutput::TerminalError(SimulationError::LightningError(e))
+                            }
+                            LightningError::ClnRpcError(_) => {
+                                // TODO: Parse CLN rpc status to determine if it's a terminal error or not.
+                                SimulationOutput::TerminalError(SimulationError::LightningError(e))
```
carlaKC, line 494: This is a bit of a layering violation to me, since in practice the code we'll fill in for the `TODO` will likely be some messy implementation-specific string parsing / error code soup. 

What do you think about defining a `LightningError::TemporaryRPCError` that both implementations return from `send_payment` so that the parsing can live in the implementation? 

Then this handling is just: 
- If `TemporaryRPCError` -> send `SendPaymentFailure`
- else -> `TerminalError`

sim-lib/src/lib.rs
```diff
@@ -679,15 +693,19 @@ async fn produce_simulation_results(
     loop {
         tokio::select! {
             biased;
-            _ = shutdown.clone() => break,
+            _ = listener.clone() => break,
             output = output_receiver.recv() => {
                 match output{
                     Some(simulation_output) => {
                         match simulation_output{
+                            SimulationOutput::TerminalError(e) => {
```
carlaKC, line 701: What's the advantage of propagating the `TerminalError` down to this results loop vs exiting in `consume_results`? 

Trying to think about where we'd want this to be handled in the case where we're going to do "smarter" handling of errors (eg, if we wanted to add a feature to allow some activities to die but others continue, this feels like one stage too deep to handle that case?). Difficult to think about it before we've done it :')

carlaKC (commented):
Just read through the shutdown strategy commit!

sim-lib/src/lib.rs
okjodom: > What's the advantage of propagating the `TerminalError` down to this results loop vs exiting in `consume_results`?

If you mean exiting in `consume_events` and not `consume_simulation_results`, I think it's quite simple and direct, and will give us the best chance of stopping specific activities on terminal error

sim-lib/src/lib.rs
okjodom: Adopted by defining `LightningError:TerminalError(String)`. Any other error we receive is assumed temporary, and Lightning rpc caller chooses how to proceed

carlaKC:
https://github.com/bitcoin-dev-project/sim-ln/pull/104/commits/f89506abbfba4875d4aa78da79b4bc0c68ee9536 looking nice 👌 thanks for going through a few iterations to get it there!

I you want to go ahead and fill in the parsing I think we can do this all in one PR?

okjodom:
> https://github.com/bitcoin-dev-project/sim-ln/pull/104/commits/f89506abbfba4875d4aa78da79b4bc0c68ee9536 looking nice 👌 thanks for going through a few iterations to get it there!
> 
> I you want to go ahead and fill in the parsing I think we can do this all in one PR? 

I was planning the parsing as follow-ups when I get a chance to build on this.

carlaKC:
> I was planning the parsing as follow-ups when I get a chance to build on this.

SGTM, no strong feelings. 


> Starting w cln because that's the most offending

GLHF :')

sim-lib/src/lib.rs
```diff
@@ -155,6 +157,15 @@ impl Display for PaymentResult {
...
 
+impl Default for PaymentResult {
+    fn default() -> Self {
+        PaymentResult {
+            htlc_count: 0,
+            payment_outcome: PaymentOutcome::Unknown,
+        }
+    }
+}
+
```
sr-gi, line 158: I don't think we should implement `Default` here, but a constrictor called `unknown` or something along those lines. The main reasoning is that we are not using default here as it is normally intended, we're just calling `PaymentResult::default()` in some places of the codebase. Plus the default being an unknown result is a bit weird design-wise

sim-lib/src/cln.rs
```diff
@@ -103,7 +103,11 @@ impl LightningNode for ClnNode {
                 amount_msat: Some(Amount { msat: amount_msat }),
                 ..Default::default()
             })
-            .await?;
+            .await
+            .map_err(|s| {
+                // TODO: Parse CLN rpc status to determine if it's a terminal error or not.
+                LightningError::TerminalError(format!("{:?}", s))
```
sr-gi: What is a `TerminalError`? Do you mean `PermanentError`?

sim-lib/src/cln.rs
```diff
@@ -105,8 +105,16 @@ impl LightningNode for ClnNode {
...
+                let message = s.message();
+                // REF: https://docs.corelightning.org/reference/lightning-keysend#return-value
+
+                // Error codes -1 and 203 indicate terminal errors
+                if message.contains("Some(-1") | message.contains("Some(203") {
+                    return LightningError::TerminalError(format!("{:?}", message));
+                }
+
+                // Error codes, 205, 206 and 210 indicate temporary errors that can be retried
+                LightningError::SendPaymentError(format!("{:?}", message))
```
sr-gi, line 117: There's no need to return here, given you're evaluating the result with `?` later.

```suggestion
// Error codes -1 and 203 indicate terminal errors
if message.contains("Some(-1") | message.contains("Some(203") {
    LightningError::TerminalError(format!("{:?}", message))
} else {
    // Error codes, 205, 206, and 210 indicate temporary errors that can be retried
    LightningError::SendPaymentError(format!("{:?}", message))
}
```

sr-gi (changes requested):
Looks pretty good overall. Check some comments inline.

f5e9eaa94841a72f80c26f947b795477c66ef98b and 5bae8768212d44906a240d5b6b9b7d5f75d6614a should be squashable with f89506abbfba4875d4aa78da79b4bc0c68ee9536

carlaKC (commented):
tACK against the error originally described in #96 
```
 ERROR [sim_lib] Error while sending payment 0327b7826a232ba4c7032f659ae0b65528f9b9f3191c54cbea423a207cff150029 -> 0346b6eeeaa8f13a64e4bff924a2cb9db94e01b3cc1819970c1f3e8cb4eabb3f5f
```

Simulation output: 
```
0327b7826a232ba4c7032f659ae0b65528f9b9f3191c54cbea423a207cff150029,0346b6eeeaa8f13a64e4bff924a2cb9db94e01b3cc1819970c1f3e8cb4eabb3f5f,2000,Unknown,1696256242579,0,Unknown
```

LGTM pending @sr-gi's comments. Nothing blocking in my review except for [shutting down if we can't send a payment result ](https://github.com/bitcoin-dev-project/sim-ln/pull/104#discussion_r1342734459) in `produce_simulation_results` (I'm also happy to pick up in followup, as shutdown hygiene needs some general work IMO).

sim-lib/src/lib.rs
```diff
@@ -670,12 +684,17 @@ async fn produce_simulation_results(
...
                                 let source_node = nodes.get(&dispatched_payment.source).unwrap().clone();
                                 set.spawn(track_payment_result(
-                                    source_node,results.clone(),simulation_output, shutdown.clone(),
+                                    source_node,results.clone(),dispatched_payment, shutdown.clone(),
                                 ));
                             },
+                            SimulationOutput::SendPaymentFailure(payment, result) => {
+                                if results.clone().send((payment, result)).await.is_err() {
+                                    log::debug!("Could not send payment result");
+                                }
```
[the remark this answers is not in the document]
okjodom: made the same note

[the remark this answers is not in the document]
okjodom: > This is pre-existing, but I think we should break if we can't send into the results channel? Because something has exit downstream.

I think we should only break out of these listeners if shutdown is called. If something exits with finality, it should trigger shutdown, which leads to break(s), which can lead to channel send fail (because we bias to shutdowns?), but we log these send failures like we do here

sim-lib/src/lib.rs
```diff
@@ -155,6 +157,15 @@ impl Display for PaymentResult {
     }
 }
 
+impl Default for PaymentResult {
+    fn default() -> Self {
+        PaymentResult {
+            htlc_count: 0,
+            payment_outcome: PaymentOutcome::Unknown,
```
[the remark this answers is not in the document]
okjodom: latest revision defines two explicit payment outcomes `NotDispatched` and `TrackPaymentFailed` which we use in case of temporary failures. Related to @sr-gi's comment on not implementing `Default`, we don't need a Default

sim-lib/src/lib.rs
```diff
@@ -658,8 +672,6 @@ async fn produce_simulation_results(
                         match simulation_output{
                             SimulationOutput::PaymentSent(dispatched_payment) => {
                                 let source_node = nodes.get(&dispatched_payment.source).unwrap().clone();
-
-                                log::debug!("Tracking payment result for: {}.", hex::encode(dispatched_payment.hash.0));
```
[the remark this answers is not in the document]
okjodom: It was redundant because we have similar logs in `track_payment_result`

sim-lib/src/lib.rs
```diff
@@ -167,25 +159,33 @@ pub enum PaymentOutcome {
...
 }
 
-impl Display for DispatchedPayment {
+impl Display for Payment {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
             "Payment {} dispatched at {:?} sending {} msat from {} -> {}",
-            hex::encode(self.hash.0),
+            hex::encode(self.hash.map(|h| hex::encode(h.0)).unwrap_or(String::new())),
```
sr-gi, line 188: This is still redundant. You're hex-encoding it twice.

```
self.hash.map(|h| hex::encode(h.0)).unwrap_or(String::new()),
```

sim-lib/src/lib.rs
```diff
@@ -429,44 +444,54 @@ async fn consume_events(
...
+                                log::error!("Simulation terminated with error: {s}.");
+                                shutdown.trigger();
                                 break;
                             }
+                            _ => SimulationOutput::SendPaymentFailure(
+                                payment,
+                                PaymentResult {
+                                    htlc_count: 0,
+                                    payment_outcome: PaymentOutcome::NotDispatched,
+                                },
```
sr-gi, line 482: I think it may be worth creating a couple of helper constructors to reduce the boilerplate here:

```
PaymentResult::not_dispatched();
PaymentResult::tracked_payment_failed();
```

sim-lib/src/lib.rs
```diff
@@ -656,14 +681,17 @@ async fn produce_simulation_results(
...
                         match simulation_output{
-                            SimulationOutput::PaymentSent(dispatched_payment) => {
-                                let source_node = nodes.get(&dispatched_payment.source).unwrap().clone();
-
-                                log::debug!("Tracking payment result for: {}.", hex::encode(dispatched_payment.hash.0));
+                            SimulationOutput::SendPaymentSuccess(payment) => {
+                                let source_node = nodes.get(&payment.source).unwrap().clone();
                                 set.spawn(track_payment_result(
-                                    source_node,results.clone(),simulation_output, shutdown.clone(),
+                                    source_node,results.clone(),payment, shutdown.clone(),
```
sr-gi, line 687: Not sure why `fmt` is not catching this, but there are some spaces missing here.

sr-gi (commented):
A few comments, mostly nits, LGTM otherwise

sim-lib/src/lib.rs
okjodom: oh I didn't catch it on previous pass

sim-lib/src/lib.rs
carlaKC: `fmt` is unreliable in `select` :(

okjodom: hm

carlaKC (approved):
tACK https://github.com/bitcoin-dev-project/sim-ln/pull/104/commits/fb981a256d906037566a381eb3c433602136c18c 🎉

sr-gi:
ACK [fb981a2](https://github.com/bitcoin-dev-project/sim-ln/pull/104/commits/fb981a256d906037566a381eb3c433602136c18c)

sr-gi (approved):

PR #105, opened 2023-09-19
refactor: Standardize names and add Arch Doc

carlaKC:
Before we run through another round of changes, let's standardize naming a little.

### DIFF
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -131,8 +131,10 @@ pub trait LightningNode {
     async fn get_node_features(&mut self, node: PublicKey) -> Result<NodeFeatures, LightningError>;
 }
 
+/// SimulationEvent describes the set of actions that the simulator can run on nodes that it has execution permissions 
+/// on.
 #[derive(Clone, Copy)]
-enum NodeAction {
+enum SimulationEvent {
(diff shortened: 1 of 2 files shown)


sim-lib/src/lib.rs
```diff
@@ -131,8 +131,9 @@ pub trait LightningNode {
     async fn get_node_features(&mut self, node: PublicKey) -> Result<NodeFeatures, LightningError>;
 }
 
+/// SimulationEvent describes the set of actions that the simulator can excute on nodes that
```
okjodom: nit: incomplete comment?

carlaKC: Whoops, [pushed fixup](https://github.com/bitcoin-dev-project/sim-ln/compare/59a3887432b937aceffe9a0f2a537c3cab547a95..73591972c3e805f7e87605d9c7cb1ff118249eeb)

okjodom:
Wish the arch doc included

- input of simulation events are input as activities. We can add to this when we have other activity types or define new ways of generating activities
- detail on how we support multiple-implementations of LN, and hinting at what it would take to add support for any other implementation

The rest looks good, i.e core arch of the simulator and how it execs events on nodes, tracks results and reports them

okjodom (approved):
naming lgtm

Issue #29, opened 2023-08-17
Robustness: Validate node networks and disallow mainnet


carlaKC:
When we're given a set of nodes, we should validate that they're all on the same network. We don't need to worry about destinations / activity, since we check this against the graph but it's _possible_ that somebody accidentally configures nodes on different networks.

We should also completely disallow mainnet while we're here, since this is a testing tool that slings money around with abandon.

Extheoisah:
sure 😁

PR #112, opened 2023-09-21
chore: validate node network and disallow mainnet

Extheoisah:
closes https://github.com/bitcoin-dev-project/sim-ln/issues/29

### DIFF
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -108,6 +108,7 @@ pub struct NodeInfo {
     pub pubkey: PublicKey,
     pub alias: String,
     pub features: NodeFeatures,
+    pub network: String,
 }
 
 /// LightningNode represents the functionality that is required to execute events on a lightning node.
@@ -275,6 +276,41 @@ impl Simulation {
         Ok(())
     }
 
+    // it validates that the nodes are all on the same network and ensures that we're not running on mainnet.
+    async fn validate_node_network(&self) -> Result<(), LightningError> {
+        let mut valid_network = Option::None;
+        let unsupported_networks: HashSet<String> = vec!["mainnet", "bitcoin"]
+            .into_iter()
+            .map(|s| s.to_string())
+            .collect();
+
+        for node in self.nodes.values() {
+            let network = node.lock().await.get_info().network.clone();
+            if unsupported_networks.contains(&network) {
+                return Err(LightningError::ValidationError(
+                    "mainnet is not supported".to_string(),
+                ));
+            }
+
+            valid_network = valid_network.take().or_else(|| Some(network.clone()));
+            if let Some(valid) = &valid_network {
+                if valid != &network {
+                    return Err(LightningError::ValidationError(format!(
+                        "nodes are not on the same network {}.",
+                        network,
+                    )));
+                }
+            }
+        }
+
+        log::info!(
+            "Simulation is running on the network: {}.",
+            valid_network.as_ref().unwrap()
+        );
+
+        Ok(())
+    }
+
     pub async fn run(&self) -> Result<(), SimulationError> {
         if let Some(total_time) = self.total_time {
             log::info!("Running the simulation for {}s.", total_time.as_secs());
@@ -282,6 +318,7 @@ impl Simulation {
             log::info!("Running the simulation forever.");
         }
 
+        self.validate_node_network().await?;
         self.validate_activity().await?;
 
         log::info!(
--- a/sim-lib/src/lnd.rs
+++ b/sim-lib/src/lnd.rs
@@ -51,6 +51,7 @@ impl LndNode {
             identity_pubkey,
             features,
             alias,
+            chains,
             ..
         } = client
             .lightning()
@@ -59,6 +60,12 @@ impl LndNode {
             .map_err(|err| LightningError::GetInfoError(err.to_string()))?
             .into_inner();
 
+        if chains.len() != 1 {
+            return Err(LightningError::GetInfoError(
+                "LND node is not connected to a single chain".to_string(),
+            ));
+        }
+
         Ok(Self {
             client,
             info: NodeInfo {
@@ -66,6 +73,7 @@ impl LndNode {
                     .map_err(|err| LightningError::GetInfoError(err.to_string()))?,
                 features: parse_node_features(features.keys().cloned().collect()),
                 alias,
+                network: chains[0].network.clone(),
             },
         })
     }
--- a/sim-lib/src/cln.rs
+++ b/sim-lib/src/cln.rs
@@ -59,6 +59,7 @@ impl ClnNode {
             id,
             alias,
             our_features,
+            network,
             ..
         } = client
             .getinfo(GetinfoRequest {})
@@ -80,6 +81,7 @@ impl ClnNode {
                     .map_err(|err| LightningError::GetInfoError(err.to_string()))?,
                 features,
                 alias: alias.unwrap_or("".to_string()),
+                network,
             },
         })
     }


sim-lib/src/lib.rs
```diff
@@ -108,6 +109,7 @@ pub struct NodeInfo {
     pub pubkey: PublicKey,
     pub alias: String,
     pub features: NodeFeatures,
+    pub networks: Vec<String>,
```
okjodom, line 112: node info should have singular `network`

sim-lib/src/lib.rs
```diff
@@ -275,6 +277,44 @@ impl Simulation {
         Ok(())
     }
 
+    // validate_node_network validates that the nodes that are on the same bitcoin network.
+    // e.g. node A and B are on testnet, node C is on mainnet should return an error.
+    async fn validate_node_network(&self) -> Result<(), LightningError> {
+        let mut networks = Vec::new();
+        let mainnet = Network::Bitcoin.to_string().to_lowercase();
+
+        for node in self.nodes.values() {
```
okjodom, line 286: if each node info declares just one `network`, maybe you could replace the validation logic below with something like

```rs

let mainnet = Network::Bitcoin.to_string().to_lowercase();
let mut valid_network = Option::None;

for node in self.nodes.values() {
    let network = node.lock().await.get_info().network;

    if network == mainnet {
        return Err(LightningError::ValidationError(
            "mainnet is not supported".to_string(),
        ));
    }

    let valid_replaced = valid_network.replace(network.clone());

    if valid_replaced.is_some() && valid_replaced != network {
        return Err(LightningError::ValidationError(format!(
            "nodes are not on the same network {}",
            network,
        )));
    }
}

Ok(())
```

sim-lib/src/lib.rs
```diff
@@ -283,6 +323,7 @@ impl Simulation {
         }
 
         self.validate_activity().await?;
+        self.validate_node_network().await?;
```
okjodom, line 326: prefer we validate network before activities :)

sim-lib/src/lnd.rs
```diff
@@ -66,6 +67,10 @@ impl LndNode {
                     .map_err(|err| LightningError::GetInfoError(err.to_string()))?,
                 features: parse_node_features(features.keys().cloned().collect()),
                 alias,
+                networks: chains
```
okjodom, line 70: so CLN advertises just a single network while LND can advertise multiple? 🤦🏿

Extheoisah: Yes. And its why i made it a vec

sim-lib/src/lib.rs
```diff
@@ -326,7 +367,7 @@ impl Simulation {
         let mut success = true;
         while let Some(res) = tasks.join_next().await {
             if let Err(e) = res {
-                log::error!("Task exited with error: {e}.");
+                log::error!("Task exited with error: {}.", e);
```
carlaKC, line 370: Does this need to change?

sim-lib/src/lnd.rs
carlaKC: FUN LIGHTNING IS FUN OKAY.

carlaKC (commented):
🙌 welcome to sim-ln 🙌 

Can be [simplified](https://github.com/bitcoin-dev-project/sim-ln/pull/112/files#r1334503033) by just assuming one network, but looking good!

sim-lib/src/lib.rs
Extheoisah: Not really. The syntax Just didn't look right to me.

Extheoisah:
comments addressed! @carlaKC @okjodom

sim-lib/src/lib.rs
```diff
@@ -275,13 +277,48 @@ impl Simulation {
...
+                        network,
+                    )));
+                }
+            } else {
+                valid_network = Some(network);
+            }
+        }
+
+        log::info!(
+            "Simulation is running on the network: {}",
```
carlaKC, line 307: nit: end logs with a period `.`

sim-lib/src/lib.rs
carlaKC: I think this is pretty canonical rust so let's leave as is?

sim-lib/src/lib.rs
```diff
@@ -275,13 +277,48 @@ impl Simulation {
...
+            if let Some(valid) = &valid_network {
+                if valid != &network {
+                    return Err(LightningError::ValidationError(format!(
+                        "nodes are not on the same network {}",
+                        network,
+                    )));
+                }
+            } else {
+                valid_network = Some(network);
+            }
```
carlaKC, line 303: We could shorten this a little with `valid_network = valid_network.unwrap_or(network.clone());` on 294, lets us get rid of the conditional / fill our own value if it's currently none.

sim-lib/src/lib.rs
```diff
@@ -275,13 +277,48 @@ impl Simulation {
         Ok(())
     }
 
+    // validate_node_network validates that the nodes that are on the same bitcoin network.
+    // e.g. node A and B are on testnet, node C is on mainnet should return an error.
+    async fn validate_node_network(&self) -> Result<(), LightningError> {
+        let mut valid_network = Option::None;
+        let mainnet = Network::Bitcoin.to_string().to_lowercase();
```
carlaKC, line 284: This string is "bitcoin" which won't match the `network` set by LND "mainnet" (there's divergence between LDK and LND in naming).

carlaKC (commented):
Looking good!  Successfully tested 2/3 validation checks: 

Multiple chains: 
```
Error: Get info error: LND node is not connected to a single chain
```

Different networks:
```
Error: Lightning Error: ValidationError("nodes are not on the same network regtest")

Caused by:
    Config validation failed: nodes are not on the same network regtest
```

Disallowing mainnet needs an update for LND (we're checking `bitcoin` but LND sets `mainnet` as its network value). Will test CLN on next pass, but looks straightforward.

sim-lib/src/lib.rs
Extheoisah: Yes, an oversight on my part. fixed here -> https://github.com/bitcoin-dev-project/sim-ln/pull/112/commits/5d2bfac7f1a9d20630899b10602bdcf8cae38b75

sim-lib/src/lib.rs
```diff
@@ -275,13 +276,50 @@ impl Simulation {
...
+                return Err(LightningError::ValidationError(
+                    "mainnet is not supported".to_string(),
+                ));
+            }
+
+            valid_network = valid_network.take().or_else(|| Some(network.clone()));
+            if let Some(valid) = &valid_network {
+                if valid != &network {
+                    return Err(LightningError::ValidationError(format!(
+                        "nodes are not on the same network {}",
```
carlaKC, line 300: nittynit: "not on the same network: {}"

sim-lib/src/lib.rs
```diff
@@ -275,13 +276,50 @@ impl Simulation {
...
     }
 
+    // validate_node_network validates that the nodes that are on the same bitcoin network.
+    // e.g. node A and B are on testnet, node C is on mainnet should return an error.
+    async fn validate_node_network(&self) -> Result<(), LightningError> {
+        let mut valid_network = Option::None;
+        let unsupported_networks: HashSet<String> = vec!["mainnet", "bitcoin"]
+            .into_iter()
+            .map(|s| s.to_string())
+            .collect();
```
carlaKC, line 286: Fewer LOC to just declare `mut HashSet` then `insert`, but no strong feelings.

sim-lib/src/lib.rs
```diff
@@ -275,13 +276,50 @@ impl Simulation {
         Ok(())
     }
 
+    // validate_node_network validates that the nodes that are on the same bitcoin network.
+    // e.g. node A and B are on testnet, node C is on mainnet should return an error.
```
carlaKC, line 280: Comment nitpick: it validates that the nodes are all on the same network and ensures that we're not running on mainnet.

carlaKC (commented):
tACK!
```
Error: Lightning Error: ValidationError("mainnet is not supported")

Caused by:
    Config validation failed: mainnet is not supported
```

I think you can go ahead and squash this all into a single commit then we can merge. Left some last nits, but none of them are blocking.

carlaKC (approved):
LGTM!

okjodom (approved):

sim-lib/src/lnd.rs
```diff
@@ -59,13 +60,20 @@ impl LndNode {
             .map_err(|err| LightningError::GetInfoError(err.to_string()))?
             .into_inner();
 
+        if chains.len() != 1 {
```
sr-gi, line 63: nit: we can use `is_empty` here to be more idiomatic

sim-lib/src/lnd.rs
```diff
@@ -59,13 +60,20 @@ impl LndNode {
...
+        }
+
         Ok(Self {
             client,
             info: NodeInfo {
                 pubkey: PublicKey::from_str(&identity_pubkey)
                     .map_err(|err| LightningError::GetInfoError(err.to_string()))?,
                 features: parse_node_features(features.keys().cloned().collect()),
                 alias,
+                network: chains[0].network.clone(),
```
sr-gi, line 76: nit: we can use `chains.swap_remove(0).network` here to avoid unnecessarily copying a value.

sr-gi (commented):
Post-merge review.

I'm looking at this given I'm trying to rebase for #103 and I just realized the way this is handled is troublesome :(

From #103 onwards, we are using `NodeInfo` to store information about all nodes in our config, not only the ones we control. However, we cannot query the network of those.

Furthermore, once we've checked the network the nodes we're controlling are running at, there is no need to store the `network` in `NodeInfo`, given we're not using that anymore. Hence, we should just add a method to query the `network` instead of making this part of `get_info`

PR #114, opened 2023-09-28
chore: add docker

Extheoisah:
To run:
```
CONFIG_PATH=/path-to-project/sim-ln/config.json docker-compose run sim-cli
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,31 @@
+LOG_LEVEL ?= info
+
+build-docker:
+	chmod +x ./docker/build.sh && ./docker/build.sh
+
+mount-volume:
+	chmod +x ./docker/setup-volume.sh && ./docker/setup-volume.sh "$(SIMFILE_PATH)"
+
+help:
+	@echo "Makefile commands:"
+	@echo "build-docker      Builds the Docker image."
+	@echo "mount-volume      Mounts the specified volume."
+	@echo "run               Runs the Docker container."
+	@echo "run-interactive   Runs the Docker container in interactive mode."
+	@echo "stop              Stops the Docker container."
+	@echo ""
+	@echo "Variables:"
+	@echo "SIMFILE_PATH       Path to the sim.json file."
+	@echo "LOG_LEVEL         Set the logging level (default: info) e.g. <make run LOG_LEVEL=debug>"
+	@echo "HELP              Set to true to print the help message (default: false) e.g. <make run HELP=true>"
+	@echo "PRINT_BATCH_SIZE  Set the batch size for printing the results e.g. <make run PRINT_BATCH_SIZE=100>"
+	@echo "TOTAL_TIME        Set the total time for the simulation e.g. <make run TOTAL_TIME=1000>"
+
+run-docker:
+	docker run -d --rm --name sim-ln --init -v simln-data:/data -e SIMFILE_PATH=/data/sim.json -e LOG_LEVEL=$(LOG_LEVEL) -e HELP=${HELP} -e PRINT_BATCH_SIZE=${PRINT_BATCH_SIZE} -e TOTAL_TIME=${TOTAL_TIME} sim-ln
+
+run-interactive:
+	docker run --rm --name sim-ln --init -v simln-data:/data -e SIMFILE_PATH=/data/sim.json -it sim-ln
+
+stop-docker:
+	docker stop sim-ln
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+# Define the start command
+START_COMMAND="/usr/local/bin/sim-cli $SIMFILE_PATH"
+
+# Check for version arg
+if [[ ! -z ${VERSION} ]]; then
+    START_COMMAND="$START_COMMAND --version"
+fi
+
+# Check for help arg
+if [[ ! -z ${HELP} ]]; then
+    START_COMMAND="$START_COMMAND --help"
+fi
+
+# Check for log level arg
+if [[ ! -z ${LOG_LEVEL} ]]; then
+    START_COMMAND="$START_COMMAND --log-level $LOG_LEVEL"
+fi
+
+# Check for total time arg
+if [[ ! -z ${TOTAL_TIME} ]]; then
+    START_COMMAND="$START_COMMAND --total-time $TOTAL_TIME"
+fi
+
+# Check for print-batch-size arg
+if [[ ! -z ${PRINT_BATCH_SIZE} ]]; then
+    START_COMMAND="$START_COMMAND --print-batch-size $PRINT_BATCH_SIZE"
+fi
+
+# start the container
+exec $START_COMMAND
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+*.log
+target/
+.git/
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /target
 node_modules
 config.json
+sim.json
 package-lock.json
 activity-generator/releases/*
 .DS_Store
--- a/README.md
+++ b/README.md
@@ -180,3 +180,6 @@ If you're looking to get started with local lightning development, we
 recommend [polar](https://lightningpolar.com/). For larger deployments, 
 see the [Scaling Lightning](https://github.com/scaling-lightning/scaling-lightning) 
 project.
+
+## Docker
+If you want to run the cli in a containerized environment, see the docker set up docs [here](./docker/README.md)
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,34 @@
+# Use the rust image as the base image for the build stage
+FROM rust:latest AS builder
+
+# Receive architecture as an argument
+ARG TARGET_ARCH
+
+# Copy the source code
+COPY . /tmp/sim-ln
+
+# Install the dependencies required for building sim-ln
+RUN apt-get update \
+    && apt-get -y --no-install-recommends install protobuf-compiler musl-tools
+
+RUN cd /tmp/sim-ln \
+    && rustup target add ${TARGET_ARCH} \
+    && rustup component add rustfmt \
+    && RUSTFLAGS='-C target-feature=+crt-static' cargo build --locked --release --target=${TARGET_ARCH}
+
+# Use a new stage with a smaller base image to reduce image size
+FROM alpine:latest
+
+ARG TARGET_ARCH
+
+RUN apk update && apk upgrade
+
+# Copy the sim-cli binaries from the build stage to the new stage
+COPY --from=builder /tmp/sim-ln/target/${TARGET_ARCH}/release/sim-cli /usr/local/bin/
+
+# Copy the entrypoint script to the container
+COPY docker/entrypoint.sh /entrypoint.sh
+
+RUN chmod +x entrypoint.sh
+
+ENTRYPOINT [ "/entrypoint.sh" ]
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,121 @@
+# sim-ln Docker Setup
+
+This guide provides instructions on how to build, run, and manage the `sim-ln` Docker container using the provided Makefile.
+
+## Prerequisites
+
+- Docker installed on your machine.
+- A `sim.json` file that contains your specific configuration.
+
+## Building the Docker Image
+
+To build the Docker image, execute:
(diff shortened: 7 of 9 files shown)


Dockerfile
sr-gi, line 1: I've tweaked this a bit based on https://github.com/talaia-labs/rust-teos/blob/master/docker/Dockerfile:

```
# Use the rust image as the base image for the build stage
FROM rust:latest AS builder

# Copy the rust-teos source code
COPY . /tmp/sim-ln

# Install the dependencies required for building sim-ln
RUN apt-get update \
    && apt-get -y --no-install-recommends install protobuf-compiler musl-tools

RUN cd /tmp/sim-ln \
    && rustup target add x86_64-unknown-linux-musl \
    # Rustfmt is needed to format the grpc stubs generated by tonic
    && rustup component add rustfmt \
    # Cross compile with musl as the target, so teosd can run on alpine
    && RUSTFLAGS='-C target-feature=+crt-static' cargo build --locked --release --target x86_64-unknown-linux-musl

# Use a new stage with a smaller base image to reduce image size
FROM alpine:latest

RUN apk update && apk upgrade

# UID and GID for the teosd user
ENV SIMLN_UID=1001 SIMLN_GID=1001

# Copy the sim-cli binaries from the build stage to the new stage
COPY --from=builder /tmp/sim-ln/target/x86_64-unknown-linux-musl/release/sim-cli /usr/local/bin/

CMD ["sim-cli", "--config", "config.json"]
```

sr-gi: Notice this won't work atm given there's still an issue with how the config file is passed to the container

sr-gi: Also it may be worth to add an `entrypoint` script to the execution of the cli can be split to add different modifiers, such as the debug log or run time. I'd suggest you check: https://github.com/talaia-labs/rust-teos/tree/master/docker

sr-gi (changes requested):
It may be worth moving all the docker-related stuff to a folder called `docker` so we can also add docs on how to run it.

docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: '3'
```
carlaKC, line 1: I think that we can live without docker compose? Since the use case that we primarily forsee is simln dropping into other people's existing setups (they'll add it to _their_ docker compose. 

For including a volume we can just add instructions to include a volume with docker run?

docker/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+# Define the start command
+START_COMMAND="/usr/local/bin/sim-cli"
+
+# Check for a custom configuration path
+if [[ ! -z ${CONFIG_PATH} ]]; then
+    START_COMMAND="$START_COMMAND --config $CONFIG_PATH"
```
carlaKC, line 8: Do we need to parse each config option? 

Looking [here](https://refine.dev/blog/docker-entrypoint/#difference-between-entrypoint-and-cmd): "Whatever you provide through entrypoint actually appends to the existing entrypoint instruction in the dockerfile", we should be able to just specify any CLI flags and they'll be passed straight through if we make our entrypoint `sim-cli`.

Makefile
```diff
@@ -0,0 +1,5 @@
+start-deps:
+	chmod +x ./docker/build.sh && ./docker/build.sh
+
+start:
+	docker run -it -e ${ARGS} sim-ln:tag
```
carlaKC, line 5: One thing that I'm missing here is the ability to include a volume that contains the certs/macaroons etc that we'll need to read to talk to the node. 

If our config provides local filesystem paths then sim-cli tries to read them from within the docker file it won't find them.  I think that a simple approach could be to require that the user mounts a single volume that contains the macaroons etc and the config.json and then we look for everything in there. 

Pretty brutal to be dealing with these IPs/paths in the docker context.

carlaKC (commented):
This is going to be _way_ trickier that I thought to do in a user-friendly way. 

Specifically because we have a config because the paths/IPs need to be relative to code executing _in_ the container. 

I'd suggest that we add some docs around this outlining: 
* That you need to mount a volume when you `docker run` which includes all the "artifacts" you need. 
* That the paths in `config.json` should be expressed _relative_ to the file structure in the volume. 
- That the IP addresses need to be _docker_ IP addresses (not localhost).

Makefile
```diff
@@ -0,0 +1,5 @@
+start-deps:
```
carlaKC, line 1: also nit that this should probably be `build-docker` and `run-docker` below.

okjodom:
> How would this work if I want to run with scenario_a.json, followed by scenario_b.json, would I need to re-deploy?

User would need to add all `scenario_*.json` files to the data dir. Re-reploy is just adding `scenario_b.json` to the data dir and then running `sim-cli` again, this time selecting `scenario_b.json` as the sim file.

Extheoisah:
> > How would this work if I want to run with scenario_a.json, followed by scenario_b.json, would I need to re-deploy?
> 
> User would need to add all `scenario_*.json` files to the data dir. Re-reploy is just adding `scenario_b.json` to the data dir and then running `sim-cli` again, this time selecting `scenario_b.json` as the sim file.

Hmm.... I'm not sure about this flow. Its a bit tacky to me except If I don't fully grasp it

Makefile
```diff
@@ -0,0 +1,30 @@
...
+	@echo "run-interactive   Runs the Docker container in interactive mode."
+	@echo "stop              Stops the Docker container."
+	@echo ""
+	@echo "Variables:"
+	@echo "CONFIG_PATH       Path to the config.json file."
+	@echo "LOG_LEVEL         Set the logging level (default: info) e.g. <make run LOG_LEVEL=debug>"
+	@echo "HELP              Set to true to print the help message (default: false) e.g. <make run HELP=true>"
+	@echo "PRINT_BATCH_SIZE  Set the batch size for printing the results (default: 1000) e.g. <make run PRINT_BATCH_SIZE=100>"
+
+run:
```
carlaKC, line 23: nit: make these `docker-run` and `docker-stop`, since we may want to use `run` for `cargo run`.

Makefile
```diff
@@ -0,0 +1,30 @@
...
+	docker run --rm --name sim-ln --init -v simln-config:/data -e CONFIG_PATH=/data/config.json -e LOG_LEVEL=$(LOG_LEVEL) -e HELP=${HELP} -e PRINT_BATCH_SIZE=${PRINT_BATCH_SIZE} -e TOTAL_TIME=${TOTAL_TIME} sim-ln
```
carlaKC, line 24: `docker run -d` so that this runs fully in the background?

docker/docker.md
```diff
@@ -0,0 +1,87 @@
...
+
+```bash
+make build-docker
+```
+
+This command will make the necessary preparations and build the Docker image.
+
+## Mounting the Volume
+
+To ensure that the Docker container can access the configuration and other necessary data, we use Docker volumes. Before you can run the container, you need to set up the volume with the necessary data.
```
carlaKC, line 22: "other necessary data" -> "authentication credentials"

docker/docker.md
```diff
@@ -0,0 +1,87 @@
...
+
+To do this:
```
carlaKC, line 24: I think that it would be helpful to note that we've provided an easy way to do this, but you can also do it yourself (eg in the case where the macaroons etc are on another machine). 

Something along the lines of: 
- If you're running nodes locally -> you can use this script, it'll copy everything and create a config suitable for the container. 
- If you're running nodes remotely, you'll need to provide your own volume and config that points to the files location *in the volume*. 

Can also add a note about IP addresses here.

docker/docker.md
```diff
@@ -0,0 +1,87 @@
...
+
+```bash
+make run LOG_LEVEL=debug
+```
+
+Other configurable variables include:
+
+- `HELP`: Set to `true` to print the help message.
+- `PRINT_BATCH_SIZE`: Set the batch size for printing the results.
+- `TOTAL_TIME`: Adjusts the total time.
```
carlaKC, line 52: nits on the variable docs: 
- `PRINT_BATCH_SIZE`: determines the number of payment results that will be written to disk at a time. 
- `TOTAL_TIME`: the total runtime for the simulation expressed in seconds.

docker/docker.md
```diff
@@ -0,0 +1,87 @@
...
+```bash
+make help
+```
+
+---
+
+## Notes for Developers:
+
+- When you use the `mount-volume` command, it sets up a Docker volume named `simln-config`. This volume contains your `config.json` file and other necessary files, ensuring that the data is accessible inside the Docker container.
+- The paths to certificates, macaroons, etc., in your `config.json` are automatically adjusted to point to their corresponding paths inside the Docker volume.
```
carlaKC, line 87: re-emphasize that IP address is not converted while we're here

docker/entrypoint.sh
```diff
@@ -0,0 +1,37 @@
...
```
carlaKC, line 8: Needs rebase, we removed `--config` recently so now it's just `sim-cli sim.json` (since it's always required). 
Also decided to refer to that file as `sim.json` since we may one day want to have a config file for the _actual_ simulation (containing things like log level and print batch etc).

carlaKC (commented):
tACK. 

This is ✨ the nicest ✨ docker setup and docs I have ever encountered. The automated setup of config makes it an absolute joy to work with. Really, really great stuff. 

I have a few nits on naming and fixing up documentation which I'd like to get in for the release - specifically making a note that folks can bring their own volume if they don't want to use our script (which will be useful in the case where we don't have everything we need locally, so we can't just copy it into a volume).

docker/docker.md
sr-gi: Agreed. I had to go look for the `host.docker.internal` thing to make it run with polar

sr-gi (commented):
Looks really good. I have a few comments, check inline.

Something I'm also wondering: should both `.dockerignore` and the `Makefile` be moved to the `docker` folder? They only make sense within docker, and you may be reading the docker readme from within that folder already, so I may expect to run the commands there too.

Lastly, this needs rebase (aliases are not available atm), and commits should be squashed IMO, given they don't make sense standalone

carlaKC:
> Makefile be moved to the docker folder

I think that it would be nice to have a `make install` which handles `cargo install --locked ...` in future so I think it's fine leave where it is. 

>  .dockerignore

No strong feelings about where this one lives.

Makefile
```diff
@@ -0,0 +1,31 @@
...
+	@echo ""
+	@echo "Variables:"
+	@echo "SIMFILE_PATH       Path to the sim.json file."
+	@echo "LOG_LEVEL         Set the logging level (default: info) e.g. <make run LOG_LEVEL=debug>"
+	@echo "HELP              Set to true to print the help message (default: false) e.g. <make run HELP=true>"
+	@echo "PRINT_BATCH_SIZE  Set the batch size for printing the results e.g. <make run PRINT_BATCH_SIZE=100>"
+	@echo "TOTAL_TIME        Set the total time for the simulation e.g. <make run TOTAL_TIME=1000>"
+
+run-docker:
+	docker run --rm --name sim-ln --init -v simln-data:/data -e SIMFILE_PATH=/data/sim.json -e LOG_LEVEL=$(LOG_LEVEL) -e HELP=${HELP} -e PRINT_BATCH_SIZE=${PRINT_BATCH_SIZE} -e TOTAL_TIME=${TOTAL_TIME} sim-ln
```
carlaKC, line 25: nit: `docker run -d` so that this is in the background?

carlaKC (approved):
tACK 🚀 

Thanks for the lightning-fast turnaround on this one 🙏 🙇  

Only last ask if to have `run-docker` updated to be a background deployment (`-d`) if you've got time for one last push. Otherwise let's shipppp it 🐑

Makefile
```diff
@@ -0,0 +1,31 @@
...
+help:
+	@echo "Makefile commands:"
+	@echo "build-docker      Builds the Docker image."
+	@echo "mount-volume      Mounts the specified volume."
+	@echo "run               Runs the Docker container."
+	@echo "run-interactive   Runs the Docker container in interactive mode."
+	@echo "stop              Stops the Docker container."
+	@echo ""
+	@echo "Variables:"
+	@echo "SIMFILE_PATH       Path to the sim.json file."
```
sr-gi, line 18: nit: remove extra leading space

sr-gi (commented):
utACK

There are some pending thigs but I'm happy to ship it with he current fixes and improve it down the line

carlaKC (approved):
tACK @ d8576dc

PR #120, opened 2023-10-03
sim-lib: Adds LightningNode::get_network

sr-gi:
The current approach to get the network from a node stores the information in `NodeInfo`. 

However, this is troublesome going forward (after #103) given we will start using `NodeInfo` to store information about nodes we do not control (and from which we cannot query the network). Furthermore, the network information is worthless once we've sanity-checked that we are running on a network we support and all nodes are running on the same one, so there is no point keeping that around

### DIFF
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -1,5 +1,6 @@
 use async_trait::async_trait;
 use bitcoin::secp256k1::PublicKey;
+use bitcoin::Network;
 use csv::WriterBuilder;
 use lightning::ln::features::NodeFeatures;
 use lightning::ln::PaymentHash;
@@ -120,7 +121,6 @@ pub struct NodeInfo {
     pub pubkey: PublicKey,
     pub alias: String,
     pub features: NodeFeatures,
-    pub network: String,
 }
 
 /// LightningNode represents the functionality that is required to execute events on a lightning node.
@@ -128,6 +128,8 @@ pub struct NodeInfo {
 pub trait LightningNode {
     /// Get information about the node.
     fn get_info(&self) -> &NodeInfo;
+    /// Get the network this node is running at
+    async fn get_network(&mut self) -> Result<Network, LightningError>;
     /// Keysend payment worth `amount_msat` from a source node to the destination node.
     async fn send_payment(
         &mut self,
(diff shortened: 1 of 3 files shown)


sim-lib/src/lib.rs
```diff
@@ -276,37 +278,33 @@ impl Simulation {
...
+    // validates that the nodes are all on the same network and ensures that we're not running on mainnet.
     async fn validate_node_network(&self) -> Result<(), LightningError> {
-        let mut valid_network = Option::None;
-        let unsupported_networks: HashSet<String> = vec!["mainnet", "bitcoin"]
-            .into_iter()
-            .map(|s| s.to_string())
-            .collect();
+        if self.nodes.is_empty() {
+            return Err(LightningError::ValidationError(
+                "we don't control any nodes.".to_string(),
```
carlaKC, line 285: nit: relate error to config "specify at least one node"

sim-lib/src/lib.rs
```diff
@@ -276,37 +278,33 @@ impl Simulation {
...
+            ));
+        }
+        let mut running_network = Option::None;
 
         for node in self.nodes.values() {
-            let network = node.lock().await.get_info().network.clone();
-            if unsupported_networks.contains(&network) {
+            let network = node.lock().await.get_network().await?;
+            if network == Network::Bitcoin {
```
carlaKC, line 292: Ran into this issue in the original PR - LND returns "mainnet" and LDK uses "bitcoin" so we need to check both values here.

carlaKC (commented):
Approach ACK, just need to deal with the different values from different impls.

sim-lib/src/lib.rs
sr-gi: Oh damn, I thought `from_core_arg` considered "mainnet" and "testnet" as valid variants, but now that I remember some contributors nitpicked that core uses "main" and "test", so we cannot use that.

I'll replace it with `from_str` which allows "testnet", and will replace "mainnet" for "bitcoin" in place.

sr-gi:
Amended comments from review

sim-lib/src/lnd.rs
```diff
@@ -88,6 +81,28 @@ impl LightningNode for LndNode {
...
+    async fn get_network(&mut self) -> Result<Network, LightningError> {
+        let info = self
+            .client
+            .lightning()
+            .get_info(GetInfoRequest {})
+            .await
+            .map_err(|err| LightningError::GetInfoError(err.to_string()))?
+            .into_inner();
+
+        if info.chains.is_empty() {
```
carlaKC, line 93: Could tighten this to check `len=1` because _theoretically this API allows multiple chains (I _believe_ there was a time where you could run litecoin and bitcoin, good god whyyy). I don't think we'll ever hit it but it's strictly the most correct way to handle it.

sim-lib/src/lnd.rs
```diff
@@ -88,6 +81,28 @@ impl LightningNode for LndNode {
...
+            return Err(LightningError::GetInfoError(
+                "LND node is not connected to a single chain".to_string(),
+            ));
+        }
+
+        Ok(Network::from_str(match info.chains[0].network.as_str() {
+            "mainnet" => "bitcoin",
+            x => x,
```
carlaKC, line 101: "simnet" is making this unhappy: 
```
Error: Lightning Error: GetInfoError("failed to parse simnet as network")

Caused by:
    Get info error: failed to parse simnet as network
```

sr-gi: Of course LND has its own regtest 🤦 

I'll patch that one too.

sim-lib/src/lnd.rs
sr-gi: So if they are connected to more than one backend we error out?

carlaKC: Yeah, just enforce one chain to keep it simple.

carlaKC: Also like ... be normal. GTFO with yo litecoin lightning.

sim-lib/src/lnd.rs
```diff
@@ -88,6 +81,38 @@ impl LightningNode for LndNode {
...
+            .map_err(|err| LightningError::GetInfoError(err.to_string()))?
+            .into_inner();
+
+        if info.chains.is_empty() {
+            return Err(LightningError::ValidationError(
+                "LND node is not connected any chain".to_string(),
+            ));
+        } else if info.chains.len() > 1 {
+            return Err(LightningError::ValidationError(format!(
+                "LND node is connected to more than one chain: {:?}",
```
sr-gi, line 99: Just a note to self to format this with the `NodeId` when rebasing for #103.

sr-gi:
Addresses comments from @carlaKC

carlaKC (approved):
💯  https://github.com/bitcoin-dev-project/sim-ln/pull/120/commits/01123bd2183ee712f0b8bf88cc6589c842f66265

okjodom (approved):
tACK.

I think LND `get_network` failing with `ValidationErrors`  is both _odd and very efficient_. Like, LND is self validating it's network, but the spec for `get_network` only says "tell me your network". Not blocking

sr-gi:
> tACK.
> 
> I think LND `get_network` failing with `ValidationErrors` is both _odd and very efficient_. Like, LND is self validating it's network, but the spec for `get_network` only says "tell me your network". Not blocking

Well, there's the issue that they have specific variants that cannot be used to construct `bitcoin::Network`, which is a bit odd by itself. Now that you mention it, however, I think the CLN one should also return `ValidationErrors` if `Network::from_core_arg` fails.

sr-gi:
Rebased

carlaKC (approved):
✅ https://github.com/bitcoin-dev-project/sim-ln/pull/120/commits/def8dd10970a8b24c19d6b75a3b211b140054152

PR #130, opened 2023-10-05
Renames config file to sim file

sr-gi:
Currently we are using the term config file for the file that contains the simulation details. This is kind of fine for now, but it will get messy down the line when we want to start storing default values for the CLI parameters in a config file. Furthermore, we are requiring the config file to be passed, but `--config` is required.

This both renames config to sim, and makes the parameter positional instead of named.

### DIFF
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -86,7 +86,7 @@ impl std::fmt::Display for NodeId {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
-pub struct Config {
+pub struct SimParams {
     pub nodes: Vec<NodeConnection>,
     #[serde(default)]
     pub activity: Vec<ActivityParser>,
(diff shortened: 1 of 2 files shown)


sim-lib/src/lib.rs
```diff
@@ -61,7 +61,7 @@ pub struct ClnConnection {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
-pub struct Config {
+pub struct SimDetails {
```
carlaKC, line 64: I don't love this name because writing "run simulation with details" or "run simulation with simulation file" in a readme doesn't _really_ make sense to me. 

gpt says setup.json or params.json but I don't know those are better.

carlaKC (approved):
tACK, naming nit but I don't have much better.

sim-lib/src/lib.rs
sr-gi: I'm happy with SimParams if you want to call them simulation parameters

carlaKC: Feelings aren't strong enough for another round, happy to go in as-is.

carlaKC (approved):

PR #132, opened 2023-10-06
sim-cli: source simulation file from data directory

okjodom:
Introduces `SIM_LN_DATA_DIR` from which `sim-cli` sources simulation files and writes simulation outputs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,10 +41,13 @@ Install the CLI:
 cargo install --locked --path sim-cli/
 ```
 
-Run with Simulation File (see [setup instructions](#simulation-file-setup) for details):
- ```
-sim-cli sim.json
+To run the simulation, create a simulation file `sim.json` in the working directory (see [setup instructions](#simulation-file-setup) for details) and run:
+
 ```
+sim-cli
+```
+
+Run `sim-cli -h` for details on `--data-dir` and `--sim-file` options that allow specifying custom simulation file locations.
 
 ### Simulation File Setup
 The simulator requires access details for a set of `nodes` that you 
@@ -190,10 +193,12 @@ described above). Activity sources and destinations may reference the
 *must* provide a valid public key.
 
 ### Simulation Output
-A summary of the results will be logged by the simulator, and a full 
-list of payments made with their outcomes is available in 
-`simulation_{timestamp}.csv` in the directory that the simulation was 
-executed in. For more detailed logs, use the `--log-level` cli flag.
+
+A summary of the results will be logged by the simulator. A full list of payments made with their outcomes
+is available in `simulation_{timestamp}.csv` within the configured `{data_dir}/results`.
+
+Run `sim-cli -h` for details on data directory (`--data-dir`) and other options including `--print-batch-size`
+which affect how simulation outputs are persisted
 
 ## Lightning Environments
 If you're looking to get started with local lightning development, we
--- a/sim-cli/Cargo.toml
+++ b/sim-cli/Cargo.toml
@@ -11,6 +11,7 @@ Instantly simulate real-world Lightning network activity
 [dependencies]
 anyhow = { version = "1.0.69", features = ["backtrace"] }
 clap = { version = "4.1.6", features = ["derive", "env", "std", "help", "usage", "error-context", "suggestions"], default-features = false }
+dialoguer = "0.11.0"
 log = "0.4.20"
 serde = "1.0.183"
 serde_json = "1.0.104"
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -10,9 +10,15 @@ use clap::Parser;
 use log::LevelFilter;
 use sim_lib::{
     cln::ClnNode, lnd::LndNode, ActivityDefinition, LightningError, LightningNode, NodeConnection,
-    NodeId, SimParams, Simulation,
+    NodeId, SimParams, Simulation, WriteResults,
 };
 use simple_logger::SimpleLogger;
+
+/// The default directory where the simulation files are stored and where the results will be written to.
+pub const DEFAULT_DATA_DIR: &str = ".";
+
+/// The default simulation file to be used by the simulator.
+pub const DEFAULT_SIM_FILE: &str = "sim.json";
 
 /// The default expected payment amount for the simulation, around ~$10 at the time of writing.
 pub const EXPECTED_PAYMENT_AMOUNT: u64 = 3_800_000;
@@ -42,8 +48,12 @@ fn deserialize_f64_greater_than_zero(x: String) -> Result<f64, String> {
 #[derive(Parser)]
 #[command(version, about)]
 struct Cli {
+    /// Path to a directory containing simulation files, and where simulation results will be stored
+    #[clap(long, short, default_value = DEFAULT_DATA_DIR)]
+    data_dir: PathBuf,
     /// Path to the simulation file to be used by the simulator
-    #[clap(index = 1)]
+    /// This can either be an absolute path, or relative path with respect to data_dir
+    #[clap(long, short, default_value = DEFAULT_SIM_FILE)]
     sim_file: PathBuf,
     /// Total time the simulator will be running
     #[clap(long, short)]
@@ -77,8 +87,9 @@ async fn main() -> anyhow::Result<()> {
         .init()
         .unwrap();
 
+    let sim_path = read_sim_path(cli.data_dir.clone(), cli.sim_file).await?;
     let SimParams { nodes, activity } =
-        serde_json::from_str(&std::fs::read_to_string(cli.sim_file)?)
+        serde_json::from_str(&std::fs::read_to_string(sim_path)?)
             .map_err(|e| anyhow!("Could not deserialize node connection data or activity description from simulation file (line {}, col {}).", e.line(), e.column()))?;
 
     let mut clients: HashMap<PublicKey, Arc<Mutex<dyn LightningNode + Send>>> = HashMap::new();
@@ -179,14 +190,22 @@ async fn main() -> anyhow::Result<()> {
         });
     }
 
+    let write_results = if !cli.no_results {
+        Some(WriteResults {
+            results_dir: mkdir(cli.data_dir.join("results")).await?,
+            batch_size: cli.print_batch_size,
+        })
+    } else {
+        None
+    };
+
     let sim = Simulation::new(
         clients,
         validated_activities,
         cli.total_time,
-        cli.print_batch_size,
         cli.expected_pmt_amt,
         cli.capacity_multiplier,
-        cli.no_results,
+        write_results,
     );
     let sim2 = sim.clone();
 
@@ -198,4 +217,56 @@ async fn main() -> anyhow::Result<()> {
     sim.run().await?;
 
     Ok(())
+}
+
+async fn read_sim_path(data_dir: PathBuf, sim_file: PathBuf) -> anyhow::Result<PathBuf> {
+    let sim_path = if sim_file.is_relative() {
+        data_dir.join(sim_file)
+    } else {
+        sim_file
+    };
+
+    if sim_path.exists() {
+        Ok(sim_path)
+    } else {
+        log::info!("Simulation file '{}' does not exist.", sim_path.display());
+        select_sim_file(data_dir).await
+    }
+}
+
+async fn select_sim_file(data_dir: PathBuf) -> anyhow::Result<PathBuf> {
+    let sim_files = std::fs::read_dir(data_dir.clone())?
+        .filter_map(|f| {
+            f.ok().and_then(|f| {
+                if f.path().extension()?.to_str()? == "json" {
+                    f.file_name().into_string().ok()
+                } else {
+                    None
+                }
+            })
+        })
+        .collect::<Vec<_>>();
+
+    if sim_files.is_empty() {
+        anyhow::bail!(
+            "no simulation files found in {}.",
+            data_dir.canonicalize()?.display()
+        );
+    }
+
+    let selection = dialoguer::Select::new()
+        .with_prompt(format!(
+            "Select a simulation file. Found these in {}",
+            data_dir.canonicalize()?.display()
+        ))
+        .items(&sim_files)
+        .default(0)
+        .interact()?;
+
+    Ok(data_dir.join(sim_files[selection].clone()))
+}
+
+async fn mkdir(dir: PathBuf) -> anyhow::Result<PathBuf> {
+    tokio::fs::create_dir_all(&dir).await?;
+    Ok(dir)
 }
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -9,6 +9,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
 use std::fmt::{Display, Formatter};
 use std::marker::Send;
+use std::path::PathBuf;
 use std::time::UNIX_EPOCH;
 use std::{collections::HashMap, sync::Arc, time::SystemTime};
 use thiserror::Error;
@@ -326,15 +327,21 @@ pub struct Simulation {
     shutdown_listener: Listener,
     // Total simulation time. The simulation will run forever if undefined.
     total_time: Option<time::Duration>,
-    /// The number of activity results to batch before printing in CSV.
-    print_batch_size: u32,
     /// The expected payment size for the network.
     expected_payment_msat: u64,
     /// The number of times that the network sends its total capacity in a month of operation when generating random
     /// activity.
     activity_multiplier: f64,
-    /// Whether we want the simulation not to produce and result file. Useful for developing, defaults to false.
-    no_results: bool,
+    /// Configurations for printing results to CSV. Results are not written if this option is None.
+    write_results: Option<WriteResults>,
+}
+
+#[derive(Clone)]
+pub struct WriteResults {
+    /// Data directory where CSV result files are written.
+    pub results_dir: PathBuf,
+    /// The number of activity results to batch before printing in CSV.
+    pub batch_size: u32,
 }
 
 impl Simulation {
@@ -342,10 +349,9 @@ impl Simulation {
         nodes: HashMap<PublicKey, Arc<Mutex<dyn LightningNode + Send>>>,
         activity: Vec<ActivityDefinition>,
         total_time: Option<u32>,
-        print_batch_size: u32,
         expected_payment_msat: u64,
         activity_multiplier: f64,
-        no_results: bool,
+        write_results: Option<WriteResults>,
     ) -> Self {
         let (shutdown_trigger, shutdown_listener) = triggered::trigger();
         Self {
@@ -354,10 +360,9 @@ impl Simulation {
             shutdown_trigger,
             shutdown_listener,
             total_time: total_time.map(|x| Duration::from_secs(x as u64)),
-            print_batch_size,
             expected_payment_msat,
             activity_multiplier,
-            no_results,
+            write_results,
         }
     }
 
@@ -561,8 +566,7 @@ impl Simulation {
             result_logger,
             results_receiver,
             listener,
-            self.print_batch_size,
-            self.no_results,
+            self.write_results.clone(),
         ));
         log::debug!("Simulator data collection set up.");
     }
@@ -886,14 +890,11 @@ async fn consume_simulation_results(
     logger: Arc<Mutex<PaymentResultLogger>>,
     receiver: Receiver<(Payment, PaymentResult)>,
     listener: Listener,
-    print_batch_size: u32,
-    no_results: bool,
+    write_results: Option<WriteResults>,
 ) {
     log::debug!("Simulation results consumer started.");
 
-    if let Err(e) =
-        write_payment_results(logger, receiver, listener, print_batch_size, no_results).await
-    {
+    if let Err(e) = write_payment_results(logger, receiver, listener, write_results).await {
         log::error!("Error while reporting payment results: {:?}.", e);
     }
 
@@ -904,20 +905,19 @@ async fn write_payment_results(
     logger: Arc<Mutex<PaymentResultLogger>>,
     mut receiver: Receiver<(Payment, PaymentResult)>,
     listener: Listener,
-    print_batch_size: u32,
-    no_results: bool,
+    write_results: Option<WriteResults>,
 ) -> Result<(), SimulationError> {
-    let mut writer = if !no_results {
-        Some(WriterBuilder::new().from_path(format!(
+    let mut writer = write_results.and_then(|write_result| {
+        let file = write_result.results_dir.join(format!(
             "simulation_{:?}.csv",
             SystemTime::now()
                 .duration_since(SystemTime::UNIX_EPOCH)
                 .unwrap()
                 .as_secs()
-        ))?)
-    } else {
-        None
-    };
+        ));
+        let writer = WriterBuilder::new().from_path(file).ok()?;
+        Some((writer, write_result.batch_size))
+    });
 
     let mut counter = 1;
     loop {
@@ -925,7 +925,7 @@ async fn write_payment_results(
             biased;
             _ = listener.clone() => {
                 log::debug!("Simulation results consumer received shutdown signal.");
-                return writer.map_or(Ok(()), |ref mut w| w.flush().map_err(|_| SimulationError::FileError))
+                return writer.map_or(Ok(()), |(ref mut w, _)| w.flush().map_err(|_| SimulationError::FileError))
             },
             payment_report = receiver.recv() => {
                 match payment_report {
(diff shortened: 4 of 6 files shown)


sim-cli/src/main.rs
```diff
@@ -15,34 +14,51 @@ use simple_logger::SimpleLogger;
 #[derive(Parser)]
 #[command(version)]
 struct Cli {
-    #[clap(index = 1)]
-    sim_file: PathBuf,
+    /// Path to folder containing simulation files and where simulation results will be written
+    #[arg(index = 1, env = "SIMLN_DATA_DIR")]
+    data_dir: PathBuf,
+    /// Name of the simulation file to use. These are located in `${SIMLN_DATA_DIR}/sims/`
+    /// and are in JSON format. If not specified, the cli will prompt the user to select one.
```
carlaKC, line 21: Small personal preference to have:
*  `sim-file` in `SIM_LN_DATA_DIR` 
* simulation results in `SIM_LN_DATA_DIR/results` sub-dir

Reasoning being:
* `simulation_xyz.csv` files can get a bit noisy when you've run the simulator a few times. 
* Sim file in a sub-dir seems a little atypical (I can see folks creating a sim file in the root, trying to run then needing to move it).

sim-cli/src/main.rs
```diff
@@ -15,34 +14,51 @@ use simple_logger::SimpleLogger;
 #[derive(Parser)]
 #[command(version)]
 struct Cli {
-    #[clap(index = 1)]
-    sim_file: PathBuf,
+    /// Path to folder containing simulation files and where simulation results will be written
+    #[arg(index = 1, env = "SIMLN_DATA_DIR")]
```
carlaKC, line 18: Breaking change to change the first argument from `sim-file` to needing to provide a data dir. 

How about assuming that the working directory is the data dir and having this as a flag to point somewhere else if desired?

carlaKC (commented):
Concept ACK, adding a data dir is super useful (esp in the docker context). 

I do think we should think about the power of defaults / making the "getting started" journey as simple as possible with this change, specifically:
* Assume `sim.json` to be the default simulation file unless otherwise specified. 
* Assume working directory to be the default data dir unless otherwise specified. 

This makes for some quite nice options IMO: 
- `sim-cli` -> looks for simulation file in `{pwd}/sim.json`
- `sim-cli my-sim.json` -> looks for simulation file at `{pwd}/my-sim.json`
- `sim-cli --data-dir=/Usr/simulations` -> looks for simulation file at `/Usr/simulations/sim.json`
- `sim-cli my-sim.json --data-dir=/Usr/simulations` -> looks for simulation file at `/Usr/simulations/my-sim.json`

I also really like having the selection option, which we can use to present other `json` files if we can't find the default value we were expecting.

okjodom:
Changes proposed introduce simln data dir, but I did not configure this to work with docker set up.
@Extheoisah, do you want to collab as part of #136

Extheoisah:
> Changes proposed introduce simln data dir, but I did not configure this to work with docker set up. @Extheoisah, do you want to collab as part of #136

Sure @okjodom 
I'll also review the pr.

sim-cli/src/main.rs
```diff
@@ -42,9 +48,12 @@ fn deserialize_f64_greater_than_zero(x: String) -> Result<f64, String> {
...
 #[command(version, about)]
 struct Cli {
+    /// Path to a directory containing simulation files, and where simulation results will be stored
+    #[clap(long, short, env = "SIM_LN_DATA_DIR", default_value = DEFAULT_DATA_DIR)]
+    data_dir: PathBuf,
     /// Path to the simulation file to be used by the simulator
-    #[clap(index = 1)]
-    sim_file: PathBuf,
+    #[clap(long, short, default_value = DEFAULT_SIM_FILE)]
+    sim_file: String,
```
carlaKC, line 56: [meta] We do have a few people using this in the wild, and this will be a breaking change for them :( 

I think this is _okay_, but we do need to eventually draw the line for this type of change because it can be super annoying when your tooling breaks every time you upgrade.

carlaKC (commented):
tACK, very nice improvement. 

Just wondering about turning that arg into a flag, I can be convinced in either direction.

sim-cli/Cargo.toml
```diff
@@ -11,6 +11,7 @@ Instantly simulate real-world Lightning network activity
 [dependencies]
 anyhow = { version = "1.0.69", features = ["backtrace"] }
 clap = { version = "4.1.6", features = ["derive", "env", "std", "help", "usage", "error-context", "suggestions"], default-features = false }
+dialoguer = "0.11.0"
```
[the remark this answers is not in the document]
okjodom: good catch

sim-lib/src/lib.rs
```diff
@@ -939,6 +940,7 @@ async fn write_payment_results(
                                 SimulationError::CsvError(e)
                             })?;
 
+                            let print_batch_size = print_results.as_ref().unwrap().batch_size;
```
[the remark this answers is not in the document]
okjodom: looks cleaner

sim-cli/src/main.rs
okjodom: Agreed on goal to limit breaking changes on public api and experience. In this case, I'm hoping `sim.json` as `DEFAULT_SIM_FILE` coupled with the prompt to select a file from some default or configured data dir more than makes up for breaking current behavior. We cannot preserve a positional `sim_file` arg because this path requirement would likely to conflict with path to data dir

carlaKC (approved):

sim-cli/src/main.rs
carlaKC: Yeah, agree it's nice enough UX improvement to justify the break in this case :)

carlaKC:
@Extheoisah could you take a look to make sure this is going to work for docker-land and then we'll merge? 

Also could use an update to the readme to explain these new snazzy features, I'm also happy to do that in a followup :)

sim-cli/src/main.rs
```diff
@@ -199,3 +217,47 @@ async fn main() -> anyhow::Result<()> {
...
+        }
+
+        let selection = dialoguer::Select::new()
+            .with_prompt("Select a simulation file")
+            .items(&sim_files)
+            .default(0)
+            .interact()?;
+
+        Ok(data_dir.join(sim_files[selection].clone()))
+    }
```
sr-gi, line 255: Do we really want to go this route? IMO a file picker for when the user has wrongly selected the file they want to read from is overkill. It's sort of mixing cmd with a GUI

sim-cli/src/main.rs
sr-gi, line 56: Why are we defining this as `String` while `data_dir` is `PathBuf`?

sim-cli/src/main.rs
```diff
@@ -199,3 +217,47 @@ async fn main() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+async fn read_sim_path(data_dir: PathBuf, sim_file: String) -> anyhow::Result<PathBuf> {
+    let sim_path = data_dir.join(sim_file);
+
+    let sim_path = if sim_path.extension().is_none() {
+        sim_path.with_extension("json")
```
sr-gi, line 225: Kind of like the same as with the file picker. Do we want to assume the user won't be able to specify the file extension?

IMO this whole function shouldn't be needed

sim-lib/src/lib.rs
```diff
@@ -326,26 +327,31 @@ pub struct Simulation {
...
     activity_multiplier: f64,
-    /// Whether we want the simulation not to produce and result file. Useful for developing, defaults to false.
-    no_results: bool,
+    /// Configurations for printing results to CSV. Results are not printed if this option is None.
+    print_results: Option<PrintResults>,
+}
+
+#[derive(Clone)]
+pub struct PrintResults {
+    /// Data directory where CSV result files are printed.
```
sr-gi, line 341: printed -> written?

sim-cli/src/main.rs
```diff
@@ -199,3 +217,47 @@ async fn main() -> anyhow::Result<()> {
...
+            .interact()?;
+
+        Ok(data_dir.join(sim_files[selection].clone()))
+    }
+}
+
+async fn mkdir(dir: PathBuf) -> anyhow::Result<PathBuf> {
+    if !dir.exists() {
+        tokio::fs::create_dir(&dir).await?;
+    }
```
sr-gi, line 261: This can be replaced by tokio::fs::create_dir_all

https://docs.rs/tokio/latest/tokio/fs/fn.create_dir_all.html

sim-cli/src/main.rs
okjodom: so users just declare a file name and we always search for it within `data_dir`

sim-cli/src/main.rs
okjodom: Not sure of the mix. This seems like a standard cmd / cli UX, no?
```
$ cargo run -p sim-cli -- -s test
    Finished dev [unoptimized + debuginfo] target(s) in 0.38s
     Running `target/debug/sim-cli -s test`
2023-11-15T19:10:18.036Z INFO  [sim_cli] Simulation file './test.json' does not exist
Select a simulation file:
> sim.json
```
User tries to run simulation with a sim file named `test.json`, but it's not found in the configured/default data directory, so potential alternatives are immediately listed. Seems handy in case of a simple misspelling of the file name.

IMO a UX improvement, though perhaps not asked for by current set of users

sim-cli/src/main.rs
okjodom: simplified

sim-cli/src/main.rs
okjodom: This ergonomic allows user to specify file name without extension.
The rest of the function just builds a 'nice UX' around the `data_dir` + `sim_file` pairing, so teh question is whether users need that UX or not

sim-cli/src/main.rs
sr-gi: I just think it is an unnecessary bit of code that, on top of it, adds a dependency. Telling the user the file cannot be found in the location they've specified should be sufficient.

I'm not going to oppose this if there is consensus that this is a nice thing to have, I just find it unnecessary

sim-cli/src/main.rs
sr-gi: I mean this could be `PathBuf` and the code will still run fine, not sure why it got changed

sim-cli/src/main.rs
sr-gi: Sure, I just find whoever is using sim-ln should be able to properly define a path and a file name with an extension. It is nice to have UX features that make it easy for the user to run the software, but I find this is way too much. We're expecting the user to be able to define the sim file which already contains paths, extensions and so on

sim-cli/src/main.rs
okjodom: Keeping it as a `PathBuf` would allow a user to specify a different path in conflict with configured / default `data_dir`. Further, since the `sim_file` configuration is now a flag and not a required positional arg, we'd need `PathBuf` value parser if we are to have a default value for it.

okjodom:
@sr-gi, in regards to your perspective on how this change proposes to change how we source `sim_file`, I could consider a smaller proposal that only introduces `data_dir` for purposes of storing result outputs.

I'm excited about the new UX that always expects `sim_file` in `data_dir` and the helper ergonomics for picking one among many sim files, but I'll only push this if the UX is helpful and necessary.

Comments [one](https://github.com/bitcoin-dev-project/sim-ln/pull/132#discussion_r1394502028), [two](https://github.com/bitcoin-dev-project/sim-ln/pull/132#discussion_r1394504980) and [three](https://github.com/bitcoin-dev-project/sim-ln/pull/132#discussion_r1394506326) are easy resolved if we can agree to change/preserve UX for passing `sim_file` to the cli

sim-cli/src/main.rs
sr-gi: I may be missing your point, but I don't think that's actually the case.

Right now you can already specify `--sim-file <path>/sim.json`. Furthermore, changing it to `Pathbuf` does not require changing the default value (notice how `DEFAULT_DATA_DIR` is defined as a `&str` and `data_dir` is `PathBuf`.

sr-gi:
> @sr-gi, in regards to your perspective against how this PR proposes to change how we source `sim_file`, I could consider a smaller proposal that only introduces `data_dir` for purposes of storing result outputs.
> 
> I'm excited about the new UX that always expects `sim_file` in `data_dir` and the helper ergonomics for picking one among many sim files, but I'll only push this if the UX is helpful and necessary.
> 
> Comments [one](https://github.com/bitcoin-dev-project/sim-ln/pull/132#discussion_r1394502028), [two](https://github.com/bitcoin-dev-project/sim-ln/pull/132#discussion_r1394504980) and [three](https://github.com/bitcoin-dev-project/sim-ln/pull/132#discussion_r1394506326) are easy resolved if we can agree to change/preserve UX for passing `sim_file` to the cli

I'm open to being convinced if you and @carlaKC find this useful. It is just not the way I'd approach it

sim-cli/src/main.rs
okjodom: well, I hope its okay to sneak in some nice UX to already happy users :p

sim-cli/src/main.rs
okjodom: > Keeping it as a PathBuf would allow a user to specify a different path in conflict with configured / default data_dir

now I think this might be desirable. applied

okjodom:
> I'm open to being convinced if you and @carlaKC find this useful. It is just not the way I'd approach it

Simple is always good. And what we had was simple + limited. I def polished the UX change too much (should have scoped it to just data dir addition) but don't want to take it back because I think users will like it :p

sim-cli/src/main.rs
okjodom: ack. totally reasonable. See my [comment here](https://github.com/bitcoin-dev-project/sim-ln/pull/132#issuecomment-1813545115)

sim-cli/src/main.rs
okjodom: 💯 now I see how it's over-polished

carlaKC:
My 2c would be: 
1. Remove the [extension completion](https://github.com/bitcoin-dev-project/sim-ln/pull/132#discussion_r1394506326), since advanced users will know what it is.
2. Keep the tab completion and remember to user-test whether people respond to it the next time we do interviews.

okjodom:
checked. trimmed out the extension completion

Extheoisah:
@okjodom @carlaKC just tested the PR with docker and it's broken. I'm working on a PR to fix it. We can merge that after this pr is merged.

okjodom:
> @okjodom @carlaKC just tested the PR with docker and it's broken. I'm working on a PR to fix it. We can merge that after this pr is merged.

Thanks for the docker help. Lmk if you need to pair on it

Extheoisah:
I've got it working already. One question though is if we want to have a data-dir for the docker setup? By default, the simulation CSV files are saved in the results folder in docker in the root. thoughts? cc @carlaKC

okjodom:
> I've got it working already. One question though is if we want to have a data-dir for the docker setup? By default, the simulation CSV files are saved in the results folder in docker in the root. thoughts? cc @carlaKC

Can we treat the data dir treated as a volume and have the csv files saved `data_dir/results` persistent beyond the containers runtime

Extheoisah:
> > I've got it working already. One question though is if we want to have a data-dir for the docker setup? By default, the simulation CSV files are saved in the results folder in docker in the root. thoughts? cc @carlaKC
> 
> Can we treat the data dir treated as a volume and have the csv files saved `data_dir/results` persistent beyond the containers runtime

Yup, just made a pr that does that. I've also fixed the docker volume setup. It wasn't updated when we removed the node type from the sim.json config
pr here -> https://github.com/bitcoin-dev-project/sim-ln/pull/159

sim-cli/src/main.rs
```diff
@@ -179,14 +189,22 @@ async fn main() -> anyhow::Result<()> {
         });
     }
 
+    let write_results = if !cli.no_results {
+        Some(WriteResults {
+            results_dir: mkdir(cli.data_dir.clone().join("results")).await?,
```
sr-gi, line 194: `clone` is not required here

sim-cli/src/main.rs
```diff
@@ -199,3 +217,47 @@ async fn main() -> anyhow::Result<()> {
...
+            .items(&sim_files)
+            .default(0)
+            .interact()?;
+
+        Ok(data_dir.join(sim_files[selection].clone()))
+    }
+}
+
+async fn mkdir(dir: PathBuf) -> anyhow::Result<PathBuf> {
+    tokio::fs::create_dir_all(dir.clone()).await?;
```
sr-gi, line 261: `dir.clone()` -> `&dir`

sim-cli/src/main.rs
```diff
@@ -42,8 +48,11 @@ fn deserialize_f64_greater_than_zero(x: String) -> Result<f64, String> {
 #[derive(Parser)]
 #[command(version, about)]
 struct Cli {
+    /// Path to a directory containing simulation files, and where simulation results will be stored
+    #[clap(long, short, env = "SIM_LN_DATA_DIR", default_value = DEFAULT_DATA_DIR)]
+    data_dir: PathBuf,
     /// Path to the simulation file to be used by the simulator
```
sr-gi, line 54: Maybe specify that is can either be an absolute path, or relative with respect to `data_dir`?

sim-cli/src/main.rs
```diff
@@ -199,3 +217,47 @@ async fn main() -> anyhow::Result<()> {
...
+        data_dir.join(sim_file)
+    } else {
+        sim_file
+    };
+
+    if sim_path.exists() {
+        Ok(sim_path)
+    } else {
+        log::info!("Simulation file '{}' does not exist", sim_path.display());
+        let sim_files: Vec<String> = std::fs::read_dir(data_dir.clone())?
```
sr-gi, line 232: `data_dir.clone()` -> `&data_dir`

sim-cli/src/main.rs
```diff
@@ -199,3 +217,47 @@ async fn main() -> anyhow::Result<()> {
...
+            .filter_map(|f| {
+                f.ok().and_then(|f| {
+                    if f.path().extension()?.to_str()? == "json" {
+                        return f.file_name().into_string().ok();
+                    }
+                    None
```
sr-gi, line 238: nit: for style consistency I'd do:

```
if f.path().extension()?.to_str()? == "json" {
    f.file_name().into_string().ok()
} else {
    None
}
```

Not blocking though

sim-cli/src/main.rs
sr-gi, line 232: Either `: Vec<String>` or `::Vec<_>>()` (when collecting) is redundant. One of them can be dropped

sim-cli/src/main.rs
```diff
@@ -199,3 +217,47 @@ async fn main() -> anyhow::Result<()> {
...
+                })
+            })
+            .collect::<Vec<_>>();
+
+        if sim_files.is_empty() {
+            anyhow::bail!("no simulation files found in {}.", data_dir.display());
```
sr-gi, line 244: Given `data_dir` defaults to `.`, this error reads pretty confusingly:

```
Error: no simulation files found in ..
```

I'd drop the trailing `.` or expand the local path to absolute

sim-cli/src/main.rs
```diff
@@ -199,3 +217,47 @@ async fn main() -> anyhow::Result<()> {
...
+        }
+
+        let selection = dialoguer::Select::new()
+            .with_prompt(format!(
+                "Select a simulation file. Found these in {}:",
```
sr-gi, line 249: This also reads pretty confusing:

```
Select a simulation file. Found these in .::
```

The trailing `:` needs to be removed, given `dialoguer` already adds it, furthermore, ending the prompt with a `.` would still make it look weird. In this case I'd get rid os the format altogether or expand it to absolute

sim-cli/src/main.rs
sr-gi, line 232: nit: (not blocking)

I think all this block could be a function of its own (like `file_picker` or similar)

sim-lib/src/lib.rs
```diff
@@ -904,48 +905,46 @@ async fn write_payment_results(
...
-                            counter = counter % print_batch_size + 1;
-                            if print_batch_size == counter {
+                            counter = counter % batch_size + 1;
+                            if batch_size == counter {
                                 w.flush().map_err(|_| SimulationError::FileError)?;
                             }
                         }
                     },
-                    None => return writer.map(|ref mut w| w.flush().map_err(|_| SimulationError::FileError)).unwrap_or(Ok(())),
+                    None => return writer.map(|(ref mut w, _)| w.flush().map_err(|_| SimulationError::FileError)).unwrap_or(Ok(())),
```
sr-gi, line 947: nit: We can use `map_or` instead of `map`  + `unwrap_or` for consistency with the biased branch

README.md
sr-gi, line 1: Not really sure this belongs here. I'm happy with linting the README, but we have several other markdown files that may also need so. A separate PR to deal with all of them may be better

okjodom: ack. will follow up with lint

sim-cli/src/main.rs
```diff
@@ -42,8 +48,12 @@ fn deserialize_f64_greater_than_zero(x: String) -> Result<f64, String> {
 #[derive(Parser)]
 #[command(version, about)]
 struct Cli {
+    /// Path to a directory containing simulation files, and where simulation results will be stored
+    #[clap(long, short, env = "SIM_LN_DATA_DIR", default_value = DEFAULT_DATA_DIR)]
```
carlaKC, line 52: Why does this have an `env` option and `sim_file` doesn't?

sim-cli/src/main.rs
```diff
@@ -199,3 +218,55 @@ async fn main() -> anyhow::Result<()> {
...
+    let sim_path = if sim_file.is_relative() {
+        data_dir.join(sim_file)
+    } else {
+        sim_file
+    };
+
+    if sim_path.exists() {
+        Ok(sim_path)
+    } else {
+        log::info!("Simulation file '{}' does not exist", sim_path.display());
```
carlaKC, line 232: nit: logs should end with `.`

README.md
```diff
@@ -41,11 +41,14 @@ Install the CLI:
 cargo install --locked --path sim-cli/
 ```
 
-Run with Simulation File (see [setup instructions](#simulation-file-setup) for details):
- ```
-sim-cli sim.json
+Run a simulation
```
carlaKC, line 44: Reading through this, the user doesn't yet know that they need to provide a simulation file. 

To run the simulation, create a simulation file `sim.json` in the working directory (see [setup instructions](#simulation-file-setup) for details) and run: 
```
sim-cli
```

Would also be nice to have some extra hints with `sim-cli -h` instructions like: 

Run `sim-cli -h` for details on  `--data-dir` and `--sim-file` options that allow specifying custom simulation file locations.

carlaKC (commented):
Code comments non-blocking. Would just like to clarify readme further then gg.

sim-cli/src/main.rs
okjodom: no other option is configurable through `env`. removed

carlaKC (approved):
🧡  thanks for updating!

sim-cli/src/main.rs
sr-gi: Oh, I thought this was there because it was being used by the docker setup

sim-cli/src/main.rs
```diff
@@ -199,3 +218,55 @@ async fn main() -> anyhow::Result<()> {
...
+                } else {
+                    None
+                }
+            })
+        })
+        .collect::<Vec<_>>();
+
+    if sim_files.is_empty() {
+        anyhow::bail!(
+            "no simulation files found in {}",
```
sr-gi, line 252: nit: Now that you've canonicalized this, it should end with a `./` for consistency with the rest of the errors.

sr-gi (approved):
Just one non-blocking nit.

LGTM

sim-cli/src/main.rs
carlaKC: Using the [flag](https://github.com/bitcoin-dev-project/sim-ln/pull/159/commits/e01198b204e2e29a009ef398c4a84663bc482fb3#diff-4f5cabe26761257a4d685a6edc7a43e0fe0f78762f50eeb48530f2bd3b3ee7caR8) so we're good 👍

carlaKC: I _think_ that we could update the docker scripts to do things more efficiently (use docker env vars to set config options rather than mapping to flag values) _but_ we'd need to make other config options `env` as well to do that consistently so good to leave it out for now.

sr-gi (approved):
ACK [9975d8a](https://github.com/bitcoin-dev-project/sim-ln/pull/132/commits/9975d8a34ef4f95bbac30054f5620d3566689f18)

PR #133, opened 2023-10-10
sim-all: Adds description to both existing creates

sr-gi:
Mainly so we can have something nice when running `--help`

### DIFF
--- a/sim-cli/Cargo.toml
+++ b/sim-cli/Cargo.toml
@@ -2,6 +2,9 @@
 name = "sim-cli"
 version = "0.1.0"
 edition = "2021"
+description = """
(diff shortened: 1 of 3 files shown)


sr-gi:
Feel free to nitpick on the descriptions, I'm almost as bad as those as I am at naming things.

okjodom (approved):

carlaKC (approved):
lgtm, only nit is to end description with `.` if that's a normal thing to do (idk)

sr-gi:
I think it does not. When in doubt I always check `bitcoind`:

```
> bitcoind --help
Bitcoin Core version v25.0.0
```

carlaKC:
Cool, then shippin it!

PR #139, opened 2023-10-13
example: test utilities and mocking

carlaKC:
Opening this up as an example of how we could approach tests and mocking. 

Main question: 
1. Anyone got strong feelings about mocking libraries? `mockall` has treated me pretty well so far, but it's the only one I've used. 
2. `example_test` doesn't even pass yet, but it's a general idea of what a high level test could look like - thoughts?


carlaKC:
Closing out since this is just an example - feel free to leave comments on approach post-close!

Issue #81, opened 2023-08-28
Feature: Simulation Time


carlaKC:
Some use cases may want to use simulation time rather than seconds. Steps to complete: 
1. Abstract time usage to trait (implemented by system time)
2. Add implementation time trait implementation (this possibly exists already - do some googling)
3. Update user config to be able to opt into sim time (high level parameter / config `interval_secs` -> more generic `interval`?)

carlaKC:
As a first step for this, we can pull out the time based functionality in sim-ln into a `Clock` trait. 

There's some debate as to whether simulation time will work with real nodes, because we can only go as fast as the underlying node can process payments. That said, with the addition of #165 we'd be interested in simulation time (as the simulated nodes are incredibly cheap to run) so I'd like to start down this path.

carlaKC:
Closed in #242

PR #141, opened 2023-10-17
sim-lib: Log payment summary every minute for more consistent results

### LINKED ISSUE #128: Logging: Log payment summary on info level based on time
Right now we info log a payment summary every 10 payments*. This can get very noisy if we have a lot of payments, or be too quiet if we have too few. 

Instead, update logging to log every minute so that we can provide constant feedback to the end user.

carlaKC:
Fixes #128.

Example of new, regular output: 
```
2023-10-17T16:09:44.984Z INFO [sim_lib] Summary of results will be reported every 60s.
...
2023-10-17T16:10:44.986Z INFO [sim_lib] Processed 4 payments sending 19832 msat total with 100.00% success rate.
2023-10-17T16:11:44.991Z INFO [sim_lib] Processed 12 payments sending 275055 msat total with 83.33% success rate.
2023-10-17T16:12:44.994Z INFO [sim_lib] Processed 14 payments sending 278414 msat total with 78.57% success rate.
2023-10-17T16:13:44.996Z INFO [sim_lib] Processed 18 payments sending 3151487 msat total with 77.78% success rate.
```


sim-lib/src/lib.rs
```diff
@@ -549,7 +549,16 @@ impl Simulation {
             listener.clone(),
         ));
 
+        let result_logger = Arc::new(Mutex::new(PaymentResultLogger::new()));
+
+        tasks.spawn(run_results_logger(
+            listener.clone(),
+            result_logger.clone(),
+            Duration::from_secs(60),
```
sr-gi, line 557: We may want to make this configurable and default it to 1 min

sr-gi (commented):
LGTM. 

As a side note, I think this approach makes more sense currently, given we are simulating in real-time. However, it is worth noting that if we add simulated time down the line, we may want to report based on:

- time for real time
- payment count for simulated time

carlaKC:
> However, it is worth noting that if we add simulated time down the line, we may want to report based on
> time for real time
> payment count for simulated time

Yeah agree, would def need to make a change if we implement #81. We'll need to do a bunch of refactoring for that, so we can just trait out the logger as well.

sim-lib/src/lib.rs
carlaKC: Clippy starts getting angry if we have too many args for `Simulation::new` - pushed example in https://github.com/bitcoin-dev-project/sim-ln/pull/141/commits/e929a2d6c28687df4b491594dc84c0ae9a1dc208 which adds a `SimulationConfig` struct to house all of our various flags. 

If we like it, I'll break that up into a refactor commit + add some validation on the config so that we fail if we're given bad config values (eg, zero expected amount).

okjodom:
tACK e929a2d6c28687df4b491594dc84c0ae9a1dc208 ✅

okjodom (approved):

sim-cli/src/main.rs
```diff
@@ -46,6 +48,9 @@ struct Cli {
     /// Do not create an output file containing the simulations results
     #[clap(long, default_value_t = false)]
     no_results: bool,
+    /// The interval at which to log payment summaries, expressed in seconds.
+    #[clap(long, default_value_t = 60)]
```
sr-gi, line 52: nit: I think we should define this as a constant, given all the other defaults are.

We can also take the chance to remove the line break between them :P

sim-cli/src/main.rs
```diff
@@ -1,7 +1,9 @@
 use bitcoin::secp256k1::PublicKey;
+use sim_lib::SimulationConfig;
```
sr-gi, line 2: Merge this with the others at L11

sim-lib/src/lib.rs
```diff
@@ -958,18 +964,44 @@ impl PaymentResultLogger {
...
+        let total_payments = self.success_payment + self.failed_payment;
+        write!(
+            f,
+            "Processed {} payments sending {} msat total with {:.2}% success rate.",
+            total_payments,
+            self.total_sent,
+            (self.success_payment as f64 / total_payments as f64) * 100.0
+        )
+    }
+}
```
sr-gi, line 981: It could now be the case that we sample before a single payment has been tracked yet, meaning that we'd get `0/0` when reporting the success rate, yielding `NaN`:

```
INFO  [sim_lib] Processed 0 payments sending 0 msat total with NaN% success rate.
```

We should treat that as a special case:

```suggestion
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        let total_payments = self.success_payment + self.failed_payment;
        let rate = self.success_payment as f64 / total_payments as f64;
        if rate.is_nan() {
            write!(
                f,
                "Processed {} payments sending {} msat total.",
                total_payments, self.total_sent
            )
        } else {
            write!(
                f,
                "Processed {} payments sending {} msat total with {:.2}% success rate.",
                total_payments,
                self.total_sent,
                rate * 100.0
            )
        }
    }
```

carlaKC: Rate is only NaN if we haven't sent anything yet, so the special case will just print "Processed 0 payments sending 0 msat total." IMO it's not worth the effort to convey essentially the same information in double the LOC, I think it's reasonably understandable as is.

carlaKC:
Dropped the config refactor off this - I'll do it in a follow up, want to close this one out in the meantime (sent simln to some people to play around with, nice to have it logging regularly for demo purposes).

sim-lib/src/lib.rs
```diff
@@ -967,18 +974,44 @@ impl PaymentResultLogger {
         }
 
         self.total_sent += details.amount_msat;
-        self.call_count += 1;
```
sr-gi, line 970: I just realized that this was a `u8`, which would have made the logger task panic after 255 additions, so that's patches 😆

sr-gi (approved):
tACK [71d44cb](https://github.com/bitcoin-dev-project/sim-ln/pull/141/commits/71d44cb8b5499e47cf12859ae51ff6efc2930c03)

It'd be good to open an issue to track the pending changes. At least so we don't forget to add an arg for the logging interval

PR #142, opened 2023-10-17
readme: add warning that IP addresses are not currently supported

carlaKC:


README.md
```diff
@@ -90,6 +90,8 @@ not "drain" from the simulation.
 }
 ```
 
+**Note that IP addresses are not currently supported.**
```
sr-gi, line 93: This is only true for `LND` nodes. `CLN` nodes do support IPs

okjodom (approved):

sr-gi (commented):
I think it looks good now, but the note is within the random activity generation section. This actually applies to manual activity too.

Should we have a "Known limitations" section of sorts by the end of the doc and put these kind of things there?

carlaKC:
> I think it looks good now, but the note is within the random activity generation section. This actually applies to manual activity too.

I figured that first time users will read through the README to figure out both options, so it's okay to just have it in there. 
Could also add it under both config examlpes if that'll be clearer?  

> Should we have a "Known limitations" section of sorts by the end of the doc and put these kind of things there?

I like having it close to the sample config because that's a section that you have to read. It's an annoying thing to have miss the first time because you got started quickly and then have to come back to re-read.

sr-gi:
Fair, lets add it to both sections then

sr-gi:
I think this may not be needed anymore after #148, right?

okjodom:
update: we can have a new tonic lnd shortly and we won't need this warning

okjodom:
> I think this may not be needed anymore after #148, right?

that's right :)

sr-gi:
I'm happy with either tbh, we can put it in and remove it after #148. Whatever makes the most sense to you

sr-gi (approved):

carlaKC:
#151 is haaaapening 😱

PR #143, opened 2023-10-17
sim-all: Adds tests coverage to `random_activity`

sr-gi:
This is a first stab at adding tests to the simulator.

Adds tests to the random activity generator and includes some sanity checks to `sim-cli` so we can fail early if we need to.

### DIFF
--- a/sim-lib/src/random_activity.rs
+++ b/sim-lib/src/random_activity.rs
@@ -1,14 +1,23 @@
 use core::fmt;
 use std::fmt::Display;
+use thiserror::Error;
 
 use bitcoin::secp256k1::PublicKey;
 use rand_distr::{Distribution, Exp, LogNormal, WeightedIndex};
 use std::time::Duration;
 
-use crate::{NetworkGenerator, NodeInfo, PaymentGenerator, SimulationError};
+use crate::{NetworkGenerator, NodeInfo, PaymentGenerationError, PaymentGenerator};
 
 const HOURS_PER_MONTH: u64 = 30 * 24;
 const SECONDS_PER_MONTH: u64 = HOURS_PER_MONTH * 60 * 60;
+
+#[derive(Debug, Error)]
+pub enum RandomActivityError {
+    #[error("Value error: {0}")]
+    ValueError(String),
+    #[error("InsufficientCapacity: {0}")]
+    InsufficientCapacity(String),
+}
 
 /// NetworkGraphView maintains a view of the network graph that can be used to pick nodes by their deployed liquidity
 /// and track node capacity within the network. Tracking nodes in the network is memory-expensive, so we use a single
@@ -23,15 +32,15 @@ impl NetworkGraphView {
     // Creates a network view for the map of node public keys to capacity (in millisatoshis) provided. Returns an error
     // if any node's capacity is zero (the node cannot receive), or there are not at least two nodes (one node can't
     // send to itself).
-    pub fn new(nodes: Vec<(NodeInfo, u64)>) -> Result<Self, SimulationError> {
+    pub fn new(nodes: Vec<(NodeInfo, u64)>) -> Result<Self, RandomActivityError> {
         if nodes.len() < 2 {
-            return Err(SimulationError::RandomActivityError(
+            return Err(RandomActivityError::ValueError(
                 "at least two nodes required for activity generation".to_string(),
             ));
         }
 
         if nodes.iter().any(|(_, v)| *v == 0) {
-            return Err(SimulationError::RandomActivityError(
+            return Err(RandomActivityError::InsufficientCapacity(
                 "network generator created with zero capacity node".to_string(),
             ));
         }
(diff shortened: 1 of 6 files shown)


sr-gi:
Notice here the sanity checks for the accepted values are done via `clap's` `value_parser`, which yield not really friendly error messages:

```
error: invalid value '0' for '--capacity-multiplier <CAPACITY_MULTIPLIER>': 0 is not in 1..18446744073709551615

For more information, try '--help'.
```

We may want to consider adding a `validate` function to CLI in which we manually implement our restrictions plus return more friendly messages.

sr-gi:
Also, after adding these tests it may be worth considering getting rid of https://github.com/bitcoin-dev-project/sim-ln/blob/main/sim-lib/src/random_activity.rs#L70-L73

sim-lib/src/random_activity.rs
```diff
@@ -220,3 +239,159 @@ impl Display for PaymentActivityGenerator {
...
+            nodes.extend(create_nodes(1, 0));
+            assert!(matches!(
+                NetworkGraphView::new(nodes),
+                Err(SimulationError::RandomActivityError { .. })
+            ));
+
+            // All of them are 0
+            assert!(matches!(
+                NetworkGraphView::new(create_nodes(2, 0)),
+                Err(SimulationError::RandomActivityError { .. })
```
carlaKC, line 290: Rust question: should we be: 
1. Returning different enum variants for these errors.
2. String matching the error case to make sure we're covering what we expect.

Not sure which is canonical, but rn it's easy to miss coverage for a branch if the test is changed because we're only asserting "some error happened" here.

sim-lib/src/random_activity.rs
```diff
@@ -220,3 +239,159 @@ impl Display for PaymentActivityGenerator {
...
+            // meaning that a source is part of the sampling pool when a destination is requested. This means that if the source
+            // capacity extremely big compared to the rest of nodes in the graph, it may take extraordinarily long for a destination
+            // to be found.
+            //
+            // This tests a completely unrealistic yet pathological setup in where a single node has the vast majority of the networks
+            // capacity, while every other single node has almost none. This scenario represents a starts in where the big node has a
+            // connection with every single other node, while the rest are only connected to it. Even in this extreme and unrealistic
+            // situation, the method return rather fast.
+
+            let small_node_count = 999;
```
carlaKC, line 315: Very nice test/comment 💯

sim-lib/src/random_activity.rs
```diff
@@ -220,3 +239,159 @@ impl Display for PaymentActivityGenerator {
...
+            ));
+            assert!(matches!(
+                PaymentActivityGenerator::new(get_random_int(1, 10), get_random_int(1, 10), 0.0),
+                Err(SimulationError::RandomActivityError { .. })
+            ));
+        }
+
+        #[test]
+        fn test_validate_capacity() {
+            // There's no much to be tested here, given a `node_capacity` and an `expected_payment`
```
carlaKC, line 376: /s/no/not

carlaKC (commented):
the the era of tests commence!

sim-lib/src/test_utils.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use bitcoin::secp256k1::{PublicKey, Secp256k1, SecretKey};
+
+/// Utility function to create a vector of pseudo random bytes.
+///
+/// Mainly used for testing purposes.
+pub fn get_random_bytes(size: usize) -> Vec<u8> {
+    let mut rng = rand::thread_rng();
+    (&mut rng)
+        .sample_iter(Uniform::new(u8::MIN, u8::MAX))
```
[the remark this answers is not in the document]
sr-gi: Sure

sim-lib/src/random_activity.rs
sr-gi: I think different errors should be better if we want finner granularity here, especially since pattern matching based on error messages is more brittle and will potentially involve larger diff down the line.

If we go with one, I can see at least two categories here that we may want to split:

- `ValueError` referring to one of the input values being invalid. This would apply to the list of provided nodes not being big enough, the source_capacity/expected_payment/multiplier being zero, ...
- `InsufficientCapacity` referring to a node not having enough capacity to process a payment, a `NetworkGraphView` created with zero capacity

This would cover all current cases I think. Do you think we should go even finner?

sim-lib/src/random_activity.rs
```diff
@@ -220,3 +239,159 @@ impl Display for PaymentActivityGenerator {
...
+                assert!(matches!(
+                    NetworkGraphView::new(create_nodes(i, 42 * (i as u64 + 1))),
+                    Err(SimulationError::RandomActivityError { .. })
+                ));
+            }
+
+            // Check that, even if we have two nodes, the the node capacity of all of them must be greater than 0
+            // One of them is 0
+            let mut nodes = create_nodes(1, 0);
+            nodes.extend(create_nodes(1, 0));
```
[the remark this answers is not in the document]
sr-gi: Oh yeah, that's wrong. It should be:

```
nodes.extend(create_nodes(1, x));
```

For any `x > 0`

sim-lib/src/random_activity.rs
carlaKC: > Do you think we should go even finner?

I think that's fine grained enough, especially since we're only using them in tests 👍

sim-lib/src/random_activity.rs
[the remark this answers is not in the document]
sr-gi: This should be fixed now

sim-lib/src/random_activity.rs
sr-gi: I've addressed this in [fd06118](https://github.com/bitcoin-dev-project/sim-ln/pull/143/commits/fd061181ab47ba88c844f1d6de9031a0e65e816b) alongside some error re-definition. Please check whether that makes sense

carlaKC: Good enough error update for now 👍  Hopefully as we include tests with new codes we need to do less of these refactor changes.

PR #145, opened 2023-10-18
sim-cli: Improves CLI docs so help has a more friendly report

sr-gi:
`sim-cli help` is pretty unfriendly atm, add docs so the args/options are covered. 

This makes some optional params to the simulator mandatory and moves the defaulting behavior to the CLI so clap can take care of it. The main rationale for it is so we can easily report the defaults.

Also, restricts `print_batch_size` to be `> 0`, as discussed https://github.com/bitcoin-dev-project/sim-ln/pull/146#issuecomment-1771101726

PR #146, opened 2023-10-18
sim-all: Allow running the simulator without it writing on disk

sr-gi:
It is rather annoying to run the simulator and have to delete the output files when we are doing testing/dev. This allows us to do so without the file being created.

### DIFF
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -43,6 +43,9 @@ struct Cli {
     /// Multiplier of the overall network capacity used by the random activity generator
     #[clap(long, short, default_value_t = ACTIVITY_MULTIPLIER)]
     capacity_multiplier: f64,
+    /// Do not create an output file containing the simulations results
+    #[clap(long, default_value_t = false)]
+    no_results: bool,
 }
 
 #[tokio::main]
@@ -164,6 +167,7 @@ async fn main() -> anyhow::Result<()> {
         cli.print_batch_size,
         cli.expected_pmt_amt,
         cli.capacity_multiplier,
+        cli.no_results,
     );
     let sim2 = sim.clone();
 
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -328,6 +328,8 @@ pub struct Simulation {
     /// The number of times that the network sends its total capacity in a month of operation when generating random
     /// activity.
     activity_multiplier: f64,
+    /// Whether we want the simulation not to produce and result file. Useful for developing, defaults to false.
+    no_results: bool,
 }
 
 impl Simulation {
@@ -338,6 +340,7 @@ impl Simulation {
         print_batch_size: u32,
         expected_payment_msat: u64,
         activity_multiplier: f64,
+        no_results: bool,
     ) -> Self {
         let (shutdown_trigger, shutdown_listener) = triggered::trigger();
         Self {
@@ -349,6 +352,7 @@ impl Simulation {
             print_batch_size,
             expected_payment_msat,
             activity_multiplier,
+            no_results,
         }
     }
 
@@ -528,7 +532,6 @@ impl Simulation {
         tasks: &mut JoinSet<()>,
     ) {
         let listener = self.shutdown_listener.clone();
-        let print_batch_size = self.print_batch_size;
         log::debug!("Setting up simulator data collection.");
 
         // Create a sender/receiver pair that will be used to report final results of simulation.
@@ -544,7 +547,8 @@ impl Simulation {
         tasks.spawn(consume_simulation_results(
             results_receiver,
             listener,
-            print_batch_size,
+            self.print_batch_size,
+            self.no_results,
         ));
         log::debug!("Simulator data collection set up.");
     }
@@ -864,10 +868,11 @@ async fn consume_simulation_results(
     receiver: Receiver<(Payment, PaymentResult)>,
     listener: Listener,
     print_batch_size: u32,
+    no_results: bool,
 ) {
     log::debug!("Simulation results consumer started.");
 
-    if let Err(e) = write_payment_results(receiver, listener, print_batch_size).await {
+    if let Err(e) = write_payment_results(receiver, listener, print_batch_size, no_results).await {
         log::error!("Error while reporting payment results: {:?}.", e);
     }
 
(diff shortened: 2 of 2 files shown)


okjodom:
Could we override `print_batch_size` parameter to signal no csv output? If a user doesn't specify the flag, we don't print csv results by default

sr-gi:
Do you mean using zero as a special case? That could work but it may also introduce potential foot shots. Using a special, non documented, flag means that you need to be pretty aware of what you're doing.

This makes me wonder though, what would happen currently if we set that to zero?

sr-gi:
> Do you mean using zero as a special case? That could work but it may also introduce potential foot shots. Using a special, non documented, flag means that you need to be pretty aware of what you're doing.
> 
> This makes me wonder though, what would happen currently if we set that to zero?

I looked at this and zero means no flushing, meaning that the data will only be written to disk when the simulation ends. Should we discourage that? It feels like we should not allow data to pile in memory. 

@okjodom @carlaKC

okjodom:
> I looked at this and zero means no flushing, meaning that the data will only be written to disk when the simulation ends. Should we discourage that?

yeah we could restrict min value to 1 for the flag

sr-gi:
> > Do you mean using zero as a special case? That could work but it may also introduce potential foot shots. Using a special, non documented, flag means that you need to be pretty aware of what you're doing.
> > This makes me wonder though, what would happen currently if we set that to zero?
> 
> I mean if the flag is not set, we don't start the report writer process. If the flag is set to any value including zero, we start the report writer

Well that's not really how it should be, should it? If the flag is not set, we default to whatever we consider to be a reasonable batch size, which currently is 500. The flag only signals whether we want to be less disk I/O intensive.

Zero, on the other side, could mean do not write to disk, but I think that has risky side-effects as previously mentioned.

> yeah we could restrict min value to 1 for the flag

Agreed

sr-gi:
I've added the `print_batch_size > 0` to https://github.com/bitcoin-dev-project/sim-ln/pull/145, and done some default refactoring for better help report.

okjodom:
> Well that's not really how it should be, should it? If the flag is not set, we default to whatever we consider to be a reasonable batch size, which currently is 500. The flag only signals whether we want to be less disk I/O intensive.

Ack. okay with me if we don't overload `print_batch_size` flag

sim-cli/src/main.rs
```diff
@@ -28,6 +28,10 @@ struct Cli {
     expected_pmt_amt: Option<u64>,
     #[clap(long, short)]
     capacity_multiplier: Option<f64>,
+    // DEV-ONLY args, set hide = true so they are not show in help
```
carlaKC, line 31: Wonder whether this should be dev-only? Some folks in user interviews said they wouldn't really need the result file (would go straight to the node to see what happened), so I can see it being useful beyond dev. 

Weak opinion loosely held.

sim-lib/src/lib.rs
```diff
@@ -334,6 +334,8 @@ pub struct Simulation {
     /// The number of times that the network sends its total capacity in a month of operation when generating random
     /// activity.
     activity_multiplier: f64,
+    /// Wether we want the simulation not to produce and result file. Useful for developing, defaults to false.
```
carlaKC, line 337: s/Wether/Whether

sim-lib/src/lib.rs
```diff
@@ -886,14 +891,19 @@ async fn write_payment_results(
...
 ) -> Result<(), SimulationError> {
-    let mut writer = WriterBuilder::new().from_path(format!(
-        "simulation_{:?}.csv",
-        SystemTime::now()
-            .duration_since(SystemTime::UNIX_EPOCH)
-            .unwrap()
-            .as_secs()
-    ))?;
+    let mut writer = if no_results {
+        WriterBuilder::new().from_path("/dev/null")?
```
carlaKC, line 897: Why not have an `Option` here and skip it all-together below so that we don't have to write anything to disk?

sim-cli/src/main.rs
sr-gi: I don't have a strong opinion here tbh, I found it useful for my testing setup so I figured it may be for other devs too, but if this is a legit use-case we can simply advertise it.

sim-lib/src/lib.rs
sr-gi: I played around with that approach yesterday and it is viable, it just adds more boilerplate.

I'll add a commit on top of this with that approach so you guys can compare it. We can either squash it or remove it depending on what you think

sr-gi: Looks like the rust wizards have been on my side today, I was able to make this pretty minimal. Check 
365746b

sim-lib/src/lib.rs
```diff
@@ -893,49 +893,49 @@ async fn write_payment_results(
...
+                                let _ = w.flush();
+                                SimulationError::CsvError(e)
+                            })?;
 
-                        if print_batch_size == counter {
-                            writer.flush().map_err(|_| SimulationError::FileError)?;
-                            counter = 1;
-                        } else {
-                            counter += 1;
+                            counter = counter % print_batch_size + 1;
```
sr-gi, line 930: I also made this fancier 💅

sim-lib/src/lib.rs
```diff
@@ -886,46 +891,51 @@ async fn write_payment_results(
...
+                            if print_batch_size == counter {
+                                w.flush().map_err(|_| SimulationError::FileError)?;
+                            }
                         }
-                        continue;
```
sr-gi, line 925: I removed this because I don't think it was doing anything (?), at this point there's no more code to be run by this variant, so it is going to loop over anyway

sim-lib/src/lib.rs
carlaKC: Let's squash it - looks good 😎

sim-cli/src/main.rs
carlaKC: Let's surface it for now? We're not drowning in flags (yet 😅 )

sr-gi:
Rebased and squashed

sim-lib/src/lib.rs
```diff
@@ -878,46 +883,51 @@ async fn write_payment_results(
...
-
-    let mut counter = 1;
+    let mut counter = 0;
     loop {
         tokio::select! {
             biased;
             _ = listener.clone() => {
                 log::debug!("Simulation results consumer received shutdown signal.");
-                break writer.flush().map_err(|_| SimulationError::FileError)
+                return writer.map(|ref mut w| w.flush().map_err(|_| SimulationError::FileError)).unwrap_or(Ok(()));
```
carlaKC, line 907: Not sure if this is canonical, but is `map_or` a good fit here? 
`writer.map_or(Ok(()), |mut w| w.flush().map_err(|_| SimulationError::FileError))`

Non-blocking, just 🦀 nit.

sim-lib/src/lib.rs
```diff
@@ -878,46 +883,51 @@ async fn write_payment_results(
...
+                            if print_batch_size == counter {
+                                w.flush().map_err(|_| SimulationError::FileError)?;
+                            }
                         }
-                        continue;
                     },
                     None => {
-                        break writer.flush().map_err(|_| SimulationError::FileError)
+                        return writer.map(|ref mut w| w.flush().map_err(|_| SimulationError::FileError)).unwrap_or(Ok(()));
+
```
carlaKC, line 930: Unnecessary newline (curses on `select` and its interference with fmt).

carlaKC (commented):
tACK, just a non-blocking question + nit left.

sim-lib/src/lib.rs
sr-gi: I think it is!

sr-gi:
Fixed nits

sim-lib/src/lib.rs
carlaKC: My first rust nit 🦀 😭

carlaKC (approved):
LGTM! Happy to go ahead with merge and I'll rebase #141 on this.

sim-lib/src/lib.rs
sr-gi: 🎉

PR #148, opened 2023-10-21
patch: support IP address connections to LND nodes

okjodom:
This patch uses an updated version of [fedimint-tonic-lnd](https://crates.io/crates/fedimint-tonic-lnd) that supports IP address connections to LND nodes.
> wip: **not for check-in**


sr-gi:
It may be worth waiting for the PR to be merged. It LGTM, but just in case things can be improved.

If it takes too long, then we can just merge it.

Really nice work!

okjodom:
closing in favor of #151

Issue #138, opened 2023-10-11
Feature Request: Image in docker hub


m3dwards:
I would love to have an image in docker hub so that I don't have to build SimLN myself. I personally use both ARM and x86 architectures and I think these two are good as a starting point for images.

I have twice (Torq + Scaling Lightning) made multi-architecture docker builds that upload to docker hub on a tag and on a github release so please feel free to take a look at the scaling repo to see how I tackled it.

I'm also happy to contribute or talk through how it works in the projects I've done it for; Torq and Scaling do it two different ways.

sr-gi:
It's been a while, but I've managed to do this before too for multi-arch. Happy to give it a look/chat if necessary to tackle this.

maxwedwards:
@sr-gi would you mind me picking this up?

sr-gi:
@maxwedwards not at all. Feel free

PR #149, opened 2023-10-23
Chore: Github actions for Docker Hub images

m3dwards:
Fixes: #138

Happy to get on a call and discuss the PR if anyone wants.

This PR provides the following build semantics:

- When pushing to main or creating a PR: Code is linted and tests are run
- When pushing a tag: Code is linted and tests are run and if successful, multi-archtecture docker images are build and pushed to docker hub
- When manually triggered, a docker tag can be promoted to `latest`

### Assumptions

- Tags on the code will be of the format `v1.2.3` (semver) although without the v should work also.
- Tags on docker images should just be `1.2.3` (without the v).
- All code tags should result in a tagged docker image.

### Steps needed *before* merging this PR

- [ ]  Create docker hub account and add the credentials to github's secrets. The two names need to be: `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`. Use docker hub to generate an application token rather than using the account password. These two secrets go in Settings -> Secrets and Variables -> Actions -> Action Secrets.
- [x]  Change line 14 of `tag-image-as-latest.yml` to be your new docker hub account (and desired image name).
- [x]  Same as point 2, change line 20 of `on-tag.yml` to be your new docker hub account (and desired image name).

### During this PR I have made the following changes to your Dockerfile

1) Removed the need and deleted `build.sh`. Now the Dockerfile itself sets the required variables based on the target platform.
2) For ARM builds, I've changed the C compiler from `gcc` to `clang`. I couldn't get ring to compile under qemu from a x86 linux host for an arm target. I only found one other person on the internet who had also hit this (https://xkcd.com/979/).

### Limitations

- Building for arm under the qemu hypervisor on a standard github runner is extremely slow. It takes somewhere in the region of 65 minutes to build. To speed this up, run the job on a native arm runner, either one you host yourself or BuildJet offer arm runners. With a native arm runner, expect build times comparable with x86.
- To make a tag `latest` on Docker Hub is manual step. In github actions see workflow called "Push latest tag to docker hub". This workflow needs to be provided with which tag should be promoted. The tag needs to exist on docker hub before running this job which might be over an hour later from tagging the source code due to limitation above. Tagging latest can be done on a dev machine but please note you must use `buildx imagetools` to tag the image to keep the multi-architecture-ness of the image. Please just refer to the workflow for what command is being run.

### I have tested on my own fork

- Pushing to main
- Creating a PR and merging it to main
- Pushing a tag
- Maunally running the "Push latest tag to docker hub" workflow from the Github interface

I have pulled and run the built images on an arm Mac, an arm RaspberryPi and an x86 linux machine.

Let me know if you want me to squash the commits or if it will just be squashed on merge.


Extheoisah:
Hi @m3dwards,
thanks for the pr and a detailed explanation of how the process would work. However, I don't fully grasp the flow for the tag releases probably because of my lack of experience with docker hub for releases.
> Tagging latest can be done on a dev machine but please note you must use buildx imagetools to tag the image to keep the multi-architecture-ness of the image.

Here you stated that we need to use buildx imagetools to tag the images. is this going to be automated with github actions? if not, how do we go about this manually?

m3dwards:
@Extheoisah apologies if you are well familiar with all of this but I will explain a bit about docker tagging just to make sure we are on the same page. There are two types of tags that are commonly put on docker images on docker hub. A version number like `1.2.3` and the tag `latest`. 

When you pull or start an image you can optionally refer to the tag `docker run bdp/sim-ln:1.2.3` for example. If you don't specify a tag (`docker run bdp/sim-ln`) then docker will try to pull an image with the tag `latest` (same as `docker run bdp/sim-ln:latest`). Important thing to note here is that you already have an image with the tag latest pulled locally, docker won't fetch from docker hub even if there is now a later image with the tag `latest`. So it's often required to run `docker pull bdp/sim-ln:latest` to get the latest `latest` tag.

For creating images with a version number, these github actions will respond to a tag being put on the source code and will build and push images with the same tag. Pushing tag `v1.2.4` onto the source code should result with images being build and pushed as `<yourproject>/sim-ln:1.2.4` (without the v).

For making an image latest, this isn't automated but I have created a job that can be manually triggered from the github actions interface. You give it a tag and it will look on docker hub for an image of that tag and give it an additional tag of `latest`.

These two github actions use buildx to make and push these multi-architecture images.

For whatever reason someone may wish to create and push a docker image manually to docker hub outside of github. Or they may wish to retag an existing image to either make it `latest` or give it some arbitrary tag. Let's say I wanted to give an existing image with the tag `1.2.5` another tag of `maxs-special-tag`. The naive way to do this would be:

```shell
docker pull bdp/sim-ln:1.2.5
docker tag bdp/sim-ln:1.2.5 bdp/sim-ln:maxs-special-tag
docker push bdp/sim-ln:maxs-special-tag
```

The problem with doing the above with multi-architecture images is that you will only push an image with the tag `maxs-special-tag` for your specific architecture. Version `1.2.5` will still be multi-architecture but version `maxs-special-tag` will only be the architecture of the machine you ran those commands on.

What we probably want to do is tell docker and docker hub that all images regardless of the architecture for version `1.2.5` should also get the tag `maxs-special-tag` and to do that you use buildx imagetools:

```shell
docker buildx imagetools create -t bdp/sim-ln:latest bdp/sim-ln:1.2.5
```

To summarise: both tagging version numbers and latest can be done through this PR and github actions (versions are automated and latest is manually triggered). The comment was just a reminder that if you want to tag manually on your own dev machine and push images to use buildx imagetools rather than naive docker pull / tag / push.

.github/workflows/on-tag.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  test:
+    uses: ./.github/workflows/reusable/build-and-test.yml
+    secrets: inherit
+  docker-hub-image:
+    needs: [test]
+    uses: ./.github/workflows/reusable/push-docker-image.yml
+    secrets: inherit
+    with:
+      image_name: scalingln/simln
```
carlaKC, line 21: Can be updated to [bitcoindevproject](https://hub.docker.com/u/bitcoindevproject) 👍

.github/workflows/reusable/build-and-test.yml
```diff
@@ -1,25 +1,23 @@
-name: CI
+name: Build, lint and test
```
carlaKC: procedural nitpick: nice to have refactors/moves separated out into their own commit so that they're not mingled with changes, makes for happier review + bisecting. No need to change for this PR!

.github/workflows/on-tag.yml
```diff
@@ -11,11 +11,11 @@ permissions:
 
 jobs:
   test:
-    uses: ./.github/workflows/reusable/build-and-test.yml
+    uses: ./.github/workflows/build-and-test.yml
```
carlaKC, line 14: Can be squashed with https://github.com/bitcoin-dev-project/sim-ln/pull/149/commits/046ee5d978b1e115df19494c353212ab0c54ab22 👍

carlaKC (commented):
Thanks for the **great** PR! 

First pass (with my uneducated docker eyes) done - I think you can go ahead and squash into a few logical commits in the meantime (we don't squash on merge). I'll test on my fork in the meantime.

.github/workflows/on-tag.yml
m3dwards: I just squashed the lot

.github/workflows/on-tag.yml
m3dwards: Updated

.github/workflows/reusable/build-and-test.yml
m3dwards: Noted, however for this particular file was changed into a reusable workflow so that it could be used by the other workflows in this PR. Not sure that counts as a standalone refactor?

carlaKC (approved):
Successfully tested against my own fork!

Will add the secrets then merge 🚀

.github/workflows/reusable/build-and-test.yml
carlaKC: Ah fair, even if it's a bit of a change I like to keep moves in their own "logical chunk" but might have overlooked whether this specific change was actually suitable for that!

Extheoisah:
> @Extheoisah apologies if you are well familiar with all of this but I will explain a bit about docker tagging just to make sure we are on the same page. There are two types of tags that are commonly put on docker images on docker hub. A version number like `1.2.3` and the tag `latest`.
> 
> When you pull or start an image you can optionally refer to the tag `docker run bdp/sim-ln:1.2.3` for example. If you don't specify a tag (`docker run bdp/sim-ln`) then docker will try to pull an image with the tag `latest` (same as `docker run bdp/sim-ln:latest`). Important thing to note here is that you already have an image with the tag latest pulled locally, docker won't fetch from docker hub even if there is now a later image with the tag `latest`. So it's often required to run `docker pull bdp/sim-ln:latest` to get the latest `latest` tag.
> 
> For creating images with a version number, these github actions will respond to a tag being put on the source code and will build and push images with the same tag. Pushing tag `v1.2.4` onto the source code should result with images being build and pushed as `<yourproject>/sim-ln:1.2.4` (without the v).
> 
> For making an image latest, this isn't automated but I have created a job that can be manually triggered from the github actions interface. You give it a tag and it will look on docker hub for an image of that tag and give it an additional tag of `latest`.
> 
> These two github actions use buildx to make and push these multi-architecture images.
> 
> For whatever reason someone may wish to create and push a docker image manually to docker hub outside of github. Or they may wish to retag an existing image to either make it `latest` or give it some arbitrary tag. Let's say I wanted to give an existing image with the tag `1.2.5` another tag of `maxs-special-tag`. The naive way to do this would be:
> 
> ```shell
> docker pull bdp/sim-ln:1.2.5
> docker tag bdp/sim-ln:1.2.5 bdp/sim-ln:maxs-special-tag
> docker push bdp/sim-ln:maxs-special-tag
> ```
> 
> The problem with doing the above with multi-architecture images is that you will only push an image with the tag `maxs-special-tag` for your specific architecture. Version `1.2.5` will still be multi-architecture but version `maxs-special-tag` will only be the architecture of the machine you ran those commands on.
> 
> What we probably want to do is tell docker and docker hub that all images regardless of the architecture for version `1.2.5` should also get the tag `maxs-special-tag` and to do that you use buildx imagetools:
> 
> ```shell
> docker buildx imagetools create -t bdp/sim-ln:latest bdp/sim-ln:1.2.5
> ```
> 
> To summarise: both tagging version numbers and latest can be done through this PR and github actions (versions are automated and latest is manually triggered). The comment was just a reminder that if you want to tag manually on your own dev machine and push images to use buildx imagetools rather than naive docker pull / tag / push.

Thanks for the clarification @m3dwards 
Looks good and clean

PR #152, opened 2023-10-25
sim-lib: Get rid of nested json for node definition

sr-gi:
I must have been way more sleep-deprived than I thought when originally designing this, but we don't really need the nested json with node type tags to parse node information (as long as the node info has, **at least** one different field name).

Mainly, this goes from:

```json
"nodes": [
    {
        "LND": {
            "id": "...",
            "address": "...",
            "macaroon": "...",
            "cert": "..."
        }
    },
    {
        "CLN": {
            "id": "...",
            "address": "...",
            "ca_cert": "...",
            "client_cert": "...",
            "client_key": "..."
        }
    }
]
```

To:

```json
"nodes": [
    {
        "id": "...",
        "address": "...",
        "macaroon": "...",
        "cert": "..."
    },
    {
        "id": "...",
        "address": "...",
        "ca_cert": "...",
        "client_cert": "...",
        "client_key": "..."
    }
]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,33 @@ sim-cli sim.json
 The simulator requires access details for a set of `nodes` that you 
 have permission to execute commands on. Note that the current version 
 of the simulator uses keysend to execute payments, which must be 
-enabled in LND using `--accept-keysend`.
+enabled in LND using `--accept-keysend` (for CLN node it is enabled by default).
+
+The required access details will depend on the node implementation. For LND, the following
+information is required:
+
+```
+{
+  "id": <node_id>,
+  "address": https://<ip:port or domain:port>,
+  "macaroon": <path_to_selected_macaroon>,
+  "cert": <path_to_tls_cert>
+}
+```
+
+Whereas for CLN nodes, the following information is required:
+
+```
+{ 
+  "id": <node_id>,
+  "address": https://<ip:port or domain:port>,
+  "ca_cert": <path_to_ca_cert>,
+  "client_cert": <path_to_client_cert>,
+  "client_key": <path_to_client_key>
+}
+```
+
+**Note that node addresses must be declare with HTTPS transport, i.e. <https://ip-or-domain:port>**
 
 Payment activity can be simulated in two different ways:
 * Random activity: generate random activity on the `nodes` provided, 
@@ -85,8 +111,6 @@ not "drain" from the simulation.
   ]
 }
 ```
-
-**Note that node addresses must be declare with HTTPS transport, i.e. <https://ip-or-domain>**
 
 Nodes can be identified by an arbitrary string ("Alice", "CLN1", etc) or
 by their node public key. If a valid public key is provided it *must* 
--- a/README.md
+++ b/README.md
@@ -70,21 +70,17 @@ not "drain" from the simulation.
 {
   "nodes": [
     {
-      "LND": {
-        "id": "Alice",
-        "address": "https://127.0.0.1:10011",
-        "macaroon": "/path/admin.macaroon",
-        "cert": "/path/tls.cert"
-      }
+      "id": "Alice",
+      "address": "https://127.0.0.1:10011",
+      "macaroon": "/path/admin.macaroon",
+      "cert": "/path/tls.cert"
     },
-    {
-      "CLN": {
-        "id": "0230a16a05c5ca120136b3a770a2adfdad88a68d526e63448a9eef88bddd6a30d8",
-        "address": "https://localhost:10013",
-        "ca_cert": "/path/ca.pem",
-        "client_cert": "/path/client.pem",
-        "client_key": "/path/client-key.pem"
-      }
+    { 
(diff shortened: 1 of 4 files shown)


sim-lib/src/lib.rs
```diff
@@ -25,10 +25,9 @@ mod random_activity;
 mod serializers;
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(untagged)]
```
okjodom, line 28: TIL of this magic! 🥇

okjodom (approved):

sim-lib/src/lib.rs
sr-gi: Oh yeah, there's a bunch of these you can use depending on what you're aiming for:

https://serde.rs/enum-representations.html
https://serde.rs/attr-flatten.html

sim-cli/src/main.rs
```diff
@@ -60,7 +61,8 @@ async fn main() -> anyhow::Result<()> {
         .unwrap();
 
     let SimParams { nodes, activity } =
-        serde_json::from_str(&std::fs::read_to_string(cli.sim_file)?)?;
+        serde_json::from_str(&std::fs::read_to_string(cli.sim_file)?)
+            .map_err(|e| anyhow!("Cannot deserialize node connection data into LndConnection nor ClnConnection (line {}, col {})", e.line(), e.column()))?;
```
sr-gi, line 65: I just realised this may not be good, given this can error out both for node parsing and activity parsing

README.md
```diff
@@ -70,21 +70,17 @@ not "drain" from the simulation.
 {
   "nodes": [
     {
-      "LND": {
-        "id": "Alice",
-        "address": "https://localhost:10011",
-        "macaroon": "/path/admin.macaroon",
-        "cert": "/path/tls.cert"
-      }
+      "id": "Alice",
```
carlaKC, line 73: I think that we need to expand the readme with an explanation of these different formats for CLN or LND - this isn't very easy for newcomers to understand. 

My instinct would be to have a `### Setup - Node Connection`  section that shows what it looks like for CLN and LND?

sr-gi: I agree. Should we do this here or in a followup?

carlaKC: I think it's generally a good idea to keep code changes in the same PR as their documentation so that we don't forget (and then docs get out of sync).

carlaKC: And in this case, needs a rebase in the readme anyway so let's def do it in the same PR.

carlaKC:
Let's get this done with high priority if we're going to change it, told the warnet folks to wait on a stable config file before they proceed.

sr-gi:
> Let's get this done with high priority if we're going to change it, told the warnet folks to wait on a stable config file before they proceed.

On it

README.md
```diff
@@ -50,7 +50,31 @@ sim-cli sim.json
...
-enabled in LND using `--accept-keysend`.
+enabled in LND using `--accept-keysend` (for CLN node it is enabled by default).
+
+The required access details will depend on the node implementation. For LND, the following
+information is required:
+
+```
+{
+  "id": <node_id>,
+  "address": <ip_and_port_or_domain_and_port>,
```
carlaKC, line 61: nit: include `https://`

sim-cli/src/main.rs
carlaKC: Maybe just make the error more holistic? 
`Could not deserialize node connection (LND or CLN) or activity description: (line {}, col {}) in simulation file.`

Most folks should be able to figure it out based on the line). Perhaps also include the error string if there is one? Since it's often `hanging ,` or some other annoying json error.

carlaKC (commented):
Just nits

carlaKC (approved):

README.md
sr-gi: Should we just move this line up? https://github.com/bitcoin-dev-project/sim-ln/blob/76c8ced0724a8a6c7989e47fef03e141b24e8221/README.md?plain=1#L113

sim-cli/src/main.rs
sr-gi: I like your version better.

Regarding including the original string, there's actually no meaningful info there, that's why I decided to go with a custom error. 

```
Error: data did not match any variant of untagged enum NodeConnection at line 11 column 8
```

Not even the line and column are that useful, they mark the end of the section where the error is at, not even the real line :/

README.md
carlaKC: Works for me! But I'd still hard code it in before to get as close as possible - `https://<ip:port or domain:port>`

sim-cli/src/main.rs
carlaKC: Hm ok, let's just get as user friendly as we can and ship it.

sr-gi:
@carlaKC should be good now

README.md
sr-gi: Ok, I'm going to open an issue so we don't require this, and we can do that in a low-prio followup

carlaKC (approved):
🧹

Issue #40, opened 2023-08-17
Error formatting: Go over the displayed errors and make them more user friendly


sr-gi:
Just as an example, if a destination node does not support keysend, this is the error we'll be getting:

```
Error: Lightning Error: ValidationError("Destination node does not support keysend 02ab4734430c485b1d15af93ff00060db86f7fe952b04c19afd42709dcd83b1c50")

Caused by:
    Config validation failed Destination node does not support keysend 02ab4734430c485b1d15af93ff00060db86f7fe952b04c19afd42709dcd83b1c50
```

AdamuAbba:
Hey @carlaKC is it fine if I pick this up?

AdamuAbba:
hey @carlaKC regarding the error messages based on  @sr-gi's example above would any of the formats below be acceptable? I'll implement whichever one you pick.

```bash
Error: Lightning Error
Caused by:
    Config validation failed, The destination node does not support keysend 02ab4734430c485b1d15af93ff00060db86f7fe952b04c19afd42709dcd83b1c50
```
or do we go with just:

```
Caused by:
    Config validation failed Destination node does not support keysend 02ab4734430c485b1d15af93ff00060db86f7fe952b04c19afd42709dcd83b1c50
```
or maybe:
```
Error: Lightning Error
Type: Validation Error
Caused by: The destination node does not support keysend 02ab4734430c485b1d15af93ff00060db86f7fe952b04c19afd42709dcd83b1c50
```
we can also consider:
```
Error: Lightning Error
Type: Validation Error
Caused by: The destination node does not support keysend
Destination: 02ab4734430c485b1d15af93ff00060db86f7fe952b04c19afd42709dcd83b1c50
```

standing by for your decision.

carlaKC:
This one is fine to me, no strong opinions though: 
```
Error: Lightning Error
Caused by:
   Config validation failed Destination node does not support keysend 02ab4734430c485b1d15af93ff00060db86f7fe952b04c19afd42709dcd83b1c50
```

AdamuAbba:
Alright got it.

carlaKC:
One place where we could do better on errors is when we parse our [sim file](https://github.com/bitcoin-dev-project/sim-ln/blob/main/sim-cli/src/main.rs#L92) - right now we just display a line an column, not the actual error. 



[...]

PR #153, opened 2023-10-25
sim-lib: Allows amount and interval ranges in manual activity definition

### LINKED ISSUE #43: Feature: Make the input file format allow ranges in the frequency (interval) and amounts
We are currently only allowing fixed values, it'll be good to have the option to define ranges (and potentially also distributions down the line).

sr-gi:
Values in activities can now be fixed values (e.g. 1000) or ranges (e.g. [500, 1000]).

If a range is provided, the value will be sampled uniformly at random from it.

Close #43

### DIFF
--- a/sim-lib/src/defined_activity.rs
+++ b/sim-lib/src/defined_activity.rs
@@ -1,4 +1,6 @@
-use crate::{DestinationGenerator, NodeInfo, PaymentGenerationError, PaymentGenerator};
+use crate::{
+    DestinationGenerator, NodeInfo, PaymentGenerationError, PaymentGenerator, ValueOrRange,
+};
 use std::fmt;
 use tokio::time::Duration;
 
@@ -7,8 +9,8 @@ pub struct DefinedPaymentActivity {
     destination: NodeInfo,
     start: Duration,
     count: Option<u64>,
-    wait: Duration,
-    amount: u64,
+    wait: ValueOrRange<u16>,
+    amount: ValueOrRange<u64>,
 }
 
 impl DefinedPaymentActivity {
@@ -16,8 +18,8 @@ impl DefinedPaymentActivity {
         destination: NodeInfo,
         start: Duration,
         count: Option<u64>,
-        wait: Duration,
-        amount: u64,
+        wait: ValueOrRange<u16>,
+        amount: ValueOrRange<u64>,
     ) -> Self {
         DefinedPaymentActivity {
             destination,
@@ -33,7 +35,7 @@ impl fmt::Display for DefinedPaymentActivity {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
-            "static payment of {} to {} every {:?}",
+            "static payment of {} to {} every {}s",
             self.amount, self.destination, self.wait
         )
     }
@@ -55,7 +57,7 @@ impl PaymentGenerator for DefinedPaymentActivity {
     }
 
     fn next_payment_wait(&self) -> Duration {
-        self.wait
+        Duration::from_secs(self.wait.value() as u64)
     }
 
     fn payment_amount(
@@ -67,7 +69,7 @@ impl PaymentGenerator for DefinedPaymentActivity {
                 "destination amount must not be set for defined activity generator".to_string(),
             ))
         } else {
-            Ok(self.amount)
+            Ok(self.amount.value())
         }
     }
 }
@@ -75,9 +77,9 @@ impl PaymentGenerator for DefinedPaymentActivity {
 #[cfg(test)]
 mod tests {
     use super::DefinedPaymentActivity;
+    use super::*;
     use crate::test_utils::{create_nodes, get_random_keypair};
     use crate::{DestinationGenerator, PaymentGenerationError, PaymentGenerator};
-    use std::time::Duration;
 
     #[test]
     fn test_defined_activity_generator() {
@@ -91,8 +93,8 @@ mod tests {
             node.clone(),
             Duration::from_secs(0),
             None,
-            Duration::from_secs(60),
-            payment_amt,
+            crate::ValueOrRange::Value(60),
+            crate::ValueOrRange::Value(payment_amt),
         );
 
         let (dest, dest_capacity) = generator.choose_destination(source.1);
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -4,6 +4,7 @@ use bitcoin::Network;
 use csv::WriterBuilder;
 use lightning::ln::features::NodeFeatures;
 use lightning::ln::PaymentHash;
+use rand::Rng;
 use random_activity::RandomActivityError;
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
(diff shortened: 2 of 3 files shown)


carlaKC:
I think that we should get some refactoring in before we make this type of change. 

`produce_events` and `produce_random` events should really just be one function that takes a `NetworkGenerator` and `PaymentGenerator`. I've started on that refactor, and we're going to have some significant churn if we do this piece first. Since this isn't _very_ high prio I'd like to hold off till that's in if it's okay.

sr-gi:
Sure, no rush on my end

sr-gi:
@carlaKC I saw you tagged #43 for the V3 release. I'm happy to rework this if you find it useful

carlaKC:
> @carlaKC I saw you tagged #43 for the V3 release. I'm happy to rework this if you find it useful

That would be great!

sim-lib/src/defined_activity.rs
```diff
@@ -37,7 +39,11 @@ impl DestinationGenerator for DefinedPaymentActivity {
 
 impl PaymentGenerator for DefinedPaymentActivity {
     fn next_payment_wait(&self) -> Duration {
-        self.wait
+        let wait = Duration::from_secs(self.wait.value() as u64);
```
sr-gi, line 42: This is the only bit I'm not super happy about, given we'll be building a `Duration` object every time we call `next_payment_wait`, even for wait payments. However, changing this would imply changing the generic for ValueOrRange, and I think potentially more code bolilerplate.

I don't think this is too bad, but I'm open to change it if you find it is.

sim-lib/src/defined_activity.rs
```diff
@@ -49,7 +55,11 @@ impl PaymentGenerator for DefinedPaymentActivity {
                 "destination amount must not be set for defined activity generator".to_string(),
             ))
         } else {
-            Ok(self.amount)
+            let amount = self.amount.value();
+            if self.amount.is_range() {
+                log::debug!("Resampling payment amount. New value: {}", amount);
```
carlaKC, line 60: I think that we can get away without these logs because we log the payment wait/amount at callsites?

sim-lib/src/defined_activity.rs
carlaKC: We could have a `RangePaymentActivity` impl of `PaymentGenerator` which deals exclusively with the range case?

We can stil parse everything as-is here, but then create either `RangePaymentActivity` or `DefinedPaymentActivity` when we [switch over from](https://github.com/bitcoin-dev-project/sim-ln/blob/main/sim-cli/src/main.rs#L185) `ActiviyParser` -> `ActivityDefinition`. Might cause a little bit of trait object agony in `main`, but we _should_  be mostly okay in `lib` 🤞 

That might also be quite boiler-platey, but could be worth considering!

carlaKC (commented):
Thanks for picking this one back up!

sim-lib/src/defined_activity.rs
sr-gi: Yeah, I think it is pretty redundant atm. e.g.:

```
2024-04-05T10:00:17.332Z DEBUG [sim_lib::defined_activity] Resampling payment amount. New value: 216731
2024-04-05T10:00:17.332Z DEBUG [sim_lib] Generated random payment: alice(03628c...04b66e) -> bob(025e30...d26140): 216731 msat.
2024-04-05T10:00:17.332Z DEBUG [sim_lib::defined_activity] Resampling interval. New value: 7
2024-04-05T10:00:17.332Z DEBUG [sim_lib] Next payment for alice(03628c...04b66e) in 7s seconds.
```

sim-lib/src/defined_activity.rs
sr-gi: I don't know if that would be even more complex.

We have two fields that can be either a range or a value (but only one of them has the initialization issue), how do we deal with the combination of these?

`DefinedPaymentActivity` would need to have `wait: Duration, amount: ValueOrRange<u64>` while `RangePaymentActivity` would have `wait: ValueOrRange<u16>, amount: ValueOrRange<u64>`. That's really counterintuitive. The alternative would be to only allow either both or none to be ranges, but that's limiting the functionality.

None of the solutions I've been trying to come up with have been satisfying for one reason or another (nor even having a two-argument generic, since deserialization becomes a mess). I've been checking the constructor of `Duration` and, honestly, it doesn't look that bad (specially considering that we are constructing with `nanos=0`): https://doc.rust-lang.org/src/core/time.rs.html#199-207

So I would bite the bullet with this one, but I'm happy to be convinced of an alternative solution if you could find something I'm missing

sim-lib/src/serializers.rs
```diff
@@ -45,6 +45,42 @@ pub mod serde_node_id {
...
+
+    pub fn deserialize<'de, D, T>(deserializer: D) -> Result<ValueOrRange<T>, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+        T: serde::Deserialize<'de> + std::cmp::PartialOrd + std::fmt::Display + Copy,
+    {
+        let a = ValueOrRange::deserialize(deserializer)?;
+        if let ValueOrRange::Range(x, y) = a {
+            if x >= y {
+                return Err(D::Error::custom(format!(
```
bjohnson5, line 73: This error message is very helpful, but it is not shown when a deserialization error occurs. I think this is because where `deserialize` is called (in main) the error string is ignored.

sim-lib/src/lib.rs
```diff
@@ -129,6 +130,52 @@ pub struct SimParams {
...
+            ValueOrRange::Value(x) => *x,
+            ValueOrRange::Range(x, y) => {
+                let mut rng = rand::thread_rng();
+                rng.gen_range(*x..*y)
+            },
+        }
+    }
+
+    /// Whether this is a range or not
+    pub fn is_range(&self) -> bool {
```
bjohnson5, line 157: nit: I believe this is an unused function?

bjohnson5 (commented):
@sr-gi Hello! I am a new contributor trying to learn my way around `sim-ln`. Helping out with reviews right now and @carlaKC mentioned this one would be good to look at. I tested out using ranges and it works great, this will be a very useful feature. Left a few small comments.

sim-lib/src/defined_activity.rs
carlaKC: Managed to completely miss this comment - sorry!! 

> DefinedPaymentActivity would need to have wait: Duration, amount: ValueOrRange<u64> while RangePaymentActivity would have wait: ValueOrRange<u16>, amount: ValueOrRange<u64>. That's really counterintuitive. The alternative would be to only allow either both or none to be ranges, but that's limiting the functionality.

Good point, didn't think about the inbetween case where we only want a range for one of them! Agreed, let's just go ahead w/ what we have here.

sr-gi: np, I'll update it with the most recent comments

sim-lib/src/lib.rs
sr-gi: Oh yeah, this was part of the previous approach: https://github.com/bitcoin-dev-project/sim-ln/blob/cc19a99e50459cd1e1ba790df3e74fd44b2af66b/sim-lib/src/lib.rs#L829-L837

Will remove it

sim-lib/src/lib.rs
```diff
@@ -146,9 +193,10 @@ pub struct ActivityParser {
     #[serde(default)]
     pub count: Option<u64>,
     /// The interval of the event, as in every how many seconds the payment is performed.
-    pub interval_secs: u16,
+    pub interval_secs: Interval,
```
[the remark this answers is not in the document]
sr-gi: Yep, you're right!

sr-gi:
CI failure is unrelated

carlaKC:
> CI failure is unrelated

Fixed in https://github.com/bitcoin-dev-project/sim-ln/pull/186

sr-gi:
Rebased to make CI green again. 

This should be review-ready.

sim-lib/src/serializers.rs
carlaKC: Added note to error-cleanup issue: https://github.com/bitcoin-dev-project/sim-ln/issues/40#issuecomment-2120685953

sim-lib/src/defined_activity.rs
```diff
@@ -67,17 +69,17 @@ impl PaymentGenerator for DefinedPaymentActivity {
...
-            Ok(self.amount)
+            Ok(self.amount.value())
         }
     }
 }
 
 #[cfg(test)]
 mod tests {
     use super::DefinedPaymentActivity;
+    use super::*;
```
carlaKC, line 80: nit: can remove `DefinedPaymentActivity` if we're using `*`

carlaKC (approved):
tACK, such a nice feature to add 😻

bjohnson5 (approved):

PR #156, opened 2023-11-01
sim-cli: improves capacity_multiplier deserializer

sr-gi:
The current `capacity_multiplier` deserializer accepts values smaller than or equal to 0, which are not allowed by the simulator. Patch it with a custom f64 deserializer that enforcer values to be greater than zero

This is related to #155.

### DIFF
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -5,6 +5,7 @@ use std::sync::Arc;
 use tokio::sync::Mutex;
 
 use anyhow::anyhow;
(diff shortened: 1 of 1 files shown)


sr-gi:
No rush merging this, opening it just so we don't forger about it

carlaKC (approved):
tACK ✨ 

`error: invalid value '0' for '--capacity-multiplier <CAPACITY_MULTIPLIER>': capacity_multiplier must be higher than 0. 0 received.`

PR #163, opened 2024-01-26
docker: add localhost replacement to volume setup

carlaKC:
Inspired by testing #159 and re-feelin the magic of `sim.json` files being magically converted to work with docker ✨ 

This PR just replaces `localhost` -> `host.docker.internal` since it's pretty likely that everything is being used on the same machine if you're using localhost. With this change, you don't need to change your local `sim.json` to use docker _at all_!


Extheoisah (approved):
Pretty straight-forward

carlaKC:
Pretty low risk so merging this now, post-merge testing would be appreciated!

PR #164, opened 2024-01-27
refactor: use listen and trigger universally

### LINKED ISSUE #162: Robustness: Use Listen and Shutdown Trigger Universally
Right now, we [rely on some of the mechanics](https://github.com/bitcoin-dev-project/sim-ln/blob/main/docs/ARCHITECTURE.md#shutdown) of `mpsc` channels to handle shutdown in places.  We should move the code base over to using `trigger` to handle shutdown everywhere to improve readability. 

## 1. Listen for shutdown in channel send/recieve 
`select!` on `listener` for _every_ channel `Send` and `Receive`, exiting if we get the signal to shut down 

Eg: rather than [send](https://github.com/bitcoin-dev-project/sim-ln/blob/main/sim-lib/src/lib.rs#L767C51-L767C51), select on listener to abort send if we need to exit:
```
select! {
    biased;
    _ = listener.clone() => {
        // log something here
        break;
    },
    send_res = sender.send(outcome) => {
        match send_res {
            Ok(_) => {},
            Err(e) => {
                log::error!("Error sending action outcome: {:?}.", e);
                break;
            }
        }
    },
}

```

## 2. Always trigger shutdown on error

Call `trigger` to signal shutdown if any task exits with an error - eg [results consumer](https://github.com/bitcoin-dev-project/sim-ln/blob/main/sim-lib/src/lib.rs#L889) and [event consumer](https://github.com/bitcoin-dev-project/sim-ln/blob/main/sim-lib/src/lib.rs#L713). Run through codebase to cover all of these cases.

## 3. Doc Update
The [shutdown](https://github.com/bitcoin-dev-project/sim-ln/blob/main/docs/ARCHITECTURE.md#shutdown) section of the docs should be updated to reflect the simplified handling as part of this issue.

enigbe:
## What this PR does
- [x] Listens for shutdown trigger across all channels
- [x] Trigger shutdown when tasks exit with error
- [x]  Documentation update

## Related Issue(s)
- Resolves #162 

## Notes
- Branched off [refactor-dispatch](https://github.com/carlaKC/sim-ln/tree/refactor-dispatch) given its advanced state of review.


sim-lib/src/lib.rs
```diff
@@ -1077,8 +1179,20 @@ async fn track_payment_result(
...
-        log::debug!("Could not send payment result.");
+    select! {
+        biased;
+        _ = listener.clone() => {
+            log::debug!("Track payment received a shutdown signal")
+        },
+        send_payment_result = results.send((payment, res)) => {
+            match send_payment_result {
+                Ok(_) => {},
+                Err(e) => {
```
carlaKC, line 1190: Can we still use `is_err` here since we don't care about the `OK` branch in the match? LIkewise elsewhere in this PR.

sim-lib/src/lib.rs
```diff
@@ -1034,6 +1134,7 @@ async fn produce_simulation_results(
     while let Some(res) = set.join_next().await {
         if let Err(e) = res {
             log::error!("Simulation results producer task exited with error: {e}.");
+            shutdown.trigger();
```
carlaKC, line 1137: Do we need to trigger shutdown here? Looks like the loop above only exits on `listener.clone`, so we've already had a shutdown trigger elsewhere?

sim-lib/src/lib.rs
```diff
@@ -994,29 +1077,46 @@ async fn produce_simulation_results(
...
             output = output_receiver.recv() => {
                 match output{
                     Some(simulation_output) => {
                         match simulation_output{
                             SimulationOutput::SendPaymentSuccess(payment) => {
-                                let source_node = nodes.get(&payment.source).unwrap().clone();
-                                set.spawn(track_payment_result(
-                                    source_node, results.clone(), payment, shutdown.clone(),
-                                ));
+                                if let Some(source_node) = nodes.get(&payment.source) {
```
carlaKC, line 1098: If we're getting rid of the `unwrap` here, we should error out if we can' t find the source node, right? Because we've generated this successful payment from a node that we don't have in our node set.

sim-lib/src/lib.rs
```diff
@@ -994,29 +1077,46 @@ async fn produce_simulation_results(
...
+                                    ));
+                                }
                             },
                             SimulationOutput::SendPaymentFailure(payment, result) => {
-                                if results.clone().send((payment, result)).await.is_err() {
-                                    log::debug!("Could not send payment result.");
+                                select! {
+                                    biased;
+                                    _ = listener.clone() => {
+                                        log::debug!("Handling for failed payment {payment} received a shutdown signal");
```
carlaKC, line 1108: nit: we're ending log lines with a `.` by convention in the repo

sim-lib/src/lib.rs
```diff
@@ -733,65 +751,101 @@ async fn consume_events(
...
-                                log::error!("Simulation terminated with error: {s}.");
-                                shutdown.trigger();
-                                break;
+    loop {
+        select! {
+            biased;
+            _ = listener.clone() => {
+                let node = node.lock().await;
+                let node_info = node.get_info();
+                log::debug!("Simulation event consumer for {node_info} received shutdown signal");
```
carlaKC, line 764: nit: inline `log::debug!("Simulation event consumer for {} received shutdown signal", node.lock().await.get_info());`

sim-lib/src/lib.rs
```diff
@@ -733,65 +751,101 @@ async fn consume_events(
...
+                break;
+            },
+            simulation_event = receiver.recv() => {
+                match simulation_event {
+                    Some(event) => {
```
carlaKC, line 769: `if let Some(event) = simulation_event` / `else return`?

carlaKC (commented):
Thanks for the PR! A few structural/stylistic comments on first pass. 

One high level question that I'm wondering about is whether we can simplify this by always triggering `shutdown` whenever we break one of the loop/select patterns in the code? Right now we have to inline on shutdown, but if the expected behavior is to break the loop on error we could just have a single trigger on break? Perhaps with the exception of `listener` triggering, because we know we can just return there. 

I think that _could_ lead to simpler understanding of the shutdown logic (which is a nightmare right now), but I haven't looked at whether it works in every instance across the codebase. Interested to hear thoughts on this from others!

.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules
 activity-generator/releases/*
 .DS_Store
 /results
+.vscode
```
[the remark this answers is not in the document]
okjodom: Pls add to your dev machine .gitignore. On my linux device, this is located at `~/.gitignore`

carlaKC:
Discussed this PR a little more offline. Another option for a cleaner/more testable solution would be: 
* Refactor to return `Result` from functions that can trigger shutdown
* Call `trigger` at spawn site rather than inside of the function

This saves us from having to pass shutdown all the way down to every task, and cuts down on the number of places where we need to call `trigger`. Also has the benefit of making some of these functions more testable, because we can assert on return values. 

Eg, for our simulation results task:
```
        tasks.spawn(async move {
            if let Err(e) =
                produce_simulation_results(nodes, output_receiver, results_sender, listener_results)
                    .await
            {
                shutdown.trigger();
                log::error!("produce simulation results exited with error: {e:?}.");
            }
        });
```

carlaKC:
Can be rebased on #160 and sincerest apolgies in advance for all the rebase conflicts :')

Cargo.lock
```diff
@@ -1734,7 +1666,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "async-trait",
- "bitcoin 0.30.1",
+ "bitcoin 0.30.2",
```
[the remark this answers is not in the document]
enigbe: Same as  [comment](https://github.com/bitcoin-dev-project/sim-ln/pull/164#discussion_r1478235066) above

sim-lib/src/lib.rs
enigbe: In hindsight, it was not necessary. I have update the handling of the error here (and across all tasks) to follow the recommendation [here](https://github.com/bitcoin-dev-project/sim-ln/pull/164#issuecomment-1919135888).

sim-lib/src/lib.rs
enigbe: Updated to handle this possibility.

enigbe:
> Discussed this PR a little more offline. Another option for a cleaner/more testable solution would be:
> 
>     * Refactor to return `Result` from functions that can trigger shutdown
> 
>     * Call `trigger` at spawn site rather than inside of the function
> 
> 
> This saves us from having to pass shutdown all the way down to every task, and cuts down on the number of places where we need to call `trigger`. Also has the benefit of making some of these functions more testable, because we can assert on return values.
> 
> Eg, for our simulation results task:
> 
> ```
>         tasks.spawn(async move {
>             if let Err(e) =
>                 produce_simulation_results(nodes, output_receiver, results_sender, listener_results)
>                     .await
>             {
>                 shutdown.trigger();
>                 log::error!("produce simulation results exited with error: {e:?}.");
>             }
>         });
> ```

I have refactored the handling of errors across all tasks. This reduces the number of places `trigger()` can be called and streamlines the logic. However, I think we can further reduce the call to `trigger()` to just one location.

From the snippet from [sim-lib/lib.rs::run](https://github.com/bitcoin-dev-project/sim-ln/blob/3050a6dcb719d94d4f930c976ae4858c7c05ef07/sim-lib/src/lib.rs#L519-L520) below, we await the completion of all tasks in the join set `tasks`. The error from the first task failure can be propagated until it gets here, in which we call `trigger()` (and `break` out of the loop) to shut down all listening tasks in the set.

```rust
            if let Err(e) = res {
                log::error!("Task exited with error: {e}.");
                success = false;
            }
        }
```

This would mean we have to have another loop waiting for all tasks to exit (post `trigger()`) to have a graceful shutdown. Uncertain if this is a good idea/approach and happy to get thoughts on this.

docs/ARCHITECTURE.md
```diff
@@ -52,38 +54,36 @@ output on to the next one.
...
-multiple asynchronous tasks at once. These selects should be `biased` 
-on the exit case (ie, the `Listener` being triggered) so that we 
+
+To ensure that all tasks shutdown on completion (or failure) of the
+simulation, we relay on the following:
+
+1. [Triggered](https://docs.rs/triggered/latest/triggered): a `Trigger`
+   that can be used to inform threads that it's time to shut down, and
+   a `Listener` that propagates this signal.
+2. The (`Trigger`, `Listener`) pair are used with channels: if a channel errors out across `send()` or `receive()`, shutdown is triggered. There is no reliance on channel mechanics, i.e. errors generated when all senders are and/or a receiver is dropped.
```
carlaKC, line 64: I would clarify this to note that our channels don't have buffers, so we have to select on the shutdown signal on send/receive to make sure that a receiver exiting before a sender does doesn't indefinitely hang. 

Ie, this scenario: 

Task 1: Sending into `sender`
Task 2: Receiving on `receiver`

* Task 1 sends into `sender`, unblocks due to buffer size of 1 that we use everywhere
* Task 2 errors out before consuming from `receiver`
* Task 1 wants to send into `sender` again, but can't because the `receiver` has shut down 

If we always select on exit, then we don't run into this.

docs/ARCHITECTURE.md
```diff
@@ -52,38 +54,36 @@ output on to the next one.
...
 prioritize exit above generating more events.
 
-Practically, this means that we're handling the various shutdown 
-scenarios in the following way: 
+Practically, this means that we're handling the various shutdown
+scenarios in the following way:
+
+1. Consumer error:
+
+- Receiving channel is dropped on consumer exit.
```
carlaKC, line 76: We shouldn't be relying on this now? We can probably cut this whole section now that we're just using listener/trigger everywhere.

sim-lib/src/lib.rs
```diff
@@ -290,11 +298,19 @@ struct Payment {
 
 impl Display for Payment {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let dispatch_time = self
+            .dispatch_time
+            .duration_since(UNIX_EPOCH)
+            .expect(&format!(
+                "Failed to compute duration since unix epoch: {:?}",
+                UNIX_EPOCH
```
carlaKC, line 306: nit: `UNIX_EPOCH` doesn't really need to be included in the panic message because it's a constant.

sim-lib/src/lib.rs
```diff
@@ -451,7 +468,7 @@ impl Simulation {
             }
         }
 
-        log::info!("Simulation is running on {}.", running_network.unwrap());
+        log::info!("Simulation is running on {}.", running_network.expect("Invalid network"));
```
carlaKC, line 471: nit: Network not provided rather than invalid?

sim-lib/src/lib.rs
```diff
@@ -558,12 +585,20 @@ impl Simulation {
...
+        let csr_write_results = self.write_results.clone();
+        tasks.spawn(async move {
+            if let Err(e) = consume_simulation_results(
+                result_logger,
+                results_receiver,
+                listener,
+                csr_write_results,
+            ).await {
+            shutdown.trigger();
+            log::error!("Consume simulation results exited with error: {e:?}.");
```
carlaKC, line 598: nit: formatting looks off here?

sim-lib/src/lib.rs
```diff
@@ -674,12 +709,22 @@ impl Simulation {
...
-                output_sender.clone(),
-                self.shutdown_trigger.clone(),
-            ));
+            // ce: consume event
+            let ce_listener = self.shutdown_listener.clone();
+            let ce_shutdown = self.shutdown_trigger.clone();
+            let ce_output_sender = output_sender.clone();
+            let ce_node = node.clone();
+            tasks.spawn(async move { 
+                if let Err(e) = consume_events(
```
carlaKC, line 718: nice to have but non-blocking: a lot of these functions that we spawn in tasks have a "function_name starting" debug log - would be nice to move that up to this level as well. 

Could also update the `if let` to a `match` and log when it exits without error (since we have that in the function as well).

sim-lib/src/lib.rs
```diff
@@ -701,89 +746,107 @@ impl Simulation {
...
-                    amount_msat: amt_msat,
-                    destination: dest.pubkey,
-                    dispatch_time: SystemTime::now(),
-                };
+    loop {
+        select! {
+            biased;
+            _ = listener.clone() => {
+                log::debug!("Simulation event consumer for node {} received shutdown signal", node.lock().await.get_info());
+                break;
```
carlaKC, line 786: Return `OK` here and then we don't need it below?

sim-lib/src/lib.rs
```diff
@@ -826,41 +888,53 @@ async fn produce_events<N: DestinationGenerator + ?Sized, A: PaymentGenerator +
...
                 }
+
             },
         }
     }
 
     log::debug!("Stopped random activity producer {source}.");
-    shutdown.trigger();
+    
+    Ok(())
```
carlaKC, line 912: Likewise here replace break above with a return?

Works well with the option of moving start/stop logging to the task spawn site.

sim-lib/src/lib.rs
```diff
@@ -826,41 +888,53 @@ async fn produce_events<N: DestinationGenerator + ?Sized, A: PaymentGenerator +
...
     write_results: Option<WriteResults>,
-) {
+) -> Result<(), SimulationError> {
     log::debug!("Simulation results consumer started.");
 
-    if let Err(e) = write_payment_results(logger, receiver, listener, write_results).await {
-        log::error!("Error while reporting payment results: {:?}.", e);
+    select! {
+        biased;
+        _ = listener.clone() => {
```
carlaKC, line 925: Do we need to add a listener at this level if we've already got one in `write_payment_results`? 

Looking at this wish fresh eyes, not sure why we have two levels of functions here - can't we just `spawn consume_simulation_results` and move the code for `write_payment_results` into this function?

carlaKC (commented):
vrrynice. Just some annoying nitpicking about logging and breaking from me, I think this is almost there! 

Only major comment is about cleaning up the result writer functions a bit, but that's pre-existing.

sim-lib/src/lib.rs
enigbe: `write_payment_results` has been moved to `consume_simulation_results`. Marking as resolved but reopen if you have any reservations.

sim-lib/Cargo.toml
```diff
@@ -12,11 +12,14 @@ Backend logic of the SimLN project. Contains all the functionality to build your
 anyhow = { version = "1.0.69", features = ["backtrace"] }
 cln-grpc = "0.1.3"
 expanduser = "1.2.2"
-serde = { version="1.0.183", features=["derive"] }
+serde = { version = "1.0.183", features = ["derive"] }
```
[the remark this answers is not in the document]
enigbe: Resolving this following #167 merge

docs/ARCHITECTURE.md
carlaKC: This comment still applies, IMO we can just delete the whole paragraph under "practically, this means that we're handling"...

docs/ARCHITECTURE.md
```diff
@@ -52,38 +54,37 @@ output on to the next one.
...
-multiple asynchronous tasks at once. These selects should be `biased` 
-on the exit case (ie, the `Listener` being triggered) so that we 
+
+To ensure that all tasks shutdown on completion (or failure) of the
+simulation, we relay on the following:
+
+1. [Triggered](https://docs.rs/triggered/latest/triggered): a `Trigger`
+   that can be used to inform threads/tasks that it's time to shut down, and
+   a `Listener` that propagates this signal.
+2. The (`Trigger`, `Listener`) pair are used with channels: if a channel errors out across `send()` or `recv()`, shutdown is triggered. There is no reliance on channel mechanics, i.e. errors generated when all senders are and/or a receiver is dropped.
```
carlaKC, line 64: nit: wrap text at 80 chars

We say 120 in contributing I think, but this file is already at 80 :')

docs/ARCHITECTURE.md
```diff
@@ -52,38 +54,37 @@ output on to the next one.
...
-  dropped (in our context, all of the producers have exited) then the 
-  corresponding receiving channel will error out on receiving. Likewise, 
-  if the receiving channel is dropped, the sending channels will error 
-  out.
-
-All events are handled in a `tokio::select` to allow waiting on 
-multiple asynchronous tasks at once. These selects should be `biased` 
-on the exit case (ie, the `Listener` being triggered) so that we 
+
+To ensure that all tasks shutdown on completion (or failure) of the
```
carlaKC, line 58: Very nice doc updates all around!

sim-lib/src/lib.rs
```diff
@@ -538,32 +555,58 @@ impl Simulation {
...
+        let result_logger_clone = result_logger.clone();
+        let result_logger_listener = listener.clone();
+        tasks.spawn(async move {
+            log::debug!("Starting results logger.");
+            run_results_logger(
+                result_logger_listener,
+                result_logger_clone,
+                Duration::from_secs(60),
+            )
+            .await
```
carlaKC, line 590: For consistency, let's log exit here like we do for the other spawned tasks. 

Also wasn't initially clear why we're not also triggering here. Could you add a comment that `run_results_logger` only exits on `listener` trigger so we don't need to handle erroring out?

sim-lib/src/lib.rs
```diff
@@ -674,12 +717,21 @@ impl Simulation {
...
+            let ce_node = node.clone();
+            tasks.spawn(async move {
+                log::debug!("Starting events consumer for {}.", ce_node.lock().await.get_info());
+                match consume_events(ce_node, receiver, ce_output_sender, ce_listener).await {
+                    Ok(_) => log::debug!("Event consumer exited due to a triggered shutdown or because all channel senders have dropped."),
+                    Err(e) => {
+                        ce_shutdown.trigger();
+                        log::error!("Event consumer exited with error: {e:?}.");
+                    },
+                }
```
carlaKC, line 733: Only learning best practises for these patterns myself recently, but I think you can use `if let Err(e) = consume ... else` here because we don't care about the actual `Ok` value.  

Also nitty: the `Ok` path log is a little long, I think we can just say `Event consumer exited.` since the only people looking at this level of logs is us, and we'll then grep for it in the codebase.

sim-lib/src/lib.rs
```diff
@@ -701,88 +753,108 @@ impl Simulation {
...
-                    destination: dest.pubkey,
-                    dispatch_time: SystemTime::now(),
-                };
+    listener: Listener,
+) -> Result<(), SimulationError> {
+    loop {
+        select! {
+            biased;
+            _ = listener.clone() => {
+                log::debug!("Simulation event consumer for node {} received shutdown signal. Exiting.", node.lock().await.get_info());
```
carlaKC, line 798: nit: can remove this log now that we have the top level one?

sim-lib/src/lib.rs
```diff
@@ -988,64 +1045,79 @@ async fn produce_simulation_results(
...
                         match simulation_output{
                             SimulationOutput::SendPaymentSuccess(payment) => {
-                                let source_node = nodes.get(&payment.source).unwrap().clone();
-                                set.spawn(track_payment_result(
-                                    source_node, results.clone(), payment, shutdown.clone(),
-                                ));
+                                if let Some(source_node) = nodes.get(&payment.source) {
+                                    set.spawn(track_payment_result(
+                                        source_node.clone(), results.clone(), payment, listener.clone()
+                                    ));
```
carlaKC, line 1067: Should we handle errors returned by `track_payment_result` and trigger shutdown here?

sim-lib/src/lib.rs
```diff
@@ -988,64 +1045,79 @@ async fn produce_simulation_results(
...
+                    None => return Ok(())
+                }
+            },
+            track_payment = set.join_next() => {
+                if let Some(res) = track_payment {
+                    match res {
+                        Ok(track_payment_res) => {
+                            if let Err(e) = track_payment_res {
+                                log::error!("Track payment result task exited");
+                                return Err(e);
```
carlaKC, line 1093: We log all of the errors returned by this function at the call site so we don't need to log each one as we return them. Applies in a few places.

sim-lib/src/lib.rs
```diff
@@ -133,6 +133,14 @@ pub enum SimulationError {
     FileError,
     #[error("{0}")]
     RandomActivityError(RandomActivityError),
+    #[error("System Time Error: {0}")]
+    SystemTimeError(#[from] SystemTimeError),
+    #[error("Missing Node Error: {0}")]
+    MissingNodeError(String),
+    #[error("Mpsc Channel Error: {msg}: {error_str}")]
+    MpscChannelError { msg: String, error_str: String },
```
carlaKC, line 141: I've seen this pattern around rust codebase quite a bit, could you explain the rationale for two fields here when we're just logging both values? Seems like we could just have `MpscChannelError(String)`, though totally open to me being completely ignorant about rust conventions here - shan't be the last time 🦀

carlaKC (commented):
Thanks for addressing comments, think this only needs one more round!

Remaining comments are really about logging consistency - if we're logging an exit error at a function's call site there's not need to also log on error return (we'll double log). Would also like to have all the starting/stopping logs moved into the `spawn` as well.

sim-lib/src/lib.rs
enigbe: Yes, we should. I refactored `produce_simulation_results` to include an additional [branch](https://github.com/bitcoin-dev-project/sim-ln/blob/b84ce8d24d5639139fa6bc8ca945557444abd2ff/sim-lib/src/lib.rs#L1087C13-L1102C14) to wait on concurrently. Within this branch, we propagate any `track_payment_result` error to `produce_simulation_results` and trigger shutdown at the latter's spawn site.

sim-lib/src/lib.rs
enigbe: Please ignore this. I have updated the variant to `MpscChannelError(String)`. 
I use this style when I want to add some context to an internal error with multiple variants, and control logic based on these variants. It has no use here (possible did this out of habit)

sim-lib/src/lib.rs
enigbe: Done the cleanup at all locations.

sim-lib/src/lib.rs
```diff
@@ -825,72 +904,51 @@ async fn produce_events<N: DestinationGenerator + ?Sized, A: PaymentGenerator +
...
     let mut counter = 1;
+
     loop {
-        tokio::select! {
+        select! {
             biased;
             _ = listener.clone() => {
-                log::debug!("Simulation results consumer received shutdown signal.");
-                return writer.map_or(Ok(()), |(ref mut w, _)| w.flush().map_err(|_| SimulationError::FileError))
+                return Ok(());
```
carlaKC, line 948: Still need to flush writer to disk otherwise pending items in the batch won't be written to disk?

sim-lib/src/lib.rs
carlaKC: Nice, realize that we weren't actually waiting on that set at all before this ☠️ 

This method is an interesting one (/different to our other ones) because it has its own set of tasks that it should wait on. As is, if we get the shutdown listener signal we won't wait for all the spawned payment tracking tasks to complete (which is messy). 

Don't need to update in this PR, let's gettit in, but note to self to create an issue/fix this up in future!

carlaKC (commented):
alrightalrightalright! 

Actually, actually last comments: 
1. We still need to flush our batched writer to disk on shutdown
2. Take a look at line wrapping at 120 (I think that a few places are over / feel free to shorten error messagse)

Otherwise, you can go ahead and squash the fixups and we'll merge. Nice stuff 🏅

sim-lib/src/lib.rs
enigbe: I've taken note of this and will create an issue for better handling of exits on all tasks spawned in `set`.

carlaKC:
Testing this and it looks like `TrackPayment` runs anymore and the simulator hangs on shutdown. 

Will take a look at the code, but iirc this is a regression since my last review (I tested last time and this was fine).

enigbe:
That's weird. Taking another look right now.

sim-lib/src/lib.rs
```diff
@@ -723,13 +728,18 @@ impl Simulation {
...
             tasks.spawn(async move {
-                log::debug!("Starting events consumer for {}.", ce_node.lock().await.get_info());
-                match consume_events(ce_node, receiver, ce_output_sender, ce_listener).await {
-                    Ok(_) => log::debug!("Event consumer exited due to a triggered shutdown or because all channel senders have dropped."),
-                    Err(e) => {
-                        ce_shutdown.trigger();
-                        log::error!("Event consumer exited with error: {e:?}.");
-                    },
+                let consume_event_node = ce_node.clone();
+                let node_guard = ce_node.lock().await;
```
carlaKC, line 694: This is deadlocking with the lock in `consume_events`!

carlaKC: lol I hate programming, can fix this with: 
```
let node_info = node_guard.get_info().clone();
drop(node_guard);
```

sim-lib/src/lib.rs
```diff
@@ -1105,9 +1172,19 @@ async fn track_payment_result(
...
+        },
+        send_payment_result = results.send((payment, res.clone())) => {
+            if send_payment_result.is_err() {
+                return Err(SimulationError::MpscChannelError(format!("Failed to send payment result {res} for payment {payment}.")))
+            }
+        }
     }
 
-    log::trace!("Payment result tracker exiting.");
+    log::trace!("Result tracking complete. Payment result tracker exiting.");
```
carlaKC, line 1196: nit: logging at call site, not in function as with others.

sim-lib/src/lib.rs
```diff
@@ -985,12 +1045,14 @@ impl Display for PaymentResultLogger {
     }
 }
 
+/// Reports a summary of payment results at a duration specified by `interval`
+/// Note that `run_results_logger` does not error in any way, thus it has no
+/// trigger. It listens for triggers to ensure clean exit.
```
carlaKC, line 1059: nit: line wrap at 120

sim-lib/src/lib.rs
```diff
@@ -859,72 +935,54 @@ async fn produce_events<N: DestinationGenerator + ?Sized, A: PaymentGenerator +
...
-        tokio::select! {
+        select! {
             biased;
             _ = listener.clone() => {
-                log::debug!("Simulation results consumer received shutdown signal.");
-                return writer.map_or(Ok(()), |(ref mut w, _)| w.flush().map_err(|_| SimulationError::FileError))
+                writer.map_or(Ok(()), |(ref mut w, _)| w.flush().map_err(|_| {
+                    SimulationError::FileError
+                }))?;
+                return Ok(());
```
carlaKC, line 991: Don't need `?` + `return Ok(())` -> can just return `writer.map_or...`

.gitignore
```diff
@@ -3,4 +3,4 @@ node_modules
 *.json
 activity-generator/releases/*
 .DS_Store
-/results
\ No newline at end of file
+/results
```
carlaKC, line 6: nit: newline still here - might be editor automatically adding it

sim-lib/src/lib.rs
```diff
@@ -735,88 +794,109 @@ impl Simulation {
...
+            // pe: produce events
+            let pe_shutdown = self.shutdown_trigger.clone();
+            let pe_listener = self.shutdown_listener.clone();
+            let pe_sender = sender.clone();
+            tasks.spawn(async move {
+                let source = executor.source_info.clone();
+
+                log::info!(
+                    "Starting activity producer for {}: {}.",
+                    source,
```
carlaKC, line 815: nit: can just inline + clone because we don't have any locks here

sim-lib/src/lib.rs
```diff
@@ -1022,64 +1082,71 @@ async fn produce_simulation_results(
...
                             }
                         };
-
                     },
-                    None => {
-                        return
+                    None => return Ok(())
+                }
+            },
+            track_payment = set.join_next() => {
```
carlaKC, line 1129: nit: add a TODO explaining that we're not going to wait for all tasks to exit, just so we don't forget

carlaKC (approved):
tACK at 5aa5f653779908ef1227a249f7f4f708abbf67cd.

Some minor comments that can be addressed in a followup if desired.

PR #166, opened 2024-02-01
feat: add simulation config

### LINKED ISSUE #157: Refactor: Create config struct for simulator
1. Add a `SimulationConfig` struct that passes all the configuration options to the simulator rather than individual params. 
2. Make `log_interval` configurable while we're here (currently hard set to 60s)

Started working on this in https://github.com/bitcoin-dev-project/sim-ln/commit/e929a2d6c28687df4b491594dc84c0ae9a1dc208 in #141 but decided to leave for a follow up.

enigbe:
## What this PR does
- ~~Adds a `SimulationConfig` struct that~~ Loads config values from an optional `INI` file and overwrites these values if corresponding command line arguments are passed 
- The config file contains custom values for the simulation
```ini
[simln.conf]
data_dir = "."
sim_file = "sim.json"
total_time=
print_batch_size = 500
log_level = "info"
expected_pmt_amt = 3800000
capacity_multiplier = 2
no_results = false
log_interval = 60
```
- Makes configurable the `log_interval` after which results are logged

## Related Issue(s)
- Resolves #157 

## Update
- This PR has changed to consider a standard way of handling configuration files as well as command line arguments. It currently introduces functionality to create and parse the `Cli` from an optionally-provided config file, to merge this `Cli` with that parsed from command line arguments, and to grant precedence to the options/values in the latter.

- Secondly, this PR also replaces the logging crate from `simple_logger` to `flexi_logger`. `flexi_logger` allows dynamic update of log levels after initialization and was required to deal with the issue described [here](https://github.com/bitcoin-dev-project/sim-ln/pull/166#discussion_r1483818333).

- Finally, documentation was updated to reflect the introduction and use of a config file.

Note: There are some changes to unrelated files because of the requirement to wrap maximum width of code and comments at 120 column spaces. Thought to include them regardless following conversation from #164 on [maintaining width](https://github.com/bitcoin-dev-project/sim-ln/pull/164#pullrequestreview-1912717117).


sim-cli/src/config.rs
```diff
@@ -0,0 +1,209 @@
...
+#[derive(Debug)]
+pub struct SimulationConfig {
+    // Path to a directory where simulation results are save
+    pub data_dir: PathBuf,
+    // Path to simulation file
+    pub sim_file: PathBuf,
+    // Duration for which the simulation should run
+    pub total_time: Option<u32>,
+    // Number of activity results to batch together before printing to csv file [min: 1]
+    pub print_batch_size: u32,
```
carlaKC, line 18: Pre-existing: could you standardize to use documentation comments here?
This was updated across the repo in #160, but won't be detected since we've moved these vaues.

sim-cli/src/config.rs
```diff
@@ -0,0 +1,209 @@
+use std::{path::PathBuf, str::FromStr};
+
+use anyhow::Context;
+use log::LevelFilter;
+use serde::Deserialize;
+
+use crate::Cli;
+
+#[derive(Debug)]
+pub struct SimulationConfig {
```
carlaKC, line 10: Can we derive serde's `Serialize` and `Deserialize` here and then provide a custom serialization for log level to save some code? Like we do for [lnd](https://github.com/bitcoin-dev-project/sim-ln/blob/main/sim-lib/src/lnd.rs#L30) paths.

sim-cli/src/config.rs
```diff
@@ -0,0 +1,209 @@
+use std::{path::PathBuf, str::FromStr};
+
+use anyhow::Context;
```
carlaKC, line 3: We've tended to use std result throughout the codebase so far.

sim-cli/src/config.rs
```diff
@@ -0,0 +1,209 @@
...
+    ) -> anyhow::Result<Self> {
+        // 1. Load simulation config from specified path
+        let mut sim_conf = serde_json::from_str::<SimulationConfig>(
+            &std::fs::read_to_string(&path)
+                .context(format!("Failed to read content of {:?} to string", path))?,
+        )
+        .context("Failed to deserialize configuration string into SimulationConfig")?;
+
+        // 2. Override config values with CLI arguments (if passed)
+        if let Some(data_directory) = data_dir {
```
carlaKC, line 96: This is quite a lot of code for similar functionality, can we do a macro to cut LOC?

conf.json
```diff
@@ -0,0 +1,11 @@
...
+  "data_dir": ".",
+  "sim_file": "sim.json",
+  "total_time": null,
+  "print_batch_size": 500,
+  "log_level": "info",
+  "expected_pmt_amt": 3800000,
+  "capacity_multiplier": 2,
+  "no_results": false,
+  "log_interval": 60
+}
```
carlaKC: Slight personal preference to use a INI file over json, _but_ we already have `json` in the project so not a strongly held opinion. If it's easy to switch over to serde (suggested in comments that follow), would like to consider it if other folks feel the same.

carlaKC (commented):
Thanks for picking this up! Just did a first pass, only major commets about serialization / DRY-in up some of the config checks with a macro.

sim-cli/src/config.rs
enigbe: I do agree too. I have written a macro for this. However the log statements are never printed to the terminal because we [initialize](https://github.com/enigbe/sim-ln/blob/237475f7b3a98dbd53e7630be95a59e200602ff8/sim-cli/src/main.rs#L69C5-L80C19) the logger after loading and substituting configs.

```rust
macro_rules! cli_overwrite {
    ($config:expr, $cli:expr, $($field:ident),*) => {
        match (&mut $config, $cli) {
            (config_struct, cli_struct) => {
                $(
                    if let Some(value) = cli_struct.$field {
                        log::info!(
                            "SimulatinConfig::{} {:?} overwritten by CLI argument {:?}",
                            stringify!($field),
                            config_struct.$field,
                            value
                        );
                        config_struct.$field = value;
                    }
                )*
            }
        }

    };
}
```

I would like to replace the `simple_logger` to something that allows runtime update to log levels but not without agreement from others.

conf.json
enigbe: Thought to use json because we already use it but happy to switch to INI.

sim-cli/src/config.rs
enigbe: Done

okjodom:
@enigbe here comes a late review!

I've read through the changes proposed and seem beyond the scope of #157 

> Add a `SimulationConfig` struct that passes all the configuration options to the simulator rather than individual params.

I was expecting we create this new struct in `sim-lib` and update `Simulation::new()` to take nodes, activity, and config as params. In `sim-cli` we can just add a type coercion which converts the cli params into `SimulationConfig` and use it after `Cli.parse()`. Example coercion below
```rs
impl Cli {
    fn to_simulation_config(&self) -> anyhow::Result<SimulationConfig> {
        Ok(SimulationConfig {
            total_time: self.total_time.unwrap_or(60),
            print_batch_size: self.print_batch_size,
            log_level: self.log_level,
            expected_pmt_amt: self.expected_pmt_amt,
            capacity_multiplier: self.capacity_multiplier,
            no_results: self.no_results,
        })
    }
}

async fn main() {
  let cli = Cli::parse();
  let opts = cli.to_simulation_config().unwrap();
  ...  
}
```
Any changes that require the `.ini` might make sense if there's a user scenario needing it. I'm not aware of any so deferring to @carlaKC 

> Make log_interval configurable while we're here (currently hard set to 60s)

cool beans on this part of the issue. change lgtm :)

carlaKC:
While this is outside of the scope of the original issue, I think it's a really nice extension to simln / pretty standard way to handle configuration. I can see it being useful when using simln in larger deployments where you want to have a config file checked in rather than deal with cli flags, so my vote is to keep the proposed functionality 👍

sim-cli/src/config.rs
carlaKC: Ah that's annoying, happy for you to shop around for a different logger 👍 

Thought we might be doing some more complex log-fu while we're running (which I didn't like the sound of), but if we're just going to set the level once off after we've parsed config works for me.

sim-cli/src/config.rs
```diff
@@ -0,0 +1,241 @@
...
+                    }
+                )*
+            }
+        }
+
+    };
+}
+
+#[derive(Debug, Deserialize)]
+pub struct SimulationConfig {
```
carlaKC, line 32: Wondering whether we can move deserialization onto the existing cli struct and then just get CLI parsing to only override values that are present? 

Then we can create a simple `SimulationConfig` struct in `lib` can just convert the cli -> config struct like @okjodom suggested.

sim-cli/src/main.rs
```diff
@@ -14,21 +17,6 @@ use sim_lib::{
...
-pub const DEFAULT_SIM_FILE: &str = "sim.json";
-
-/// The default expected payment amount for the simulation, around ~$10 at the time of writing.
-pub const EXPECTED_PAYMENT_AMOUNT: u64 = 3_800_000;
-
-/// The number of times over each node in the network sends its total deployed capacity in a calendar month.
-pub const ACTIVITY_MULTIPLIER: f64 = 2.0;
-
-/// Default batch size to flush result data to disk
-const DEFAULT_PRINT_BATCH_SIZE: u32 = 500;
```
carlaKC, line 30: Not sure about moving these defaults from in the codebase to into a config file. That moves us from being able to run `sim-ln` easily on its own to always needing a config file (or lots of flags) which makes things less usable (thinking about server contexts like docker where we'd now need to copy this config file into the container).

carlaKC (commented):
Some nits + one major comment about having separate structs for CLI and config=] parsing and how we manage defaults. If possible, would like to squash those into a single struct (I _think_ clap has some settings that would make that possible).

sim-cli/src/config.rs
```diff
@@ -0,0 +1,241 @@
...
+
+/// Overwrites the fields on a SimulationConfig if corresponding arguments are supplied via the command line
+macro_rules! cli_overwrite {
+    ($config:expr, $cli:expr, $($field:ident),*) => {
+        match (&mut $config, $cli) {
+            (config_struct, cli_struct) => {
+                $(
+                    if let Some(value) = cli_struct.$field {
+                        log::info!(
+                            "SimulatinConfig::{} {:?} overwritten by CLI argument {:?}",
```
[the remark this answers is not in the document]
enigbe: This has been removed following changes to how the config values are loaded. Marking as resolved as its no longer relevant.

sim-cli/src/config.rs
```diff
@@ -0,0 +1,241 @@
...
+        } else {
+            anyhow::bail!("Failed to convert path {path:?} to &str.")
+        };
+
+        let config = Config::builder()
+            .add_source(File::with_name(path_str))
+            .build()?;
+        let mut sim_conf: SimulationConfig = config.try_deserialize()?;
+
+        // 2. Overwrite config values with CLI arguments (if passed)
```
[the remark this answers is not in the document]
enigbe: Same as [this](https://github.com/bitcoin-dev-project/sim-ln/pull/166#discussion_r1519706358).

sim-lib/src/lib.rs
```diff
@@ -339,6 +339,8 @@ pub struct Simulation {
     activity_multiplier: f64,
     /// Configurations for printing results to CSV. Results are not written if this option is None.
     write_results: Option<WriteResults>,
+    /// Duration after which results are logged
```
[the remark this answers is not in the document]
enigbe: Done.

sim-cli/src/main.rs
enigbe: I noted this in the changes I have introduced. The defaults remain, and allow us to handle the absence of a config file.

carlaKC:
Needs rebase!

carlaKC:
Closing due to inactivity, please feel free to re-open if you've got capacity to pick this back up with the [suggested split](https://github.com/bitcoin-dev-project/sim-ln/pull/166#issuecomment-2113242974).

PR #173, opened 2024-03-12
Issue #168: Adding optional start and count values to sim.json file

### LINKED ISSUE #168: Add start/duration options to the defined activity configuration
**Describe desired feature**

Currently when you are defining specific activity you can only define an interval and amount. It would be helpful to be able to pick a start time and duration so that some nodes are sending transactions at different times throughout the simulation.

**Use case for feature**

For example:
This activity definition would tell Alice to send 2000 msats every second, starting 10 seconds after the simulation starts and ending after 30 seconds. In other words, Alice would send Bob 30 transactions between 10 and 40 seconds in the simulation.
```
  "activity": [
    {
      "source": "Alice",
      "destination": "Bob",
      "start": 10,
      "duration": 30
      "interval_secs": 1,
      "amount_msat": 2000
    }
  ]
```

**Would you like to contribute code for this feature?**

I am interested and hope to contribute toward this feature.

bjohnson5:
Adding `start` and `count` options to the sim.json file.

`start`: Specifies when in the simulation the user wants this activity to start (for example... start sending payments at 10s)
`count`: Specifies the number of payments to send throughout the simulation (for example... send 15 payments and then stop)

Resolves #168

### DIFF
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -139,6 +139,12 @@ pub struct ActivityParser {
     /// The destination of the payment.
     #[serde(with = "serializers::serde_node_id")]
     pub destination: NodeId,
+    /// The time in the simulation to start the payment
+    #[serde(default)]
+    pub start_secs: u16,
+    /// The number of payments to send over the course of the simulation
+    #[serde(default)]
+    pub count: Option<u64>,
     /// The interval of the event, as in every how many seconds the payment is performed.
     pub interval_secs: u16,
     /// The amount of m_sat to used in this payment.
@@ -153,6 +159,10 @@ pub struct ActivityDefinition {
     pub source: NodeInfo,
     /// The destination of the payment.
     pub destination: NodeInfo,
+    /// The time in the simulation to start the payment
+    pub start_secs: u16,
+    /// The number of payments to send over the course of the simulation
+    pub count: Option<u64>,
     /// The interval of the event, as in every how many seconds the payment is performed.
     pub interval_secs: u16,
     /// The amount of m_sat to used in this payment.
@@ -261,6 +271,12 @@ pub trait DestinationGenerator: Send {
 pub struct PaymentGenerationError(String);
 
 pub trait PaymentGenerator: Display + Send {
+    /// Returns the time that the payments should start
+    fn payment_start(&self) -> Duration;
+
+    /// Returns the number of payments that should be made
+    fn payment_count(&self) -> Option<u64>;
+
     /// Returns the number of seconds that a node should wait until firing its next payment.
     fn next_payment_wait(&self) -> time::Duration;
 
@@ -667,6 +683,8 @@ impl Simulation {
             for description in self.activity.iter() {
                 let activity_generator = DefinedPaymentActivity::new(
                     description.destination.clone(),
+                    Duration::from_secs(description.start_secs.into()),
+                    description.count,
                     Duration::from_secs(description.interval_secs.into()),
                     description.amount_msat,
                 );
@@ -918,8 +936,21 @@ async fn produce_events<N: DestinationGenerator + ?Sized, A: PaymentGenerator + 
     sender: Sender<SimulationEvent>,
     listener: Listener,
 ) -> Result<(), SimulationError> {
+    let mut current_count = 0;
     loop {
-        let wait = node_generator.next_payment_wait();
+        if let Some(c) = node_generator.payment_count() {
+            if c == current_count {
+                log::info!("Payment count has been met: {c} payments. Stopping the activity.");
+                return Ok(());
+            }
+        }
+
+        let wait: Duration = if current_count == 0 {
+            node_generator.payment_start()
+        } else {
+            node_generator.next_payment_wait()
+        };
+
         log::debug!("Next payment for {source} in {:?} seconds.", wait);
 
         select! {
@@ -956,6 +987,7 @@ async fn produce_events<N: DestinationGenerator + ?Sized, A: PaymentGenerator + 
                     return Err(SimulationError::MpscChannelError (format!("Stopped random producer for {amount}: {source} -> {destination}.")));
                 }
 
+                current_count += 1;
             },
         }
     }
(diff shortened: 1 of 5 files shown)


carlaKC (commented):
cACK, going to need a rebase now that #164 is in!

carlaKC:
Missed that this is ready for a look - will give it a pass soon!

sim-lib/src/lib.rs
```diff
@@ -139,6 +139,12 @@ pub struct ActivityParser {
     /// The destination of the payment.
     #[serde(with = "serializers::serde_node_id")]
     pub destination: NodeId,
+    /// The time in the simulation to start the payment
+    #[serde(default)]
+    pub start_secs: u16,
```
enigbe, line 144: `#[serde(rename(deserialize = "start"))]` to deserialize from activity field name `start`

enigbe (commented):
@bjohnson5 Thanks for contributing this. I have reviewed and done some local testing. Here are my comments:

- I think the documentation should be modified to mention the addition of the optional fields `start` (or `start_secs`) and `count` activity fields.

- Also, I noticed that the `ActivityParser` defaults to `u64`'s default (`0`) if the `start` field isn't `start_secs`. Whichever field name you pick is fine but sticking to the format [here](https://github.com/bitcoin-dev-project/sim-ln/issues/168#issue-2162099069) would require an extra serde field attribute to rename when deserializing.
```json
  "activity": [
    {
      "source": "Alice",
      "destination": "Bob",
      "start": 10,  // or start_sec
      "count": 30
      "interval_secs": 1,
      "amount_msat": 2000
    }
  ]
```

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct ActivityParser {
    ...
    /// The time in the simulation to start the payment
    #[serde(default, rename(deserialize = "start"))]
    pub start_secs: u16,
    ...
}

```

sim-lib/src/lib.rs
bjohnson5: @enigbe I decided to switch from `start` to `start_secs` to be consistent with `interval_secs` and to make sure the user knows what units to use.

bjohnson5:
@enigbe Thanks for the review! I added some details to the README about the new `start_secs` and `count` fields. Let me know if there is anywhere else you think this should be documented.

I think leaving the field name as `start_secs` makes sense for consistency and clarity. Open to your thoughts on that though, let me know if you have a better name in mind.

enigbe (approved):
Thanks for addressing my comments @bjohnson5. No additional review from me. I'm happy to approve this.

sim-lib/src/lib.rs
```diff
@@ -918,8 +936,23 @@ async fn produce_events<N: DestinationGenerator + ?Sized, A: PaymentGenerator +
...
+            if c == current_count {
+                log::info!(
+                    "Payment count has been met for {source}: {c} payments. Stopping the activity."
+                );
+                return Ok(());
+            }
+        }
+
+        let wait: Duration = if current_count == 0 {
+            node_generator.payment_start()
```
carlaKC, line 951: let's add a `debug` log in here that we're using the `start` delay for the payment, could be useful.

carlaKC (commented):
Code looks good! 

One high level thing that I think we want to think about before going ahead with merge is how we want to handle the case where all of our activities have finished. Right now, the simulator will just keep running without _really_ telling the end user that there's nothing left to be done. 

Off the top of my head, we'd want to find a way to trigger `shutdown` when all the tasks created by `dispatch_producers` are done - perhaps giving them their own joinset + spawning a task to kill the simulation if they all complete (though I haven't full thought through that appraoch) - @enigbe any thoughts on this given you've just been thinking about all this?

carlaKC:
Also happy to go ahead with merge and add this shutdown handling in a followup @bjohnson5 if that's going to make your life easier, since you're the primary consumer of this feature rn :)

bjohnson5:
@carlaKC I think that your suggestion for shutting down the simulation if all producers finish makes sense. I added a separate `JoinSet` for the producers and a new task that will trigger the shutdown if all the producer threads complete. @enigbe I would love to get you to review that new change if you have time, since you just worked on the shutdown process.

I also addressed the other review comments. Thanks!

enigbe (approved):

sim-lib/src/lib.rs
```diff
@@ -554,9 +570,25 @@ impl Simulation {
         );
 
         // Next, we'll spin up our actual producers that will be responsible for triggering the configured activity.
-        self.dispatch_producers(activities, consumer_channels, &mut tasks)
+        // The producers will use their own JoinSet so that the simulation can be shutdown if they all finish.
+        let mut producer_tasks = JoinSet::new();
```
carlaKC, line 574: Nice simple approach ✨ 

Very nifty that we don't need to wait on `producer_tasks` at the end of the simulation because they're represented by our high level `tasks` set! 

@enigbe possibly something that we could look at for `track_payment`?

sim-lib/src/lib.rs
```diff
@@ -918,8 +952,27 @@ async fn produce_events<N: DestinationGenerator + ?Sized, A: PaymentGenerator +
...
+                log::info!(
+                    "Payment count has been met for {source}: {c} payments. Stopping the activity."
+                );
+                return Ok(());
+            }
+        }
+
+        let wait: Duration = if current_count == 0 {
+            log::debug!(
+                "Using a start delay. The first payment for {source} will be at {:?} seconds.",
```
carlaKC, line 968: nit: this will log even if no delay is set:
` Using a start delay. The first payment for alice(020f3b...b44c96) will be at 0ns seconds.`

move `node_generator.payment_start()` up and log/use if !=0?

carlaKC (approved):
tACK both cases where:
* All producers exit -> we shutdown
* Some producers exit -> we stay up / shutdown cleanly when the sim is killed.

sim-lib/src/lib.rs
bjohnson5: Ah good catch. I just added a check to only log that if the start time != 0. Also moved the `Next payment for {source}...` message into the `else` block because it was kind of redundant at the start of the simulation. Also removed the word `seconds` from both of those log messages because the `Duration` class prints the units automatically, so it was also redundant.

carlaKC:
LGTM - will merge when CI's done ✅

PR #175, opened 2024-03-14
sim_network: Add unit tests + squash bugs

carlaKC:
This PR adds coverage for our simulated network and (unsurprisingly) squashes some bugs: 
* Do not double-return liquidity for failed HTLCs: previously we'd return liquidity to local balance on failures twice over
* Off by one removal: if a HTLC failed at index 0, we would not remove the first HTLC because our loop was not running for `=0`.

Removal bugs not caught in original testing because we evenly distribute liquidity, so didn't run into any payment failure - going to set liquidity to zero and re-run for a while to see if anything falls apart.

### DIFF
--- a/sim-lib/src/sim_node.rs
+++ b/sim-lib/src/sim_node.rs
@@ -894,7 +894,7 @@ async fn remove_htlcs(
     payment_hash: PaymentHash,
     success: bool,
 ) -> Result<(), ForwardingError> {
-    for (i, hop) in route.hops[0..resolution_idx].iter().enumerate().rev() {
+    for (i, hop) in route.hops[0..=resolution_idx].iter().enumerate().rev() {
         // When we add HTLCs, we do so on the state of the node that sent the htlc along the channel so we need to
         // look up our incoming node so that we can remove it when we go backwards. For the first htlc, this is just
         // the sending node, otherwise it's the hop before.
@@ -967,8 +967,15 @@ async fn propagate_payment(
         }
     } else {
         // If we successfully added the htlc, go ahead and remove all the htlcs in the route with successful resolution.
-        if let Err(e) =
-            remove_htlcs(nodes, route.hops.len(), source, route, payment_hash, true).await
+        if let Err(e) = remove_htlcs(
+            nodes,
+            route.hops.len() - 1,
+            source,
+            route,
+            payment_hash,
+            true,
+        )
+        .await
         {
             if e.is_critical() {
                 shutdown.trigger();
--- a/sim-lib/src/sim_node.rs
+++ b/sim-lib/src/sim_node.rs
@@ -225,26 +225,11 @@ impl ChannelState {
         Ok(())
     }
 
-    /// Removes the HTLC from our set of outgoing in-flight HTLCs, failing if the payment hash is not found. If the
-    /// HTLC failed, the balance is returned to our local liquidity. Note that this function is not responsible for
-    /// reflecting that the balance has moved to the other side of the channel in the success-case, calling code is
-    /// responsible for that.
-    fn remove_outgoing_htlc(
-        &mut self,
-        hash: &PaymentHash,
-        success: bool,
-    ) -> Result<Htlc, ForwardingError> {
-        match self.in_flight.remove(hash) {
-            Some(v) => {
-                // If the HTLC failed, pending balance returns to local balance.
-                if !success {
-                    self.local_balance_msat += v.amount_msat;
-                }
-
-                Ok(v)
-            },
-            None => Err(ForwardingError::PaymentHashNotFound(*hash)),
-        }
+    /// Removes the HTLC from our set of outgoing in-flight HTLCs, failing if the payment hash is not found.
+    fn remove_outgoing_htlc(&mut self, hash: &PaymentHash) -> Result<Htlc, ForwardingError> {
+        self.in_flight
+            .remove(hash)
+            .ok_or(ForwardingError::PaymentHashNotFound(*hash))
     }
 }
 
@@ -349,7 +334,7 @@ impl SimulatedChannel {
     ) -> Result<(), ForwardingError> {
         let htlc = self
             .get_node_mut(sending_node)?
-            .remove_outgoing_htlc(hash, success)?;
+            .remove_outgoing_htlc(hash)?;
         self.settle_htlc(sending_node, htlc.amount_msat, success)?;
         self.sanity_check()
     }
--- a/sim-lib/src/sim_node.rs
+++ b/sim-lib/src/sim_node.rs
@@ -231,6 +231,24 @@ impl ChannelState {
             .remove(hash)
             .ok_or(ForwardingError::PaymentHashNotFound(*hash))
     }
+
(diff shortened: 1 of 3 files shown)


sim-lib/src/sim_node.rs
```diff
@@ -1035,3 +1038,742 @@ impl UtxoLookup for UtxoValidator {
...
+    /// Contains elements required to test dispatch_payment functionality.
+    struct DispatchPaymentTestKit<'a> {
+        graph: SimGraph,
+        nodes: Vec<PublicKey>,
+        routing_graph: NetworkGraph<&'a WrappedLog>,
+        shutdown: triggered::Trigger,
+    }
+
+    impl<'a> DispatchPaymentTestKit<'a> {
+        /// Creates a test graph with a set of nodes connected by channels three channels, with all the capacity of
```
enigbe, line 1525: nit: "connected by three channels"

sim-lib/src/sim_node.rs
```diff
@@ -1035,3 +1038,742 @@ impl UtxoLookup for UtxoValidator {
...
+    }
+
+    /// Contains elements required to test dispatch_payment functionality.
+    struct DispatchPaymentTestKit<'a> {
+        graph: SimGraph,
+        nodes: Vec<PublicKey>,
+        routing_graph: NetworkGraph<&'a WrappedLog>,
+        shutdown: triggered::Trigger,
+    }
+
```
enigbe, line 1523: I'm curious about the purpose of the `nodes` field in the test kit. Since public keys for the nodes are accessible from both `SimGraph` and `NetworkGraph`, is there a specific reason this additional field is needed?

sim-lib/src/sim_node.rs
```diff
@@ -1035,3 +1038,742 @@ impl UtxoLookup for UtxoValidator {
...
+        assert!(channel_state.remove_outgoing_htlc(&hash_1).is_ok());
+        channel_state.settle_outgoing_htlc(htlc_1.amount_msat, true);
+
+        assert!(channel_state.remove_outgoing_htlc(&hash_2).is_ok());
+        channel_state.settle_outgoing_htlc(htlc_2.amount_msat, true);
+
+        // Now we're going to add many htlcs so that we hit our in-flight count limit (unique payment hash per htlc).
+        for i in 0..channel_state.policy.max_htlc_count {
+            let hash = PaymentHash {
+                0: [i.try_into().unwrap(); 32],
```
enigbe, line 1308: This test might become flaky due to the conversion of a `u64` value to a `u8`. If `max_htlc_count` exceeds 255, this could lead to a panic.

sim-lib/src/sim_node.rs
```diff
@@ -1035,3 +1038,742 @@ impl UtxoLookup for UtxoValidator {
...
+            channel_state.check_htlc_forward(channel_state.policy.cltv_expiry_delta - 1, 0, 0),
+            Err(ForwardingError::InsufficientCltvDelta(_, _))
+        ));
+
+        // Test insufficient fee.
+        let htlc_amount = 1000;
+        let htlc_fee = (channel_state.policy.base_fee as f64
+            + ((channel_state.policy.fee_rate_prop as f64 * htlc_amount as f64) / 1000000.0))
+            as u64;
+
```
enigbe, line 1219: Are there any concerns about the precision of `f64` (especially for the `fee_rate`) or is it too early to have such worries? On my device for example, a 4.7 `f64` value is sometimes represented as 4.700000002. If you're open to exploring other options, [rust_decimal](https://crates.io/crates/rust_decimal) comes highly recommended (and used) in the community and I'd be happy to work on replacing the built-in decimal type with something more precise for financial calculations

sim-lib/src/sim_node.rs
```diff
@@ -1035,3 +1038,742 @@ impl UtxoLookup for UtxoValidator {
...
+        // this test.
+        let hash_1 = PaymentHash { 0: [1; 32] };
+        let htlc_1 = Htlc {
+            amount_msat: 1000,
+            cltv_expiry: 40,
+        };
+
+        assert!(channel_state.check_outgoing_addition(&htlc_1).is_ok());
+        assert!(channel_state.add_outgoing_htlc(hash_1, htlc_1).is_ok());
+        assert_channel_balances!(
```
enigbe, line 1150: Why the call to `check_outgoing_addition(...)` if the method is called in `add_outgoing_htlc(...)`? Wouldn't the result of the assertion be the same if a single (the second) call is made?

sim-lib/src/sim_node.rs
```diff
@@ -1035,3 +1038,742 @@ impl UtxoLookup for UtxoValidator {
...
+        // CLTV delta insufficient (one less than required).
+        assert!(matches!(
+            channel_state.check_htlc_forward(channel_state.policy.cltv_expiry_delta - 1, 0, 0),
+            Err(ForwardingError::InsufficientCltvDelta(_, _))
+        ));
+
+        // Test insufficient fee.
+        let htlc_amount = 1000;
+        let htlc_fee = (channel_state.policy.base_fee as f64
+            + ((channel_state.policy.fee_rate_prop as f64 * htlc_amount as f64) / 1000000.0))
```
enigbe, line 1217: Would extracting this into a constant improve readability? I'd wondered why the division with `1,000,000` until I consulted the lightning textbook to review how the fee rate is calculated.

enigbe (commented):
@carlaKC I think you've done a really good job with the tests. I have found them very helpful in navigating `sim_node` and better understanding aspects of the I'd cursorily reviewed before.
I have a few comments/questions on flakiness, decimal precision, multiple assertions, and readability.

sim-lib/src/sim_node.rs
carlaKC: Indeed, don't need both calls!

sim-lib/src/sim_node.rs
```diff
@@ -1035,3 +1038,742 @@ impl UtxoLookup for UtxoValidator {
...
+    }
+
+    /// Test addition of outgoing htlc to local state.
+    #[test]
+    fn test_check_outgoing_addition() {
+        // Create test channel with low local liquidity so that we run into failures.
+        let local_balance = 100_000;
+        let mut channel_state =
+            ChannelState::new(create_test_policy(local_balance / 2), local_balance);
+
```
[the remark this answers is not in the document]
carlaKC: Thought about this, but it's such a thin function that I didn't think it was worth it.

sim-lib/src/sim_node.rs
carlaKC: > become flaky 

It'll just outright fail rather than flake, so I think we're okay ot leave as-is for now. If we add tests that hit that count, then we can update this? Would rather just keep it simple to start (don't see a need to test > 255 IMO)

carlaKC:
Thanks for the review @enigbe! Updated to address most feedback, left some of the test setup as-is for the sake of simplicity (can always come back and improve as is required).

enigbe (approved):
Thanks for addressing all comments. I have no additional review on this.

PR #177, opened 2024-03-28
Adds clippy to gh actions and fixes clippy issues

sr-gi:
We are currently only doing cargo check to check that our code compiles. That is good, but rustlang has a pretty good tool called clippy, that helps find non-idiomatic constructions and calls that may be error-prone.

Replaces check by clippy + fmt and fixes some existing clippy issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,8 +34,8 @@ stop-docker:
 	docker stop sim-ln
 
 check-code:
-	cargo fmt --check --all
-	cargo clippy --all-features
+	cargo fmt --verbose --check --all -- --color always
+	cargo clippy --all-features --all-targets --color always -- -D warnings
 
 stable-output:
 	@if [ -n "$$(git status --porcelain)" ]; then \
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -6,16 +6,22 @@ permissions:
...
+  lint:
     name: Check Code with stable output
     runs-on: ubuntu-latest
     steps:
       - name: Install protoc
         run: sudo apt install -y protobuf-compiler
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
-      - name: run check
-        run: make check
```
carlaKC, line 18: We do have a clippy run in [check](https://github.com/bitcoin-dev-project/sim-ln/blob/main/Makefile#L48), perhaps we just need to add `--all-targets --color always -- --deny warnings`? 

Noticed that all the fixes are in tests, so that might be what we're missing?

carlaKC (commented):
cACK, live and die by the word of clippy 🫡

Perhaps update the existing makefile instead?

.github/workflows/build-and-test.yml
sr-gi: Should be fixed in [877ac51](https://github.com/bitcoin-dev-project/sim-ln/pull/177/commits/877ac51d571275c79978f6d7b5412f53b18f7dde)

carlaKC (approved):
Thx for catchin + cleanin up!

Issue #171, opened 2024-03-12
Feature: Run random activity generation with fixed seed


carlaKC:
Add the ability to run simln with `--fixseed` so that random activity generation starts with the same seed each time. While this won't be perfectly deterministic (because the order in which things execute on the lightning nodes will be different / change per-run if conditions are different), it will at least dispatch the activity in the same order across runs.

carlaKC:
This is mostly addressed in #178, though some nit-followups are [outstanding](https://github.com/bitcoin-dev-project/sim-ln/pull/178#pullrequestreview-2163171973).

We can probably also think about making some improvements, as we've not managed to produce perfectly deterministic runs [as described here](https://github.com/bitcoin-dev-project/sim-ln/pull/178#discussion_r1658520552). Closing as finished for now, we can always open a new issue for improvements.

PR #178, opened 2024-04-05
feat: run simulator with fixed seed

enigbe:
## What this PR does
Adds functionality to run the simulator with an optional fixed seed. If present, the seed allows the random activity generator to run deterministically.

Note the following:
- The RNG used, i.e. `ChaCha8Rng`, expects a seed type of `u64`

## Related Issue
- Addresses the issue outlined in #171


carlaKC:
Needs rebase!

sim-lib/src/random_activity.rs
```diff
@@ -206,8 +215,16 @@ impl PaymentGenerator for RandomPaymentActivity {
...
+        let mut rng: Box<dyn SimRng> = if let Some(seed_u64) = self.seed {
+            let chacha_8_rng = rand_chacha::ChaCha8Rng::from_seed(seed_u64);
+            Box::new(chacha_8_rng)
+        } else {
+            let thread_rng = rand::thread_rng();
+            Box::new(thread_rng)
+        };
+        // let mut rng = rand::thread_rng();
+        let duration = Duration::from_secs(self.event_dist.sample(&mut rng) as u64);
+        duration
```
carlaKC: remove commented out line + inline return of `duration`

sim-lib/src/random_activity.rs
```diff
@@ -348,26 +382,50 @@ mod tests {
...
             // here. Mainly, if the `capacity < expected_payment_amnt / 2`, the generator will fail building
             let expected_payment = get_random_int(1, 100);
-            assert!(
-                RandomPaymentActivity::new(2 * expected_payment, expected_payment, 1.0).is_ok()
-            );
+            assert!(RandomPaymentActivity::new(
+                2 * expected_payment,
+                expected_payment,
+                1.0,
+                Some([42; 32])
```
carlaKC, line 389: Any specific reasoning for using `None` above and `Some` here for seeds? non-blocking

sim-lib/src/random_activity.rs
```diff
@@ -266,8 +286,22 @@ impl Display for RandomPaymentActivity {
     }
 }
 
+/// Random number generator that takes an optional `u64` seed with which deterministic
+/// behaviour can be simulated
+fn sim_rng(seed: Option<[u8; 32]>) -> impl SimRng {
```
carlaKC, line 291: Can this be `Box<dyn RngCore>` to save us having to declare the supertrait?

sim-cli/src/main.rs
```diff
@@ -74,6 +90,10 @@ struct Cli {
     /// Do not create an output file containing the simulations results
     #[clap(long, default_value_t = false)]
     no_results: bool,
+    /// Seed to run random activity generator "deterministically", i.e. in the same order.
+    /// Expected values are hexadecimal-encoded [u8;32] seeds, e.g. hex::encode([42;32])
+    #[clap(long, short, value_parser = clap::builder::StringValueParser::new().try_map(deserialize_fix_seed))]
+    fix_seed: Option<[u8; 32]>,
```
carlaKC, line 96: usability question: do we actually need this much entropy for a simulator, or do you think we could get away with a u64 here? I ask because it would make deserializing easier and (IMO) is a little simpler for the end user.

carlaKC (commented):
Just reacquainting myself with this PR!

sim-cli/src/main.rs
carlaKC: @bjohnson5 any opinions here or are you mostly running with defined activity rn?

bjohnson5: I have pretty much only used defined activity. Hard to know how much entropy is actually helpful without testing a lot of different values and looking at the activity. I guess the reason for this type is due to the `rand_chacha.from_seed()` function. So at some point you have to turn it into an array of `u8` anyway?

sim-lib/src/random_activity.rs
enigbe: The seed parameter is optional so I thought one test could run with it provided, and another without it. If consistency matters to you I can write a second string of tests for both cases.

sim-cli/src/main.rs
enigbe: > usability question: do we actually need this much entropy for a simulator, or do you think we could get away with a u64 here? I ask because it would make deserializing easier and (IMO) is a little simpler for the end user.

I understand your concern here. This was the first approach I took, i.e. using a `u64`. However, `u64` is not considered [suitable for cryptography](https://rust-random.github.io/rand/rand/trait.SeedableRng.html#method.seed_from_u64). In light of the attackaton you were preparing for, I thought it best to not use something "exploitable" and to use a [recommended seed size](https://rust-random.github.io/rand/rand/trait.SeedableRng.html#associatedtype.Seed), i.e. `[u8; 32]`. Also, not sure about how this would be exploited but thought to cover all bases.

`ChaCha8Rng` provides a convenient `seed_from_u64` API I have used and can switch to if ease is better preferred.

sim-lib/src/random_activity.rs
carlaKC: > The seed parameter is optional so I thought one test could run with it provided, and another without it. If consistency matters to you I can write a second string of tests for both cases.

Was just wondering if there was any reason for the choice for those specific tests using the specific params, but makes sense!

> What are you referring to with "non-blocking"?

"non-blocking" for merge - so I have a question here but it's not something that _must_ be addressed/answered for the PR to move forward, just asking for my own understanding 👍

sim-cli/src/main.rs
carlaKC: For my use case, I'd just want to be able to experiment with fixing behavior for a few different values (out of scope for people to attack the random activity in the attackathon, it just passively runs). 

I think let's go ahead with u64, keeps it simple for now and we can expand if there's a end user need.

bjohnson5 (approved):

sim-lib/src/random_activity.rs
```diff
@@ -266,6 +282,18 @@ impl Display for RandomPaymentActivity {
     }
 }
 
+/// Random number generator that takes an optional seed with which deterministic
+/// behaviour can be simulated.
+fn sim_rng(seed: Option<u64>) -> Box<dyn RngCore> {
```
carlaKC, line 287: We can "just in time" create a `thread_rng` because it gets fresh entropy on each instantiation, but if we do this with a set seed, we'll always get the same value because we're creating the same fixed seed generator each time. This means that we'll always make the same choices every time we sample, which isn't what we want with a fixed seed. 

When I first implemented this, I tried to keep a single instance of `thread_rng` in the struct and it got really hairy because it's not `Send`. I think that a reasonable approach would be to replace `seed` in the various structs with `seeded_rng: Option<Box<dyn RngCore + Send>>` which is `Some` if we have a set seed and `None` otherwise. Should be relatively painless, and then we can use `unwrap_or` to use the seeded rng or just `thread_rng` inline where we need it?

See [test demonstrating issue](https://github.com/bitcoin-dev-project/sim-ln/files/15379642/test_rng.patch), sorry that I didn't pick this up on first pass!

sim-lib/Cargo.toml
```diff
@@ -30,6 +30,7 @@ csv = "1.2.2"
 serde_millis = "0.1.1"
 rand_distr = "0.4.3"
 mockall = "0.12.1"
+rand_chacha = "0.3.1"
 
 [dev-dependencies]
-ntest = "0.9.0"
\ No newline at end of file
+ntest = "0.9.0"
```
carlaKC, line 36: nit: unnecessary newline

carlaKC:
@enigbe is there any chance you'll be able to get to this in the next few days? I'd really like to get it merged because I've got some projects depending on it (and would like to tag a patch so that I can use our built in docker workflow). 

If not, I'd be interested in taking over the PR to move it along. Please let me know!

enigbe:
@carlaKC I am sorry for the delay but I can finish this on Wednesday if it's not too late.

sim-lib/src/random_activity.rs
```diff
@@ -94,12 +125,14 @@ impl Display for NetworkGraphView {
...
 /// randomness both in the time between events and the variance of payment amounts sent to mimic more realistic
-/// payment flows.
+/// payment flows. For deterministic time between events and payment amounts, the `RandomPaymentActivity` keeps a
+/// handle on a seed.
 pub struct RandomPaymentActivity {
     multiplier: f64,
     expected_payment_amt: u64,
     source_capacity: u64,
     event_dist: Exp<f64>,
+    seeded_rng: SeededRng,
```
carlaKC, line 135: Not ideal in terms of abstraction, but I think we're going to need to create a _single_ RNG from the fixed seed, and provide it to every instance of `RandomPaymentActivity` that we have. Because as is, we create _multiple_ rngs with the _same_ seed, so each node will execute payments with the same series of rng values (if you have nodes with the same capacity, they all end up sending the same amount for their payments). 

[Tried this out here](https://github.com/user-attachments/files/16014458/single_rng.patch).

carlaKC (commented):
Thanks for the update, heading in the right direction!

Should have thought of this on the last review, but I think that we actually want a _single_ fixed RNG, otherwise we'll have several rngs all producing the same set of values which leads to weird and not-desirable behavior (eg, all payment amounts being the same).

sim-lib/src/random_activity.rs
enigbe: I have made some changes so that all instances of `RandomPaymentActivity` share a single RNG but the deterministic outcome we want is not consistent. This could be down to how I have set my nodes but there is some `undesired` variance in the payment amount generated after multiple simulator runs.

Here is a small [table ](https://docs.google.com/spreadsheets/d/1MvN2tRgiOndapyyLi-gVFfhyhNVzC6TUKoETSsXavIk/edit?usp=sharing)showing the simulation run count and the payment amount generated at each run. The table is being updated every time I run the simulator to identify if the variance is as a result of my node setup or the code.

Please try this on your PC to see if you get similar outcomes.

sim-lib/src/lib.rs
```diff
@@ -1051,7 +1071,9 @@ async fn produce_events<N: DestinationGenerator + ?Sized, A: PaymentGenerator +
                 "Next payment for {source} in {:?}.",
                 node_generator.next_payment_wait()
```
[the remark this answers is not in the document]
enigbe: Resolved.

sim-lib/src/random_activity.rs
```diff
@@ -205,9 +245,18 @@ impl PaymentGenerator for RandomPaymentActivity {
     }
 
     /// Returns the amount of time until the next payment should be scheduled for the node.
-    fn next_payment_wait(&self) -> Duration {
-        let mut rng = rand::thread_rng();
-        Duration::from_secs(self.event_dist.sample(&mut rng) as u64)
+    fn next_payment_wait(&mut self) -> Result<Duration, PaymentGenerationError> {
+        let mut rng_guard = self
```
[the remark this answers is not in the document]
enigbe: Gone ahead to implement this. See this [comment ](https://github.com/bitcoin-dev-project/sim-ln/pull/178#discussion_r1658520552)for more information about the results I am getting from multiple simulation runs.

sim-lib/src/lib.rs
```diff
@@ -322,7 +331,7 @@ pub trait PaymentGenerator: Display + Send {
     fn payment_count(&self) -> Option<u64>;
 
     /// Returns the number of seconds that a node should wait until firing its next payment.
-    fn next_payment_wait(&self) -> time::Duration;
+    fn next_payment_wait(&mut self) -> Result<time::Duration, PaymentGenerationError>;
```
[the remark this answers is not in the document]
enigbe: Thanks for sharing this. Learned something new after reviewing your patch and doing some reading. The shared access is to the Mutex, while the exclusive access is to the type it wraps.

sim-lib/src/random_activity.rs
carlaKC: I think this is to be expected - tasks will complete in slightly different orders on each run (and payments will complete in different amounts of time), so they'll grab the common RNG is different orders. We can't make promises for totally deterministic runs, but I think that this is a good shot at it because it'll aggregate out over several runs. 

But well observed, let's make a note of it in the readme that we're not promising fully deterministic simulations.

sim-lib/src/lib.rs
```diff
@@ -435,6 +447,23 @@ enum SimulationOutput {
     SendPaymentFailure(Payment, PaymentResult),
 }
 
+type MutRngType = Arc<StdMutex<Box<dyn RngCore + Send>>>;
+
+#[derive(Clone)]
+struct MutRng(pub MutRngType);
+impl MutRng {
+    /// Creates a new MutRng
```
carlaKC, line 455: nit: explain options for seed vs no seed

sim-lib/src/lib.rs
```diff
@@ -453,6 +482,8 @@ pub struct Simulation {
     activity_multiplier: f64,
     /// Configurations for printing results to CSV. Results are not written if this option is None.
     write_results: Option<WriteResults>,
+    /// Random number generator created from fixed seed
```
carlaKC, line 485: nit: end comments with `.` - applies in a few places

sim-lib/src/lib.rs
```diff
@@ -481,8 +512,10 @@ impl Simulation {
         expected_payment_msat: u64,
         activity_multiplier: f64,
         write_results: Option<WriteResults>,
+        seed: Option<u64>,
     ) -> Self {
         let (shutdown_trigger, shutdown_listener) = triggered::trigger();
+
```
carlaKC, line 518: nit: unnecessary newline added

sim-lib/src/random_activity.rs
```diff
@@ -19,20 +22,25 @@ pub enum RandomActivityError {
     InsufficientCapacity(String),
 }
 
-/// NetworkGraphView maintains a view of the network graph that can be used to pick nodes by their deployed liquidity
-/// and track node capacity within the network. Tracking nodes in the network is memory-expensive, so we use a single
-/// tracker for the whole network (in an unbounded environment, we'd make one _per_ node generating random activity,
-/// which has a view of the full network except for itself).
+/// `NetworkGraphView` maintains a view of the network graph that can be used to pick nodes by their deployed liquidity
+/// and track node capacity within the network. The `NetworkGraphView` also keeps a handle on an optional seed that
+/// allows it to deterministically pick nodes. Tracking nodes in the network is memory-expensive, so we
```
carlaKC, line 27: nit: "keeps a handle on an optional seed" is outdated now

sim-lib/src/random_activity.rs
carlaKC: Can also confirm this by just writing a small unit test for `MutRng::new` and asserting that we get the same values for the same seeds.

sim-lib/src/random_activity.rs
```diff
@@ -94,12 +112,14 @@ impl Display for NetworkGraphView {
 /// node's capacity, it will produce payments such that the node sends multiplier * capacity over a calendar month.
 /// While the expected amount to be sent in a month and the mean payment amount are set, the generator will introduce
 /// randomness both in the time between events and the variance of payment amounts sent to mimic more realistic
-/// payment flows.
+/// payment flows. For deterministic time between events and payment amounts, the `RandomPaymentActivity` keeps a
+/// handle on a seed.
```
carlaKC, line 116: similar nit: update comment about seed

sim-lib/src/lib.rs
```diff
@@ -435,6 +447,23 @@ enum SimulationOutput {
     SendPaymentFailure(Payment, PaymentResult),
 }
 
+type MutRngType = Arc<StdMutex<Box<dyn RngCore + Send>>>;
+
+#[derive(Clone)]
+struct MutRng(pub MutRngType);
+impl MutRng {
```
carlaKC, line 454: Don't personally feel like adding a wrapper struct on `MutRngType` adds much value (assuming it's there so we can have an `impl` block + constructor) here over just passing around params of the type, but not blocking merge.

sim-lib/src/lib.rs
```diff
@@ -435,6 +447,23 @@ enum SimulationOutput {
     SendPaymentFailure(Payment, PaymentResult),
 }
 
+type MutRngType = Arc<StdMutex<Box<dyn RngCore + Send>>>;
```
carlaKC, line 450: `StdMutex` so that we don't need to worry about making this trait `async`? Let's add a comment here, perhaps we want to go down that refactor road one day, and it would be good to remember why we chose this.

carlaKC (commented):
Just some nitpicks left then this is ready to go! Thanks for picking this back up 🫶 

Go ahead and squash fixes on next push. Would also be great to add a note to the readme that this option is now available, but with limitations.

sim-lib/src/lib.rs
enigbe: It makes it easier to create new `MutRng` types within code and in tests. I think it is much cleaner to have `MutRng::new(T)` that `Arc::new(Mutex::new(Box::new(T)))`

bjohnson5 (commented):
LGTM

sim-lib/src/lib.rs
```diff
@@ -435,6 +447,36 @@ enum SimulationOutput {
...
+/// for deterministic outcomes.
+///
+/// **Note**: `StdMutex`, i.e. (`std::sync::Mutex`), is used here to avoid making the traits
+/// `DestinationGenerator` and `PaymentGenerator` async.
+type MutRngType = Arc<StdMutex<Box<dyn RngCore + Send>>>;
+
+/// Newtype for `MutRngType` to encapsulate and hide implementation details for
+/// creating new `MutRngType` types. Provides convenient API for the same purpose.
+#[derive(Clone)]
+struct MutRng(pub MutRngType);
```
carlaKC, line 462: Does not need `pub`?

sim-lib/src/lib.rs
```diff
@@ -435,6 +447,36 @@ enum SimulationOutput {
...
 }
 
+/// MutRngType is a convenient type alias for any random number generator (RNG) type that
+/// allows shared and exclusive access. This is necessary because a single RNG
+/// is to be shared across multiple `DestinationGenerator`s and `PaymentGenerator`s
+/// for deterministic outcomes.
+///
+/// **Note**: `StdMutex`, i.e. (`std::sync::Mutex`), is used here to avoid making the traits
+/// `DestinationGenerator` and `PaymentGenerator` async.
+type MutRngType = Arc<StdMutex<Box<dyn RngCore + Send>>>;
```
carlaKC, line 457: `Box` is unnecessary here?

README.md
```diff
@@ -132,6 +132,7 @@ of the random activity that is generated:
      capacity in a month.
   * `capacity-multiplier=0.5` means that each node sends half their 
     capacity in a month.
+* `--fix-seed`: a `u64` value that allows you to generate random activities deterministically from the provided seed, albeit with some limitations. The  simulations are not guaranteed to be perfectly deterministic because tasks complete in slightly different orders on each run of the simulator. With a fixed seed, we can guarantee that the order in which activities are dispatched will be deterministic.
```
carlaKC, line 135: Nice docs!

nit: double space in `The  simulations`

carlaKC (approved):
Last few comments (can be done in a followup) - thanks for all the work on this one!

PR #179, opened 2024-04-14
Update method parameter to take reference to PaymentHash

jrakibi:
This PR updates the `track_payment` method in the `LightningNode` trait to accept a reference to PaymentHash instead of taking ownership

- Changed method signature to `&PaymentHash` in the `LightningNode` trait.
- Updated all implementations and usages of `track_payment` accordingly.


carlaKC (approved):
tACK, thanks for the PR!

Issue #26, opened 2023-08-16
Feature: Multi-Implementation Support


carlaKC:
Add alternative implementations of `LightningNode` trait for: 
- [x] Core Lightning
- [x] Eclair
- [ ] LDK Node

carlaKC:
I'd suggest starting with CLN or LDK-node first, so those are _likely_ the easiest.

carlaKC:
CLN: https://crates.io/crates/cln-grpc ?

okjodom:
New tasks. Blocked by incompleteness in CLN rpc types

- [ ] CLN track payment needs to report failure reason
- [ ] CLN track payment needs to report htlc count

We use `lightning-listpays` rpc, from which we don't get either of the above response fields

sr-gi:
> New tasks. Blocked by incompleteness in CLN rpc types
> 
> * [ ]  CLN track payment needs to report failure reason
> * [x]  CLN track payment needs to report htlc count
>   
>   * fix: we use listpays `number_of_parts` where available, or default to 1
> 
> We use `lightning-listpays` rpc, from which we don't get either of the above response fields

Let me take a look at that payment error data and how to parse it

carlaKC:
I think we're probably going to have to go with our own mapping - something like this:

`PAYMENT_NOT_SENT` represents: 
* LND/`FAILURE_REASON_TIMEOUT` 
* LND/`FAILURE_REASON_NO_ROUTE`
* LND/`FAILURE_REASON_INSUFFICIENT_BALANCE`

`INCORRECT_DETAILS` represents: 
* LND/`FAILURE_REASON_INCORRECT_PAYMENT_DETAILS`
* `failure_code` / 18
* `failure_code` / 19

`PAYMENT_FAILED` represents:
* LND/`FAILURE_REASON_ERROR`
* `failure_code`/ any other value

sr-gi:
For the record: https://github.com/ElementsProject/lightning/issues/6634

carlaKC:
Dropping this down to medium now that we have LND and CLN.

sb1752:
So I spent some time looking into LDK-Node integration and wanted to share some of my findings here in case there were any suggestions before I move forward.

[...]

PR #181, opened 2024-04-24
Try to add eclair/ldk support

PurpleTimez:
I don't understand the framework, and dunno if other implems support can be useful.

### DIFF
--- a/sim-lib/src/eclair.rs
+++ b/sim-lib/src/eclair.rs
@@ -1,0 +1,79 @@
+use async_trait::async_trait;
+use bitcoin::secp256k1::PublicKey;
+use bitcoin::Network;
+
+use lightning::ln::PaymentHash;
+
+use serde::{Deserialize, Serialize};
+
+use triggered::Listener;
+
+use crate::{
+	serializers, LightningError, LightningNode, PaymentResult, NodeId, NodeInfo,
+};
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct EclairConnection {
+	#[serde(with = "serializers::serde_node_id")]
+	pub id: NodeId,
+	pub address: String,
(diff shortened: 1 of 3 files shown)


PurpleTimez:
## Eclair Support

From eclair documentation's (`README.md`), there is an API where one can install plugin written in Scala, Java or any JVM-compatible language.

I think from then a plugin could be designed to expose a GRPC interface and this interface driven by a `sim-lib` module to act sim actions (`get_info`, `send_payment`, `track_payment`, `get_node_info`, `list_channels`).

## LDK Support

LDK has a sample node called `ldk-sample`. It is a test lightning node that one can run on the command-line a REPL fashion. Internally, it's using a module (`cli.rs`) which is pipping user command to the internal interface of the LDK library.

I think a GRPC interface alternatively integrated in `ldk-sample` runtime, instead of the command-line cli could be a way to have a LDK runtime driven by `sim-lib`. Accessing directly the LDK interface allow a stable interface to act on sim actions
(`get_info`, `send_payment`, `track_payment`, `get_node_info`, `list_channels`).

I don't know if the two support approaches will work end-to-end, though I think it's an interesting way to have eclair / ldk support in `sim-ln`. I don't know how much hacking work it is, I can try.

carlaKC:
See #26 for ongoing discussion on multi-impl support

PurpleTimez:
> See https://github.com/bitcoin-dev-project/sim-ln/issues/26 for ongoing discussion on multi-impl support

Left thoughts here. Sounds eclair support better to prioritize. Dunno if way proposed is not too much maintenance.

carlaKC:
Closing boilerplate PR after discussion in #26, feel free to re-open an eclair implementation [if reasonably achievable](https://github.com/bitcoin-dev-project/sim-ln/issues/26#issuecomment-2122928053).

Issue #182, opened 2024-04-25
High CPU usage in between payments


bjohnson5:
**Describe the bug**

When `sim-ln` is waiting until the next payment, the CPU usage jumps to 100%.

I believe this can be traced back to the `produce_simulation_results` function. It contains a loop with a `tokio::select!` statement that has 3 branches. The last branch is the following:
```
track_payment = set.join_next() => {
    if let Some(res) = track_payment {
        match res {
            Ok(track_payment_res) => {
                track_payment_res?
            },
            Err(_) => {
                return Err(SimulationError::TaskError);
            },
        }
    }
}
```
The `set.join_next()` function documentation says it "Returns None if the set is empty." When `sim-ln` is not processing payments the set will be empty, so the `join_next()` function will continuously return None. 

The `tokio::select!` docs say "Waits on multiple concurrent branches, returning when the first branch completes, cancelling the remaining branches."

Because the `set.join_next()` branch is immediately returning None, it will be the first branch to complete and the loop will iterate. This leads to a high speed loop always running while `sim-ln` waits for another payment.

**To Reproduce**

Run a simple simulation that has extended time between payments.

**Config**

Running a simple simulation: 2 nodes, 1 channel, random payment activity.

bjohnson5:
@carlaKC @enigbe Have either of you experienced this issue? I am wondering if `produce_simulation_results` should not be using `join_next()` since it returns None immediately most of the time.

carlaKC:
>  I am wondering if produce_simulation_results should not be using join_next() since it returns None immediately most of the time.

Definitely not, nice catch! Would be nice to wait for all the `track_payment_res` to clean up before we exit this function - in golang-land I'd remove them from the loop and just `defer` waiting, afaik rust doesn't have something like this (closest is custom impl of `drop` perhaps?)

bjohnson5:
> > I am wondering if produce_simulation_results should not be using join_next() since it returns None immediately most of the time.
> 
> Definitely not, nice catch! Would be nice to wait for all the `track_payment_res` to clean up before we exit this function - in golang-land I'd remove them from the loop and just `defer` waiting, afaik rust doesn't have something like this (closest is custom impl of `drop` perhaps?)

I moved the join_next() outside of the `tokio::select!` statement and this seemed to fix the issue. I opened a PR, but there might be a better way to do it.

[...]

PR #183, opened 2024-04-26
Issue #182 Fixes a problematic loop in produce_simulation_results

bjohnson5:
Resolves Issue #182 by removing `set.join_next()` from the `tokio::select!` statement and waiting for tasks to finish outside of the loop. Also does some error handling in the `run` function in case we encounter an error after starting data collection threads.

### DIFF
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -559,7 +559,21 @@ impl Simulation {
         self.run_data_collection(event_receiver, &mut tasks);
 
         // Get an execution kit per activity that we need to generate and spin up consumers for each source node.
-        let activities = self.activity_executors().await?;
+        let activities = match self.activity_executors().await {
+            Ok(a) => a,
+            Err(e) => {
+                // If we encounter an error while setting up the activity_executors,
+                // we need to shutdown and wait for tasks to finish. We have started background tasks in the
+                // run_data_collection function, so we should shut those down before returning.
+                self.shutdown();
+                while let Some(res) = tasks.join_next().await {
+                    if let Err(e) = res {
+                        log::error!("Task exited with error: {e}.");
+                    }
+                }
+                return Err(e);
+            },
+        };
         let consumer_channels = self.dispatch_consumers(
             activities
                 .iter()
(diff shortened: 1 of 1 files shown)


sim-lib/src/lib.rs
```diff
@@ -559,7 +559,21 @@ impl Simulation {
...
+            Err(e) => {
+                // If we encounter an error while setting up the activity_executors,
+                // we need to shutdown and wait for tasks to finish. We have started background tasks in the
+                // run_data_collection function, so we should shut those down before returning.
+                self.shutdown();
+                while let Some(res) = tasks.join_next().await {
+                    if let Err(e) = res {
+                        log::error!("Task exited with error: {e}.");
+                    }
+                }
```
carlaKC, line 573: Let's pull this out into a closure to de-dup it a little?

sim-lib/src/lib.rs
```diff
@@ -1158,7 +1188,7 @@ async fn produce_simulation_results(
         tokio::select! {
             biased;
             _ = listener.clone() => {
-                return Ok(())
+                break;
```
carlaKC, line 1191: TIL that you can return a value with `break` in rust - perhaps we can do this here? 

`let result = loop ....` and then `break Ok(())` / `break Err(())` and then go on to the joinset wait at the end. Has the advantage of always waiting for everything to close out cleanly vs current approach where we only wait on `Ok()` exits.

carlaKC (commented):
Thanks for the fix (and detection!!), only one real comment here. 

Going to tag a patch once this goes in so that we can get a fresh docker image out.

sim-lib/src/lib.rs
bjohnson5: Any suggestions on what that closure should look like? It will have to be `async` which is always tricky to me.

sim-lib/src/lib.rs
bjohnson5: Yes, that is a great idea! Just added that in the latest commit.

sim-lib/src/lib.rs
carlaKC: Async closures are dark magic indeed :') Since we're also using it in `produce_simulation_results` it can probably just be a utility [function](https://github.com/bitcoin-dev-project/sim-ln/files/15149267/wait_fn.patch) instead. 

Closure could have been something like this: 
```
let wait_for_tasks = |mut tasks: JoinSet<()>| async move {
            if let Err(e) = res {
                log::error!("Task exited with error: {e}.");
            }
        }
};
```

carlaKC (approved):
tACK, last [comment](https://github.com/bitcoin-dev-project/sim-ln/pull/183/files#r1583029025) non-blocking (only saves us a few lines).

CPU looking muuch better 😻 
Screenshot 2024-04-29 at 8 28 04 AM

PR #184, opened 2024-04-26
Refactor: error messages to be concise and formatted correctly #40

AdamuAbba:
closes issue #40 

This PR contains minor code changes that modify the structure of some error messages. I also did not modify some "one-liners" that seemed convenient enough. I'd appreciate a review and any pointers for adjustments thank you.

### DIFF
--- a/sim-lib/src/random_activity.rs
+++ b/sim-lib/src/random_activity.rs
@@ -13,7 +13,7 @@ const SECONDS_PER_MONTH: u64 = HOURS_PER_MONTH * 60 * 60;
 
 #[derive(Debug, Error)]
 pub enum RandomActivityError {
-    #[error("Value error: {0}")]
+    #[error("Value error\nCaused by: {0}")]
     ValueError(String),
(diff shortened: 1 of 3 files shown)


sim-lib/src/random_activity.rs
```diff
@@ -13,7 +13,7 @@ const SECONDS_PER_MONTH: u64 = HOURS_PER_MONTH * 60 * 60;
 
 #[derive(Debug, Error)]
 pub enum RandomActivityError {
-    #[error("Value error: {0}")]
+    #[error("Value error\nCaused by: {0}")]
```
carlaKC, line 16: Why do we have `Caused by` here but not below?

carlaKC (commented):
I'm hesitant to just add newlines to our errors for a marginal improvement in formatting - seems like somewhere we'll quickly get inconsistent and forget so I'd like to find a more sustainable way to approach this.

carlaKC:
Closing due to inactivity, feel free to re-open if you've got an update here.

PR #186, opened 2024-05-15
workflows: pin clippy version

carlaKC:
Pin the version of clippy that we use in CI so that new releases don't break existing CI.

Testing to see whether installing clippy at a set version will help.


carlaKC:
As seen in: https://github.com/bitcoin-dev-project/sim-ln/actions/runs/8975768754/job/24651141071?pr=153

PR #188, opened 2024-06-18
Fixing the serializer for ValueOrRange

### LINKED ISSUE #187: ValueOrRange Serialize Issue
**Describe the bug**
Serializing a ValueOrRange type to json turns the object into a string. Then when deserializing the sim.json file you get this error:

`data did not match any variant of untagged enum ValueOrRange`

This is really only an issue if you are using `simln` as a library and creating simulations directly in rust (not loading `sim.json` files)

**To Reproduce**

- Use `sim-lib` to create a simulation with a list of `ActivityParser` types
- Use `serde_json::to_string` to serialize the list to json
- Write the json data to a file so that it can be used as a `sim.json` for in a later simulation
- Try to load the saved `sim.json` file and get the error

I believe this is due to the custom serializer that simply uses `serialize_str` to serialize the ValueOrRange

bjohnson5:
Using a more specific serialization process for the ValueOrRange type. This will allow the type to be correctly serialized and deserialized.

Close #187

### DIFF
--- a/sim-lib/src/serializers.rs
+++ b/sim-lib/src/serializers.rs
@@ -47,7 +47,7 @@ pub mod serde_node_id {
 
 pub mod serde_value_or_range {
     use super::*;
-    use serde::de::Error;
+    use serde::{de::Error, ser::SerializeTuple};
 
     use crate::ValueOrRange;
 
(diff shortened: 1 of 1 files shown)


bjohnson5:
@carlaKC @sr-gi Let me know what you think about this. I have been using simln in another Rust project and came across this issue when serializing my activity definitions.

sr-gi:
Nice catch! We don't use the serializer internally it went completely unnoticed.

LGTM

carlaKC (approved):
Thanks for the fix!

PR #190, opened 2024-06-28
sim-lib: only use start delay if explicitly specified

### LINKED ISSUE #189: Use payment delay for offset if no start_secs is specified
For defined activity, if we do not have `start_secs` set all of our payments will fire immediately on startup and thereafter run at regular intervals. Eg, if we have payments A running every 5 seconds and payment B every 10 seconds our payment dispatch will be:
- 0 seconds: send A and B
- 5 seconds: send A
- 10 seconds: send B

This seems like undesirable behavior when the `start_secs` field isn't set in our simulation file. Ideally, we'd distinguish between:
* `start_secs: 0` in sim file: fire the payment immediately
* No `start_secs` in sim file: only fire the payment after its desired interval

So, using the example above when no `start_secs` is specified our payment dispatch should be: 
- 0 seconds: no payments
- 5 seconds: send A
- 10 seconds: send B

This also impacts randomly generated payments - on startup all random payments will fire, _then_ we'll take a look at the wait times that's been randomly generated and wait the appropriate amount of time. These should behave similarly to when no `start_secs` is defined in the sim file for defined behavior.

carlaKC:
As is, we'll run with a 0 second start delay for:
- Random activities
- Defined activities with no `start_secs` specified

This very likely isn't the intent for defined activities (people should set `start_secs=0` if they want their payments to fire immediately), and doesn't work for random activity because we just fire all payments on start.

Instead, we switch over to using our payment wait as the start delay in the absence of this field being set for defined activities and always for random activities.

Fixes #189

### DIFF
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -185,8 +185,7 @@ pub struct ActivityParser {
     #[serde(with = "serializers::serde_node_id")]
     pub destination: NodeId,
     /// The time in the simulation to start the payment.
-    #[serde(default)]
-    pub start_secs: u16,
+    pub start_secs: Option<u16>,
     /// The number of payments to send over the course of the simulation.
     #[serde(default)]
     pub count: Option<u64>,
@@ -207,7 +206,7 @@ pub struct ActivityDefinition {
     /// The destination of the payment.
     pub destination: NodeInfo,
     /// The time in the simulation to start the payment.
-    pub start_secs: u16,
+    pub start_secs: Option<u16>,
     /// The number of payments to send over the course of the simulation.
     pub count: Option<u64>,
     /// The interval of the event, as in every how many seconds the payment is performed.
@@ -328,7 +327,7 @@ pub struct PaymentGenerationError(String);
 
 pub trait PaymentGenerator: Display + Send {
     /// Returns the time that the payments should start
-    fn payment_start(&self) -> Duration;
+    fn payment_start(&self) -> Option<Duration>;
 
     /// Returns the number of payments that should be made
     fn payment_count(&self) -> Option<u64>;
@@ -819,7 +818,9 @@ impl Simulation {
             for description in self.activity.iter() {
                 let activity_generator = DefinedPaymentActivity::new(
                     description.destination.clone(),
-                    Duration::from_secs(description.start_secs.into()),
+                    description
+                        .start_secs
+                        .map(|start| Duration::from_secs(start.into())),
                     description.count,
                     description.interval_secs,
                     description.amount_msat,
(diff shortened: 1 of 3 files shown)


carlaKC:
cc @bjohnson5 WDYT?

sim-lib/src/lib.rs
```diff
@@ -1037,14 +1038,15 @@ async fn produce_events<N: DestinationGenerator + ?Sized, A: PaymentGenerator +
...
-        let wait: Duration = if current_count == 0 {
-            let start = node_generator.payment_start();
-            if start != Duration::from_secs(0) {
-                log::debug!(
-                    "Using a start delay. The first payment for {source} will be at {:?}.",
-                    start
-                );
-            }
+        // On our first run, add a one-time start delay if set for our payment.
+        let opt_start = node_generator.payment_start();
```
bjohnson5, line 1042: nit: This line doesn't really need to happen every loop. 

Calling `payment_start()` and assigning `opt_start` is only necessary on the first iteration. It might be worth moving that into the second condition so that `if current_count == 0` fails it will short circuit and not execute the `payment_start()` call.

```rust
 let wait = if current_count == 0 && node_generator.payment_start().is_some() {
      let start = node_generator.payment_start().unwrap(); // Safe due to is_some check above.
      log::debug!(
          "Using a start delay. The first payment for {source} will be at {:?}.",
          start
      );

      start
  } else {
...
```
Still not ideal to call `payment_start()` twice but I could not think of a way to get the value into the if block otherwise.

bjohnson5:
@carlaKC Untested ACK. I can test this change next week sometime, but don't let that be a blocker for merging.

sim-lib/src/lib.rs
carlaKC: > Still not ideal to call payment_start() twice but I could not think of a way to get the value into the if block otherwise.

Yeah, this was my original reasoning for creating `op_start` (+ learned that `if let` has more limited syntax than I thought). Agree that this is probably the best we can do - will add a note on the trait that it's expected to return the same value on every call.

carlaKC: Actually just went for a more verbose set of branches, I think it's nicer than adding an external expectation on `payment_start`'s behavior (even if in practice it always is the same value) + added a comment explaining.

carlaKC:
Merging so that I can tag a patch with the fix (need the docker image elsewhere). Would be great to get post-merge review on this one anyway 🙏

sim-lib/src/lib.rs
bjohnson5: Totally agree, that looks really good now. It is also easier to read and understand what is happening.

PR #193, opened 2024-08-01
shutdown: Select on listener on all channel sends

carlaKC:
Ran into a deadlock on shutdown when running with > 100 nodes. Cause was multiple `send` calls trying to pass events on to `receiver` channels that had already exited (because `listener` signaled that they should), example [here](https://github.com/bitcoin-dev-project/sim-ln/pull/164#discussion_r1483419179). 

This is difficult to hit with smaller networks, because most of the time everything just exits with the `listener` signal - it's only when we're in the middle of an operation in multiple tasks where we hit this. 

This PR takes the approach of just selecting on sender/listener to exit on the sender side. An alternative approach would be to use `receiver.close()` and drain the channel. This is a little more boilerplate to update all of our receiver sites, so I went with the sender approach. 

Tracing via [tokio-console](https://github.com/tokio-rs/console) is also added behind a development flag, so that debugging issues like this is easier in the future (it has gorgeous dashboards 😻 ).

Related to #162.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -29,7 +29,9 @@ jobs:
         run: sudo apt install -y protobuf-compiler
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
-      - name: cargo build
-        run: cargo build --release --all-features
+      - name: Install sim-cli
+        run: make install
+      - name: Install dev sim-cli
+        run: make dev-install
       - name: cargo test
         run: cargo test --all-features --all-targets --benches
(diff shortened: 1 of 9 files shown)


carlaKC:
Looking into the build error, can't reproduce locally.

sr-gi:
> Looking into the build error, can't reproduce locally.

I think it may be a bug upstream with `time`. Updating the library may fix it.

ref: https://github.com/time-rs/time/issues/693

carlaKC:
> Updating the library may fix it.

Indeed! Didn't break with my version of rustc - will rebsae this once #195 is in!

enigbe (approved):
Good job on this @carlaKC. Happy to approve these changes.

Issue #196, opened 2024-08-13
Expose the simulation results through sim-lib


bjohnson5:
**Describe desired feature**
Currently the payment results are logged to a csv file, it would be nice to be able to query simln during the simulation and get the current stats for the running simulation. This could be achieved by adding some public functions to `sim-lib` that return the current simulation results. This would allow a project that uses `sim-lib` to get payment statistics directly during the simulation, in addition to the results being written out to a csv file.

**Use case for feature**

Projects that depend on `sim-lib` could get runtime statistics to show the user while the simulation is running.

**Would you like to contribute code for this feature?**

Yes, I would like to submit a PR for this feature.

PR #197, opened 2024-08-13
Exposing the simulation results through public functions in sim-lib

bjohnson5:
Closes Issue #196. Adds public functions to sim-lib that let a library user get the current simulation statistics.

### DIFF
--- a/sim-lib/src/lib.rs
+++ b/sim-lib/src/lib.rs
@@ -494,6 +494,8 @@ pub struct Simulation {
     write_results: Option<WriteResults>,
     /// Random number generator created from fixed seed.
     seeded_rng: MutRng,
+    /// Results logger that holds the simulation statistics.
+    results: Arc<Mutex<PaymentResultLogger>>,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


sim-lib/src/lib.rs
```diff
@@ -770,7 +781,7 @@ impl Simulation {
             }
         });
 
-        let result_logger = Arc::new(Mutex::new(PaymentResultLogger::new()));
+        let result_logger = self.results.clone();
```
carlaKC, line 784: Do we need both `let result_logger = self.results.clone();` and `let result_logger_clone = result_logger.clone();` below? Seems like we could reduce to a single clone?

carlaKC (approved):
lgtm, one nit

sim-lib/src/lib.rs
bjohnson5: Hm. I think you need a results logger in the `run_results_logger` spawned task and the `consume_simulation_results` spawned task as well as the main simulation thread. So the Arc should have a reference count of 3? Which means you need to clone it twice. Am I understanding that correctly? Very possible I am missing something.

carlaKC: Nope you're totally right - missed `consume_simulation_results` below!

Issue #198, opened 2024-08-22
Installation instructions needs to mention protobuf as a requirement 


zealsham:
**Describe the bug**
when trying to build sim-cli  , i ran into an issue where the build process panicked because i did not have the protoc binary installed on my device 

```
failed to run custom build command for cln-grpc v0.1.6

Caused by:
  process didn't exit successfully: /Users/shammah/sim-ln/target/release/build/cln-grpc-a808399b7e9d80ba/build-script-build (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=proto/node.proto
  cargo:rerun-if-changed=proto

  --- stderr
  thread 'main' panicked at /Users/shammah/.cargo/registry/src/index.crates.io-6f17d22bba15001f/prost-build-0.11.9/src/lib.rs:1457:10:
  Could not find protoc installation and this build crate cannot proceed without
      this knowledge. If protoc is installed and this crate had trouble finding
      it, you can set the PROTOC environment variable with the specific path to your
      installed protoc binary.You could try running brew install protobuf or downloading it from https://github.com/protocolbuffers/protobuf/releases

  For more information: https://docs.rs/prost-build/#sourcing-protoc

  note: run with RUST_BACKTRACE=1 environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: failed to compile sim-cli v0.1.0 (/Users/shammah/sim-ln/sim-cli), intermediate artifacts can be found at /Users/shammah/sim-ln/target.
To reuse those artifacts with a future compilation, set the environment variable CARGO_TARGET_DIR to that path.
```

it would ease the installation process if  the installation instructions mentions something like , "make sure protobuf is installed on your device "

i can push a Pr if you sim-ln devs agree with this suggestion

carlaKC:
A PR with instructions would be great, thank you!

zealsham:
so i went on a mission to install sim-ln on linux , windows and macos for this  task . However on m1 mac i'm running into an issue that i spent a couple of days trying to resolve and i could not . 

After starting a local network with Polar and running sim-cli with a test configuration , i get the following error 
``` 
[hyper::client::client] Connection is HTTP/1, but request requires HTTP/2
Error: Get info error: status: Unknown, message: "request has unsupported HTTP version", details: [], metadata: MetadataMap { headers: {} }
```
[...]

PR #199, opened 2024-09-04
Improved installations instructions to include Protobuf installation

zealsham:
Protobuf is an important dependency of the sim-ln project  as it powers GRPC communication which the lighning nework uses for client communications . 

During installation of Sim-ln , one is likely to run into error due to protobuf not being installed on their device . This PR spells out how to install protobuf. This PR is as result of this issues [198](https://github.com/bitcoin-dev-project/sim-ln/issues/198)


carlaKC:
Closing due to inactivity, please re-open if you'd like to pick this back up!

PR #201, opened 2024-11-12
sim-node: validate channel policy and enforce unique scid

carlaKC:
Add some validation to make sure that we get sane channels in, specifically:
- Policies that have valid max in flight / htlc size (this would be caught by our sanity checks, but this catches it early)
- No repeated channel ids, this will break other things (and can't be true irl)

### DIFF
--- a/sim-lib/src/sim_node.rs
+++ b/sim-lib/src/sim_node.rs
@@ -122,6 +122,25 @@ pub struct ChannelPolicy {
     pub fee_rate_prop: u64,
 }
 
+impl ChannelPolicy {
+    /// Validates that the channel policy is acceptable for the size of the channel.
+    fn validate(&self, capacity_msat: u64) -> Result<(), SimulationError> {
+        if self.max_in_flight_msat > capacity_msat {
+            return Err(SimulationError::SimulatedNetworkError(format!(
+                "max_in_flight_msat {} > capacity {}",
+                self.max_in_flight_msat, capacity_msat
+            )));
+        }
+        if self.max_htlc_size_msat > capacity_msat {
+            return Err(SimulationError::SimulatedNetworkError(format!(
+                "max_htlc_size_msat {} > capacity {}",
+                self.max_htlc_size_msat, capacity_msat
+            )));
+        }
+        Ok(())
+    }
+}
+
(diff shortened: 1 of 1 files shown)


sim-lib/src/sim_node.rs
```diff
@@ -291,6 +310,21 @@ impl SimulatedChannel {
...
+        if self.node_1.policy.pubkey == self.node_2.policy.pubkey {
+            return Err(SimulationError::SimulatedNetworkError(format!(
+                "Channel should have distinct node pubkeys, got: {} for both nodes.",
+                self.node_1.policy.pubkey
+            )));
+        }
+
+        self.node_1.policy.validate(self.capacity_msat)?;
+        self.node_2.policy.validate(self.capacity_msat)?;
+
```
bjohnson5, line 324: Would it be worth also validating that the `ChannelState`'s have valid local balances here? It appears that you could create a `SimulatedChannel` with a total capacity that is different than `node_1.local_balance_msat + node_2.local_balance_msat`? This is caught in a sanity check later, but if the goal of this PR is to catch things earlier, this could be an easy one to check.

```rust
if self.node_1.local_balance_msat + self.node_2.local_balance_msat != self.capacity_msat {
    return Err(SimulationError::SimulatedNetworkError(format!(
        "Local balances (node1: {}, node2: {}) do not equal the total channel capacity: {}",
        self.node_1.local_balance_msat,
        self.node_2.local_balance_msat,
        self.capacity_msat
    )));            
}
```
This does not take into account in-flight sats, but if we are validating the creation of this channel that might not matter?

bjohnson5 (approved):
LGTM! I tested a few invalid channel policies and scids and the correct errors were thrown. Left one totally optional suggestion.

sim-lib/src/sim_node.rs
carlaKC: Yeah that's great! 

I think we can take it further and not allow a non-zero in-flight balance, since as-is we'll never be able to clear out that in-flight balance if you start with it. Sure we _might_ get a use case for that one day, but start strict + follow complaint-driven-develpoment sgtm.

Pushed the above + a (quite lazy) test

PR #202, opened 2024-11-12
multi: add cfg struct to simulation

carlaKC:


sim-lib/src/lib.rs
```diff
@@ -498,6 +492,38 @@ pub struct Simulation {
...
+        write_results: Option<WriteResults>,
+        seed: Option<u64>,
+    ) -> Self {
+        Self {
+            total_time: total_time.map(|x| Duration::from_secs(x as u64)),
+            expected_payment_msat,
+            activity_multiplier,
+            write_results,
+            seeded_rng: MutRng::new(seed),
+            results: Arc::new(Mutex::new(PaymentResultLogger::new())),
```
bjohnson5, line 509: Is there a particular reason that the `results` Arc object needs to be in the `SimulationCfg` struct? I would expect the `SimulationCfg` type to be entirely user defined values that are simply simulation settings. The `PaymentResultLogger` feels more like a runtime object that would be a part of the `Simulation` type.

No technical reason why this won't work, just a thought on overall design. Feel free to ignore.

bjohnson5 (approved):
Ran a quick sanity check test and left one optional thought. Everything appears to be working as normal.

sim-lib/src/lib.rs
carlaKC: > Is there a particular reason that the results Arc object needs to be in the SimulationCfg struct?

None, nice catch! I just copied everything over, it def belongs in `Simulation` 👍

Issue #208, opened 2024-12-03
Unable to use CLN node as destination in defined activity


bjohnson5:
**Describe the bug**
When I attempt to use a CLN node as a destination in the defined activity, the simulation does not start and gives the error: 

```
ValidationError("Destination node does not support keysend, cln-0000(0287eb...c97d33)"
```

However, CLN appears to have keysend enabled and it can send keysend payments when it is used as the source node in the defined activity.

**To Reproduce**

Configure a simple defined activity with an LND node as the source and a CLN node as the destination.

**Possible Solution**
I believe this is due to the feature flags not being parsed correctly when creating a CLN node in SimLN.

In the `get_node_info` function in `cln.rs` the feature bits are reversed before calling `NodeFeatures::from_le_bytes`
```rust
        if let Some(node) = nodes.pop() {
            Ok(NodeInfo {
                pubkey: *node_id,
                alias: node.alias.unwrap_or(String::new()),
                features: node
                    .features
                    .clone()
                    .map_or(NodeFeatures::empty(), |mut f| {
                        // We need to reverse this given it has the CLN wire encoding which is BE
                        f.reverse();
                        NodeFeatures::from_le_bytes(f)
                    }),
            })
``` 

But in the `new` function for `ClnNode` the feature bits are passed directly into `NodeFeatures::from_le_bytes`.
```rust
        let features = if let Some(features) = our_features {
            NodeFeatures::from_le_bytes(features.node)
        } else {
            NodeFeatures::empty()
        };

        Ok(Self {
            client,
            info: NodeInfo {
                pubkey,
                features,
                alias,
            },
        })
```

bjohnson5:
@carlaKC have you tried using a default CLN node as a destination? Still wondering if this has something to do with my config.

PR #209, opened 2024-12-03
Fixing the CLN feature bit parsing

bjohnson5:
Resolves Issue #208 by flipping the feature bits before parsing them.

### DIFF
--- a/simln-lib/src/cln.rs
+++ b/simln-lib/src/cln.rs
@@ -87,9 +87,8 @@ impl ClnNode {
             .map_err(|err| LightningError::GetInfoError(err.to_string()))?;
         connection.id.validate(&pubkey, &mut alias)?;
 
-        let features = if let Some(mut features) = our_features {
-            features.node.reverse();
-            NodeFeatures::from_le_bytes(features.node)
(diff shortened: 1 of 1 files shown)


simln-lib/src/cln.rs
```diff
@@ -87,7 +87,8 @@ impl ClnNode {
             .map_err(|err| LightningError::GetInfoError(err.to_string()))?;
         connection.id.validate(&pubkey, &mut alias)?;
 
-        let features = if let Some(features) = our_features {
+        let features = if let Some(mut features) = our_features {
+            features.node.reverse();
             NodeFeatures::from_le_bytes(features.node)
```
sr-gi, line 92: You should be able to use `NodeFeatures::from_be_bytes` if I'm not mistaken. I PRd a change to make that public on LDK a while ago but it looks like we forgot to apply the changes downstream

bjohnson5: Ah even better. I will switch both calls to use `from_be_bytes`. Thanks!

carlaKC (approved):
utACK

PR #212, opened 2025-01-24
simln-lib/feat: Parse address detecting if starts with https

### LINKED ISSUE #154: Do not require `https` in `address` fields on the simulation file
Currently, we force addresses in the simulation file to start with `https://`. That's not really user-friendly. 

Make it so the can be specified but not required (as in, if they are not required we just add it in).

f3r10:
fix #154 
Detect if the parsed address starts with the https protocol; if not, add it without requiring a user to do it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ information is required:
 ```
 {
   "id": <node_id>,
-  "address": https://<ip:port or domain:port>,
+  "address": <ip:port or domain:port>,
   "macaroon": <path_to_selected_macaroon>,
   "cert": <path_to_tls_cert>
 }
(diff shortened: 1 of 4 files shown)


simln-lib/src/lib.rs
```diff
@@ -48,6 +48,9 @@ pub enum NodeId {
     Alias(String),
 }
 
+#[derive(Serialize, Debug, Clone)]
+pub struct Address(String);
```
carlaKC, line 52: I think that we can live without this type alias and just use `#[serde(with = "serializers::serde_address")]` to point to custom serialization/deserialization of a `String` to reduce LOC a little.

carlaKC:
Also a small nitpick on commit naming (which is unconventional mostly personal preference, so feel free to ignore), generally we've been using the structure: `scope/{action}`, eg `sim-ln/refactor`, `sim-cli/test` followed by message.

Also in this case it's not actually a refactor, it's a (tiny) `feat`!

carlaKC (approved):
tACK, thanks for the PR 🙏 

such a nice UX improvement to not need to `https://` everything!

PR #213, opened 2025-01-27
chore(sim-ln): Bump version cargo and workflow

f3r10:
Bump cargo version


f3r10:
@carlaKC would be enough just bumping `sim-cli` and `simln-lib`? 
or would it be better to run `cargo update` updating all the dependencies :thinking:

carlaKC:
Ah fun, we get some nice new clippy errors when we upgrade :') can fix them in the same commit as the version bump

f3r10:
weird, I do not get those clippy errors locally :thinking:

carlaKC:
I get them on `rustc 1.84.0 (9fc6b4312 2025-01-07)`, checked toolchain version and it's also on [1.84.0](https://github.com/carlaKC/sim-ln/actions/runs/12996514078/job/36245459793#step:5:9).

carlaKC:
The last two clippy fix commits can be squashed to the update commit (https://github.com/bitcoin-dev-project/sim-ln/pull/213/commits/07b2147865e47c905bbc1a5c9622d2323a919e0b), then this is gtg!

carlaKC (approved):
LGTM, let's get this in so we can move on the other prs.

nitty-nit: I would have squashed 9d431ea7f48bc1d5c7ea0088af8971e8b29fc8ae into 07b2147865e47c905bbc1a5c9622d2323a919e0b so that clippy's not complaining on any of the commits

PR #218, opened 2025-02-11
simln-lib/feat: do not allow defined activity with zero amount

### LINKED ISSUE #176: Validation: Do not allow defined activity with zero amount
Currently, we allow [zero amount](https://github.com/bitcoin-dev-project/sim-ln/blob/main/sim-lib/src/lib.rs#L941) payments to be skipped over to allow for the possibility that our random payment amount generation spits out a zero value every once in an hour. 

Right now we don't validate that we're provided with a non-zero defined activity amount, so we'll hit the same check and just endlessly not dispatch payments if we're given a zero value. 

We should update [validate_activity](https://github.com/bitcoin-dev-project/sim-ln/blob/main/sim-lib/src/lib.rs#L441) to disallow zero values.

f3r10:
fix #176


f3r10:
@carlaKC what do you think about my approach to creating the test?

simln-lib/src/lib.rs
```diff
@@ -1559,4 +1592,47 @@ mod tests {
...
+            bitcoin::secp256k1::PublicKey,
+            std::sync::Arc<tokio::sync::Mutex<dyn crate::LightningNode>>,
+        > = HashMap::new();
+        clients.insert(
+            node_1.pubkey,
+            std::sync::Arc::new(tokio::sync::Mutex::new(mock_node_1)),
+        );
+        clients.insert(
+            node_2.pubkey,
+            std::sync::Arc::new(tokio::sync::Mutex::new(mock_node_2)),
```
carlaKC, line 1621: nit: can import these in `use` statements

carlaKC (commented):
approach looks good, just some minor comments to cut down the test a little

simln-lib/src/lib.rs
```diff
@@ -1559,4 +1593,39 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_validate_zero_amount_no_valid() {
+        let mut node_1 = test_utils::create_nodes(1, 100_000)
+            .first()
+            .unwrap()
+            .0
+            .clone();
+        node_1.features.set_keysend_optional();
+        let mut node_2 = test_utils::create_nodes(1, 0).first().unwrap().0.clone();
+        node_2.features.set_keysend_optional();
```
carlaKC, line 1606: Rather than setting keysend here, let's update `create_nodes` to set this feature bit? We just didn't need it in the past, which is why it's an empty vector.

simln-lib/src/lib.rs
```diff
@@ -1559,4 +1593,39 @@ mod tests {
             payment_interval
         );
     }
+
+    #[tokio::test]
+    async fn test_validate_zero_amount_no_valid() {
+        let mut node_1 = test_utils::create_nodes(1, 100_000)
```
carlaKC, line 1599: nit: create two nodes at once using `create_nodes(2, 100_000)`?

simln-lib/src/lib.rs
```diff
@@ -1559,4 +1593,39 @@ mod tests {
...
+            .0
+            .clone();
+        node_1.features.set_keysend_optional();
+        let mut node_2 = test_utils::create_nodes(1, 0).first().unwrap().0.clone();
+        node_2.features.set_keysend_optional();
+
+        let mock_node_1 = MockLightningNode::new();
+        let mock_node_2 = MockLightningNode::new();
+        let mut clients: HashMap<
+            bitcoin::secp256k1::PublicKey,
```
carlaKC, line 1611: nit: import with `use`, same for `SimulationConfig` below - generally have always been importing like this in the project rather than inlining

carlaKC (approved):
just some last nitpicks on the test, non-blocking if you'd like to merge as-is!

simln-lib/src/lib.rs
f3r10: ok, maybe it would be better to open another PR doing this? because several places call `create_nodes` that are not related with this PR. WDYT?

simln-lib/src/lib.rs
f3r10: yeah much better :sweat_smile:

simln-lib/src/lib.rs
carlaKC: I don't think it'll break any existing tests, so I think we can go ahead in this PR? I was thinking just set it inside of `create_nodes` so that fn signature doesn't change so it would be a small change.

_But_ if I'm wrong (always the most likely outcome 😅 ) and it breaks a bunch of stuff then yes let's do a followup!

carlaKC (approved):
tACK bb4b7e42eb6fb530cb26b036a9adb9940e491622!

Going to go ahead and merge this so that we don't run into conflicts with any of the incoming PRs. I would like to [move keysend into test utils](https://github.com/bitcoin-dev-project/sim-ln/pull/218/commits/bb4b7e42eb6fb530cb26b036a9adb9940e491622#r1978173523), but we can do that in another PR.

Issue #217, opened 2025-02-07
Bug: Interaction between simulation file path and results directory


carlaKC:
**Describe the bug**

If you run the simulator with two relative paths, it'll look for the simulation file in the data dir rather than the path provided and fail:
```
sim-cli --clock-speedup 1000 -s "lvl_acinq/simln_peacetime.json" -d "lvl_acinq/peacetime/"

[sim_cli::parsing] Simulation file 'lvl_acinq/peacetime/lvl_acinq/simln_peacetime.json' does not exist.
Error: no simulation files found in /Users/carla/Work/jam-ln/data/lvl_acinq/peacetime.
```

We should check whether we've got a simfile first and only fall back to the dir if it's not specified.

alexzaitsev:
As I'm almost done with https://github.com/bitcoin-dev-project/sim-ln/pull/225, I'd like to work on this next.

alexzaitsev:
@carlaKC should it be closed?

carlaKC:
> [@carlaKC](https://github.com/carlaKC) should it be closed?

Yes! In future you can say "Fixes {issue number}" in your PR and then it'll auto-close the issue on merge!

PR #228, opened 2025-03-08
sim-cli/bug: Interaction between simulation file path and results directory

alexzaitsev:
Addressing https://github.com/bitcoin-dev-project/sim-ln/issues/217
Logic:
1. If sim file is present (absolute or relative, doesn't matter) - use it.
2. If it's not present as it is - fallback to the current behaviour:
     a. try to join with data_dir, if it's present -great
     b. if it's not present - fallback to datadir lookup
3. If it's neither present of relative path - fallback to datadir lookup.

### DIFF
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -233,16 +233,18 @@ async fn main() -> anyhow::Result<()> {
 }
 
 async fn read_sim_path(data_dir: PathBuf, sim_file: PathBuf) -> anyhow::Result<PathBuf> {
-    let sim_path = if sim_file.is_relative() {
-        data_dir.join(sim_file)
+    if sim_file.exists() {
+        Ok(sim_file)
+    } else if sim_file.is_relative() {
+        let sim_path = data_dir.join(sim_file);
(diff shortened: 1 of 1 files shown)


carlaKC (approved):
tACK 94f56c6834b96e929df98785df4c8fba95620913, such a nice UX improvement 🥇 

Cases tested:

1) Custom sim file and custom data dir: `sim-cli -s custom_sim.json -d custom_data_dir/`
```
custom_sim.json
custom_data_dir/
```

2) Custom sim file in custom data dir: `sim-cli -s custom_sim.json -d custom_data_dir/`:
```
custom_data_dir/custom_sim.json
```

3)  Default sim file in custom data dir: `sim-cli -d custom_data_dir`
```
custom_data_dir/sim.json
```

---

The _only_ this that _could_ be confusing with this UX is if you have a sim file with the same name in multiple locations, eg:
```
duplicate_sim.json
custom_data_dir/duplicate_sim.json
```

`sim-cli -s custom_sim.json -d custom_data_dir/` will use the one in the current directory, which I think is the right decision.

PR #229, opened 2025-03-09
fix memory buildup from JoinSet

elnosh:
I ended up taking a different approach from what I mentioned here https://github.com/bitcoin-dev-project/sim-ln/pull/221#issuecomment-2707278793 to fix the memory buildup.

This adds a new dependency `tokio-util` but since it's from tokio which is already used, I though it would be ok. If not, let me know and I can try something else. 

It uses a `TaskTracker` from `tokio-util` which works very similar to a `JoinSet` but with the difference that it will free the memory from tasks when they are finished without having to call something like `join_next` for a `JoinSet`.

From their docs https://docs.rs/tokio-util/latest/tokio_util/task/task_tracker/struct.TaskTracker.html#comparison-to-joinset: 

> A [JoinSet](https://docs.rs/tokio/1.43.0/x86_64-unknown-linux-gnu/tokio/task/join_set/struct.JoinSet.html) keeps track of the return value of every inserted task. This means that if the caller keeps inserting tasks and never calls [join_next](https://docs.rs/tokio/1.43.0/x86_64-unknown-linux-gnu/tokio/task/join_set/struct.JoinSet.html#method.join_next), then their return values will keep building up and consuming memory, even if most of the tasks have already exited. This can cause the process to run out of memory. With a TaskTracker, this does not happen. Once tasks exit, they are immediately removed from the TaskTracker.

I tested it with the scenario from here: https://github.com/carlaKC/sim-ln/tree/review-club-memleak and it's avoiding the memory buildup that was happening with the `JoinSet`.

### DIFF
--- a/simln-lib/Cargo.toml
+++ b/simln-lib/Cargo.toml
@@ -32,6 +32,7 @@ serde_millis = "0.1.1"
 rand_distr = "0.4.3"
 mockall = "0.12.1"
 rand_chacha = "0.3.1"
+tokio-util = { version = "0.7.13", features = ["rt"] }
 
 [dev-dependencies]
 ntest = "0.9.0"
--- a/simln-lib/src/lib.rs
+++ b/simln-lib/src/lib.rs
@@ -19,8 +19,8 @@ use std::{collections::HashMap, sync::Arc, time::SystemTime};
 use thiserror::Error;
 use tokio::sync::mpsc::{channel, Receiver, Sender};
 use tokio::sync::Mutex;
-use tokio::task::JoinSet;
 use tokio::{select, time, time::Duration};
+use tokio_util::task::TaskTracker;
 use triggered::{Listener, Trigger};
 
 use self::defined_activity::DefinedPaymentActivity;
@@ -518,6 +518,9 @@ pub struct Simulation {
     activity: Vec<ActivityDefinition>,
     /// Results logger that holds the simulation statistics.
     results: Arc<Mutex<PaymentResultLogger>>,
+    /// Track all tasks spawned for use in the simulation. When used in the `run` method, it will wait for
+    /// these tasks to complete before returning.
+    tasks: TaskTracker,
     /// High level triggers used to manage simulation tasks and shutdown.
     shutdown_trigger: Trigger,
     shutdown_listener: Listener,
@@ -546,6 +549,7 @@ impl Simulation {
         cfg: SimulationCfg,
         nodes: HashMap<PublicKey, Arc<Mutex<dyn LightningNode>>>,
         activity: Vec<ActivityDefinition>,
+        tasks: TaskTracker,
     ) -> Self {
         let (shutdown_trigger, shutdown_listener) = triggered::trigger();
         Self {
@@ -553,6 +557,7 @@ impl Simulation {
             nodes,
             activity,
             results: Arc::new(Mutex::new(PaymentResultLogger::new())),
+            tasks,
             shutdown_trigger,
             shutdown_listener,
         }
@@ -644,7 +649,19 @@ impl Simulation {
         Ok(())
     }
 
+    /// run until the simulation completes or we hit an error.
+    /// Note that it will wait for the tasks in self.tasks to complete
+    /// before returning.
     pub async fn run(&self) -> Result<(), SimulationError> {
+        self.internal_run().await?;
+        // Close our TaskTracker and wait for any background tasks
+        // spawned during internal_run to complete.
+        self.tasks.close();
+        self.tasks.wait().await;
+        Ok(())
+    }
+
+    async fn internal_run(&self) -> Result<(), SimulationError> {
         if let Some(total_time) = self.cfg.total_time {
             log::info!("Running the simulation for {}s.", total_time.as_secs());
         } else {
@@ -659,7 +676,6 @@ impl Simulation {
             self.activity.len(),
             self.nodes.len()
         );
-        let mut tasks = JoinSet::new();
 
         // Before we start the simulation up, start tasks that will be responsible for gathering simulation data.
         // The event channels are shared across our functionality:
@@ -668,21 +684,15 @@ impl Simulation {
         // - Event Receiver: used by data reporting to receive events that have been simulated that need to be
         //   tracked and recorded.
         let (event_sender, event_receiver) = channel(1);
-        self.run_data_collection(event_receiver, &mut tasks);
+        self.run_data_collection(event_receiver, &self.tasks);
 
         // Get an execution kit per activity that we need to generate and spin up consumers for each source node.
         let activities = match self.activity_executors().await {
             Ok(a) => a,
             Err(e) => {
                 // If we encounter an error while setting up the activity_executors,
-                // we need to shutdown and wait for tasks to finish. We have started background tasks in the
-                // run_data_collection function, so we should shut those down before returning.
+                // we need to shutdown and return.
                 self.shutdown();
-                while let Some(res) = tasks.join_next().await {
-                    if let Err(e) = res {
-                        log::error!("Task exited with error: {e}.");
-                    }
-                }
                 return Err(e);
             },
         };
@@ -692,27 +702,20 @@ impl Simulation {
                 .map(|generator| generator.source_info.pubkey)
                 .collect(),
             event_sender.clone(),
-            &mut tasks,
+            &self.tasks,
         );
 
         // Next, we'll spin up our actual producers that will be responsible for triggering the configured activity.
-        // The producers will use their own JoinSet so that the simulation can be shutdown if they all finish.
-        let mut producer_tasks = JoinSet::new();
+        // The producers will use their own TaskTracker so that the simulation can be shutdown if they all finish.
+        let producer_tasks = TaskTracker::new();
         match self
-            .dispatch_producers(activities, consumer_channels, &mut producer_tasks)
+            .dispatch_producers(activities, consumer_channels, &producer_tasks)
             .await
         {
             Ok(_) => {},
             Err(e) => {
-                // If we encounter an error in dispatch_producers, we need to shutdown and wait for tasks to finish.
-                // We have started background tasks in the run_data_collection function,
-                // so we should shut those down before returning.
+                // If we encounter an error in dispatch_producers, we need to shutdown and return.
                 self.shutdown();
-                while let Some(res) = tasks.join_next().await {
-                    if let Err(e) = res {
-                        log::error!("Task exited with error: {e}.");
-                    }
-                }
                 return Err(e);
             },
         }
@@ -720,12 +723,9 @@ impl Simulation {
         // Start a task that waits for the producers to finish.
         // If all producers finish, then there is nothing left to do and the simulation can be shutdown.
         let producer_trigger = self.shutdown_trigger.clone();
-        tasks.spawn(async move {
-            while let Some(res) = producer_tasks.join_next().await {
-                if let Err(e) = res {
-                    log::error!("Producer exited with error: {e}.");
-                }
-            }
+        self.tasks.spawn(async move {
+            producer_tasks.close();
+            producer_tasks.wait().await;
             log::info!("All producers finished. Shutting down.");
             producer_trigger.trigger()
         });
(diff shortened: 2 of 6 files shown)


carlaKC:
> It uses a TaskTracker from tokio-util which works very similar to a JoinSet but with the difference that it will free the memory from tasks when they are finished without having to call something like join_next for a JoinSet.

Cool 😻 happy to add the dep!

A few bigger picture thoughts for this PR:
1. Let's go ahead and replace all the `Joinset`s so that we can be consistent across the project
2. WDYT about having a single `TaskTracker` that we pass into `Simulation` and `SimGraph`? That way we can get rid of `SimGraph.wait_for_shutdown` and `Simulation.run` takes care of waiting for all our tasks to clean up.

I'm also tempted to switch out `triggered` with [CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) to standardize on this library, but one thing at a time :')

elnosh:
> WDYT about having a single `TaskTracker` that we pass into `Simulation` and `SimGraph`? That way we can get rid of `SimGraph.wait_for_shutdown` and `Simulation.run` takes care of waiting for all our tasks to clean up.

sounds good to me :) Only a couple of questions. I see the `JoinSet`s used in `Simulation` are created as needed inside of `Simulation.run`, so you mean have something like a shared `TaskTracker` as fields in both the `Simulation` and `SimGraph`? If that's the case, they will need to be created together in a method like this one: https://github.com/carlaKC/sim-ln/blob/review-club-memleak/simln-lib/src/lib.rs#L590 when `SimGraph` starts being used?

> Let's go ahead and replace all the `Joinset`s so that we can be consistent across the project

cool, will do and consolidate on `TaskTracker` when we settle on approach regarding previous point of sharing it between `Simulation` and `SimGraph`. 


> I'm also tempted to switch out `triggered` with [CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) to standardize on this library, but one thing at a time :')

👍

carlaKC:
> so you mean have something like a shared TaskTracker as fields in both the Simulation and SimGraph

Yes that's what I was thinking, you create one `TaskTracker` in `main.rs` and then pass it in to whatever constructor needs it. That approach will also be helpful for people using sim-ln as a library - you can provide the simulator with a `TaskTracker` and also use it for whatever other tasks you need to run.

elnosh:
I've added the changes as suggested to replace `JoinSet`s with `TaskTracker` in other places and adding a tasks field to `Simulation`. Still creating a `producer_tasks` inside `run` to track activity tasks alone and know if it can shutdown if those have finished.

simln-lib/src/lib.rs
```diff
@@ -678,11 +681,8 @@ impl Simulation {
                 // we need to shutdown and wait for tasks to finish. We have started background tasks in the
                 // run_data_collection function, so we should shut those down before returning.
                 self.shutdown();
-                while let Some(res) = tasks.join_next().await {
-                    if let Err(e) = res {
-                        log::error!("Task exited with error: {e}.");
-                    }
-                }
+                self.tasks.close();
+                self.tasks.wait().await;
```
elnosh, line 685: I'm usually calling these two methods to wait for the tasks to finish after calling `shutdown` so I'm wondering if maybe they could be inside the `shutdown` method itself and wait for the tasks there?

simln-lib/src/lib.rs
```diff
@@ -746,18 +740,9 @@ impl Simulation {
...
-        while let Some(res) = tasks.join_next().await {
-            if let Err(e) = res {
-                log::error!("Task exited with error: {e}.");
-                success = false;
-            }
-        }
-
-        success.then_some(()).ok_or(SimulationError::TaskError)
+        self.tasks.close();
+        self.tasks.wait().await;
```
elnosh, line 744: this is the only place I called them alone to wait for any tasks and return

simln-lib/Cargo.toml
```diff
@@ -32,6 +32,7 @@ serde_millis = "0.1.1"
 rand_distr = "0.4.3"
 mockall = "0.12.1"
 rand_chacha = "0.3.1"
+tokio-util = { version = "0.7.13", features = ["rt"] }
```
carlaKC, line 35: nittynitnit: try to limit the [commit message's title to 72 chars](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#git-commit-message-structure)

simln-lib/src/lib.rs
```diff
@@ -1395,12 +1396,9 @@ async fn produce_simulation_results(
         }
     };
 
+    set.close();
+    set.wait().await;
```
carlaKC, line 1384: nit: let's leave this log above the `close` + `wait`?

simln-lib/src/lib.rs
```diff
@@ -519,6 +518,8 @@ pub struct Simulation {
     activity: Vec<ActivityDefinition>,
     /// Results logger that holds the simulation statistics.
     results: Arc<Mutex<PaymentResultLogger>>,
+    /// track all tasks spawned for use in the simulation.
+    tasks: TaskTracker,
```
carlaKC, line 522: nitnit: capitalize "track" and make a note that the `run` function will wait for all tasks to complete so that people passing a tracker in that they're using elsewhere know that it'll be handled. 

I think that we can put that both here and on `run` because it's important.

simln-lib/src/lib.rs
carlaKC: right around now is when I start to miss golang's `defer` 😭

simln-lib/src/lib.rs
carlaKC: This is pre-existing from the current design, but WDYT about having no task management in this method at all (except for `producer_tasks` which is internal), and then just wrapping it in a thin method that does the task management for us?

Eg:
```
pub fn run() {
   self.internal_run()
   self.tasks.close()
   self.tasks.wait()
}

fn internal_run() {
    <this method logic>
}
```

simln-lib/src/lib.rs
```diff
@@ -747,18 +740,9 @@ impl Simulation {
             });
         }
 
-        // We always want to wait for all threads to exit, so we wait for all of them to exit and track any errors
```
carlaKC, line 749: I don't want to lose this comment completely, so let's update it to fit the new approach.

Something like:
```We expect our producers and consumers to run until the simulation completes or we hit an error, so we can immediately close our TaskTracker and wait on existing tasks to exit.```

**Review by carlaKC — COMMENTED** [review]
Looking good! 

Only one major comment about how we're going to clean up `run` a bit 👍

File: simln-lib/src/sim_node.rs [hunk 7]
```diff
@@ -691,13 +691,8 @@ impl SimGraph {
...
     pub async fn wait_for_shutdown(&mut self) {
         log::debug!("Waiting for simulated graph to shutdown.");
-
-        while let Some(res) = self.tasks.join_next().await {
-            if let Err(e) = res {
-                log::error!("Graph task exited with error: {e}");
-            }
-        }
-
+        self.tasks.close();
```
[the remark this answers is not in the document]
elnosh: sorry I didn't clean up the commits. I was thinking this could all be squashed into one commit as it all encapsulates the change to `TaskTracker` from `JoinSet` or should I have 3 separate ones as I currently have?

simln-lib/src/lib.rs
elnosh: I'd like it better with everything in `run` but I don't have strong opinions so I can do the `internal_run` if you prefer

carlaKC: > I'd like it better with everything in run

motivation?

My thinking was that it's  _a little_ more future proof to have an `internal_run` just handle task cleanup, so that we don't run the risk of forgetting to handle tasks if we add a new exit point.

simln-lib/src/sim_node.rs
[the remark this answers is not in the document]
carlaKC: > I was thinking this could all be squashed into one commit as it all encapsulates the change to TaskTracker from JoinSet

That sounds good! I was just looking at them individually because they weren't fixups, but that makes sense.

Why don't you go ahead and squash the next time you push, and then we can do fixups if there are any further changes? Just so it's easier to track in review.

simln-lib/src/lib.rs
elnosh: > motivation?

no reason really :sweat_smile: 

> My thinking was that it's _a little_ more future proof to have an `internal_run` just handle task cleanup,

yeah that's why I thought about maybe putting them in the `shutdown` method so that it could handle that cleanup too but `shutdown` isn't always called. Anyways, I'm sold, I'll move the cleanup to an `internal_run` method

elnosh:
I've squashed everything into one commit and tried to address the comments from your last review (:

simln-lib/src/lib.rs
```diff
@@ -518,6 +518,10 @@ pub struct Simulation {
     activity: Vec<ActivityDefinition>,
     /// Results logger that holds the simulation statistics.
     results: Arc<Mutex<PaymentResultLogger>>,
+    /// Track all tasks spawned for use in the simulation.
```
carlaKC, line 521: nit: comments can wrap at 100, here and a few other places

simln-lib/src/sim_node.rs
```diff
@@ -1741,7 +1726,6 @@ mod tests {
         assert_eq!(test_kit.channel_balances().await, expected_balances);
 
         test_kit.shutdown.trigger();
-        test_kit.graph.wait_for_shutdown().await;
```
carlaKC, line 1744: Let's replace this with `close` + `wait` for the `SimGraph`'s `tasks`?

simln-lib/src/sim_node.rs
```diff
@@ -682,24 +682,10 @@ impl SimGraph {
         Ok(SimGraph {
             nodes,
             channels: Arc::new(Mutex::new(channels)),
-            tasks: JoinSet::new(),
+            tasks: TaskTracker::new(),
```
carlaKC, line 685: To be consistent with `Simulation`, let's pass `tasks` in here as well? 

Even though we'll probably use a different constructor to run with a `SimGraph` when we implement this internally, we probably want other use cases that are using `sim-ln` as a library to have to provide (and manage) their own `tasks` - just gives more flexibility.

simln-lib/src/sim_node.rs
```diff
@@ -639,7 +639,7 @@ pub struct SimGraph {
     channels: Arc<Mutex<HashMap<ShortChannelID, SimulatedChannel>>>,
 
     /// track all tasks spawned to process payments in the graph.
-    tasks: JoinSet<()>,
+    tasks: TaskTracker,
```
carlaKC, line 642: If we pass this in, add a comment that `close` / `wait` of these tasks must be externally managed.

simln-lib/src/lib.rs
```diff
@@ -1351,7 +1341,7 @@ async fn produce_simulation_results(
     results: Sender<(Payment, PaymentResult)>,
     listener: Listener,
 ) -> Result<(), SimulationError> {
-    let mut set = tokio::task::JoinSet::new();
+    let set = TaskTracker::new();
```
carlaKC, line 1344: I think that we can pass the top level task set in here as well rather than having a new set?

This is pre-existing, sorry I didn't mention it earlier!

carlaKC (commented):
Last few comments from me, sorry I didn't pick them up on the last review round!

simln-lib/src/sim_node.rs
```diff
@@ -638,8 +638,9 @@ pub struct SimGraph {
     /// channels maps the scid of a channel to its current simulation state.
     channels: Arc<Mutex<HashMap<ShortChannelID, SimulatedChannel>>>,
 
-    /// track all tasks spawned to process payments in the graph.
-    tasks: JoinSet<()>,
+    /// track all tasks spawned to process payments in the graph. Note that handling the shutdown of tasks
```
carlaKC, line 641: nit: capitalize comment

simln-lib/src/lib.rs
```diff
@@ -644,7 +649,19 @@ impl Simulation {
         Ok(())
     }
 
+    /// run until the simulation completes or we hit an error.
+    /// Note that it will wait for the tasks in self.tasks to complete
+    /// before returning.
     pub async fn run(&self) -> Result<(), SimulationError> {
+        self.internal_run().await?;
+        // Close our TaskTracker and wait for any background tasks
+        // spawned during internal_run to complete.
```
carlaKC, line 658: nit: line wrapping on comments

carlaKC (approved):
tACK on 3a94d61d944ebc55a4b0330ae97ef4ab55fc39dc, thanks for picking this one up!

PR #231, opened 2025-03-15
simln-lib/refactor: move parsing into its own module

### LINKED ISSUE #214: Refactor: Move parsing into its own module
**As is**: we have a monolith `main` function that performs all parsing and creates a simulation.

**Motivation**: if using simln as a library, it would be nice to have a parsing module in `sim-cli/src/parsing.rs` to pull in some of the parsing/validation that main currently performs.

Some suggestions for things that we can pull out in a refactor:
- `Cli`: move into module so that it can be re-used externally
- [SimParams](https://github.com/bitcoin-dev-project/sim-ln/blob/8834888f141f6c12d582774fe29233a5a2ea32a4/simln-lib/src/lib.rs#L130): move to parsing module (it doesn't belong in the lib) (also move `NodeConnection` and `ActivityParser`)
- `get_clients`: builds the map of `dyn LightningNode` that'll be passed to the simulation
- `validate_activities`:  performs `ActivityParser` -> `ActivityDefinition` conversion
- `read_sim_path`: this is a nice to have frill, might as well make it accessible for other folks

The only thing that's tricky here is that `validate_activities` needs to make calls to our `dyn LightningNode` to validate that pubkeys exist in their graph, so that needs to be threaded through. Ideally this could just be a generic closure that looks up the node, but it gets ugly when you start dealing with async closures (I briefly tried this in the past and gave up).

f3r10:
fix #214


f3r10:
Hi, @carlaKC I found that async closures are already stabilized on the latest version of rust: https://github.com/rust-lang/rust/pull/132706 so maybe I could update the CI to use version 1.85. WDYT?

btw. most of the changes are based on this previous commit of yours: https://github.com/carlaKC/sim-ln/commit/1951bb816b392f937bf436ccb7de0b3dd402db3d

f3r10:
yes, you are right, @chuksys. `NodeConnection` also has to be moved.
 I am going to close this PR so we can focus on yours  #232

carlaKC:
> Hi, @carlaKC I found that async closures are already stabilized on the latest version of rust: https://github.com/rust-lang/rust/pull/132706 so maybe I could update the CI to use version 1.85. WDYT?

Oh good news! I think let's bump version if we end up using them in the PR that addresses this?

Issue #216, opened 2025-02-04
Feature: Use a single scorer for each sim_node impl of Lightning Node


carlaKC:
**As is**:  The `LightningNode` impl for `SimNode` will [create a new scorer](https://github.com/bitcoin-dev-project/sim-ln/blob/main/simln-lib/src/sim_node.rs#L557) every time it [sends a payment](https://github.com/bitcoin-dev-project/sim-ln/blob/main/simln-lib/src/sim_node.rs#L557).

**To Be**: `scorer` should be stored in `SimNode` and passed to pathfinding for re-use.

**Motivation**: our pathfinding heuristics start from fresh every time we send a payment, which makes pathfinding very dumb (it doesn't remember any of our previous attempts because we just throw them away). To better match reality we should track scores across payments.

While we're here:
* It could be nice to look into the payment parameters used by LDK and perhaps set something other than the default (if it makes sense). I've just never had the time but perhaps we can do something smarter here.

carlaKC:
@jirigrill any updates here? If you're not able to work on this at the moment I'd like to un-assign it so that somebody else can pick it up.

jirigrill:
@carlaKC Sorry, I have been busy with other stuff. You can un-assign it. I dont want block you. I am gonna work on it in slower than needed pace.

sangbida:
Hey, I was going to pick this up, if that's alright :)

sangbida:
I started working on this here: https://github.com/bitcoin-dev-project/sim-ln/pull/234 

Would appreacite any feedback!

carlaKC:
Closed by #234

PR #234, opened 2025-03-19
Add Scorer to SimNode

sangbida:
## Description
This PR modifies the SimNode implementation to store and reuse a single ProbabilisticScorer instance instead of creating a new one for each payment. This contributes to the first part of https://github.com/bitcoin-dev-project/sim-ln/issues/216 

## Changes
- Added ProbabilisticScorer as a field in SimNode struct
- Modified SimNode::new() to initialize scorer
- Updated find_payment_route to accept the stored scorer reference
- Updated `DispatchPaymentTestKit` to handle Arc'd NetworkGraph and scorer reuse


carlaKC (approved):
utACK, nice concise change! 

Difficult to test this without the ability to run with `SimNode` in `sim-ln` (SOON ™️ ) so I think we can just land this as is once @f3r10 has had a look.

f3r10 (commented):
utACK

Issue #185, opened 2024-04-26
Robustness: Improve Test Coverage of Repo


carlaKC:
SimLN started as a hackathon project, so it doesn't have much by way of unit tests. This is very, very bad. 

We should aim to dramatically improve the coverage of the codebase to ensure that everything is working as expected, and there are no future regressions. 

A great first starting point would be to add test coverage for `Simulator` with [mock out](https://docs.rs/mockall/latest/mockall/) `LightningNode` implementations. See [sim node mocking](https://github.com/bitcoin-dev-project/sim-ln/blob/main/sim-lib/src/sim_node.rs#L1415) as an example of how to go about this. 

Low hanging fruit:
* `validate_activity`
* `validate_node_network`

Middle tier:
Test individual functions for their expected behavior, eg: 
* Spin up `consume_events` 
* Test various exit conditions - `send_payment` failure, `listener` shutdown etc
* Test successful `send_payment` for both `SimulationOutput::SendPaymentSuccess`/`SimulationOutput::SendPaymentFailure` etc

A few considerations here: 
* Will need to spin up blocking functions like `consume_events` in a separate task so the test can run
* Consider buffering channels to make life a bit easier

The rest of the elephant: 
* Test the high level `Simulation::run` function with mocked nodes
* Refactors that make life easier here are welcome!

<p align="center">
  <img src="https://github.com/bitcoin-dev-project/sim-ln/assets/42311294/0eba5982-8e69-4ced-91dd-db028462eb00"/>
</p>

joseguru:
Hey @carlaKC. Following you recommendation #180   to write tests on validate_activity (currently working on it), Can I pick this up an work on the low hanging fruits

Godwin-88:
Hi , @carlaKC if this is still available , I'd like to take this on.

carlaKC:
I'm not currently available for review here, but more tests are always better.

PR #236, opened 2025-03-25
Unit tests for validate_activity and validate_node_network

jirigrill:
Based on this [issue](https://github.com/bitcoin-dev-project/sim-ln/issues/185), unit tests for methods `validate_activity` and `validate_node_network` in `simln-lib/src/lib.rs` were created.

### DIFF
--- a/simln-lib/Cargo.toml
+++ b/simln-lib/Cargo.toml
@@ -37,3 +37,4 @@ tokio-util = { version = "0.7.13", features = ["rt"] }
 
 [dev-dependencies]
 ntest = "0.9.0"
+mockall = "0.13.1"
--- a/simln-lib/src/lib.rs
+++ b/simln-lib/src/lib.rs
@@ -1460,6 +1460,7 @@ mod tests {
         PaymentGenerationError, PaymentGenerator, Simulation,
     };
     use async_trait::async_trait;
+    use bitcoin::Network;
     use bitcoin::secp256k1::PublicKey;
     use mockall::mock;
     use std::collections::HashMap;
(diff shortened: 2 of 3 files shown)


chuksys:
Running `make check` locally will help in detecting formatting and linting issues before pushing.

jirigrill:
@chuksys thx for pointing out. I have made updates to Makefile to add it into `make help` and added `make fix-all` to update linting issues

chuksys (changes requested):

Makefile
```diff
@@ -10,27 +10,33 @@ mount-volume:
...
+	@echo "  mount-volume       Mounts the specified volume."
+	@echo "  run                Runs the Docker container in detached mode."
+	@echo "  run-interactive    Runs the Docker container in interactive mode."
+	@echo "  stop               Stops the Docker container."
+	@echo "  check              Runs code formatting and linting checks."
+	@echo "  check-code         Runs code formatting and linting without stability check."
+	@echo "  fix-all            Fixes both formatting and linting issues in one go."
+	@echo "  stable-output      Checks for unstaged/uncommitted changes after checks."
+	@echo "  install            Installs the simulation CLI."
+	@echo "  dev-install        Installs the simulation CLI with development features."
```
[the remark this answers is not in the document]
f3r10: I agree; it would be better to include this on a new PR

sim-cli/src/parsing.rs
```diff
@@ -159,14 +159,17 @@ pub async fn create_simulation(cli: &Cli) -> Result<Simulation, anyhow::Error> {
...
+        let clients_clone = clients.clone(); // Clone to avoid moving
+        let pk_owned = *pk;
+        Box::pin(async move {
+            if let Some(c) = clients_clone.values().next() {
+                return c.lock().await.get_node_info(&pk_owned).await;
+            }
+            Err(LightningError::GetNodeInfoError(
+                "no nodes for query".to_string(),
+            ))
+        })
```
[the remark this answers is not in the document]
f3r10: I agree. Let's do this by opening a new issue and discussing why this change is necessary.

f3r10 (commented):
Tested ACK

Makefile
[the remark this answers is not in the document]
jirigrill: I will move it to new PR

jirigrill:
@f3r10 I will use the commit structure once the PR is approved and I will do squash & merge. In next PR I will use this structure from beginning.

jirigrill:
Changes in Makefile were moved to this PR: https://github.com/bitcoin-dev-project/sim-ln/pull/239

jirigrill:
Issue regarding clippy error in `sim-cli/src/parsing.rs/get_node` method filed here: https://github.com/bitcoin-dev-project/sim-ln/issues/240

chuksys (commented):
Just one minor observation and I think this is good if @f3r10 approves.

carlaKC:
> @f3r10 I will use the commit structure once the PR is approved and I will do squash & merge. In next PR I will use this structure from beginning.

We don't squash commits on merge so the commit structure will need to be fixed up in order to be approved - for example, there's a merge commit in this PR that needs to go. Our goal is to have atomic, meaningful commits that pass CI so that we can use `git bisect`. 

It would be great if you could go ahead and squash some of the "revert .." commits down so that we can get an idea of what the commits will look like while we review 🙏  In future using `--fixup` commits will be helpful, as GH will naturally group them together which makes review much easier.

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +53,140 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
         })
         .collect()
 }
+
+// Mock implementations for tests
+#[cfg(test)]
```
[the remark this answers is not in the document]
carlaKC: We're currently not telling the compiler that this is a module that has entirely test materiel in it, so this is the right move IMO. 

I think the strictly-most-correct thing to do would either be to move this into a `tests/` dir or inline the module and have a tests config tag on it. All this is pre-existing so happy to leave this as-is and fix holistically in another PR.

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +53,140 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
         })
         .collect()
 }
+
+// Mock implementations for tests
```
carlaKC, line 57: nit: comment doesn't add that much, can delete it

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +53,140 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
...
+
+/// Type alias for the result of setup_test_nodes
+type TestNodesResult = (
+    Vec<NodeInfo>,
+    HashMap<PublicKey, Arc<Mutex<dyn LightningNode>>>,
+);
+
+/// Creates a set of nodes with mock implementations
+#[cfg(test)]
+pub fn setup_test_nodes(node_count: usize, keysend_indices: &[usize]) -> TestNodesResult {
```
carlaKC, line 104: This is pretty fussy, so feel free to ignore.

I'm often peeved by test helper functions that have parameters for a very specific thing because they're inflexible - you often end up creating multiple helpers or adding tons of parameters that aren't necessary. I feel like we're hitting this with `setup_test_nodes` and `setup_network_test_nodes`?

Perhaps we could use functional options or a [builder](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) pattern here to have one single network creation util that's flexible enough to serve different use cases?

simln-lib/src/lib.rs
```diff
@@ -1548,33 +1526,184 @@ mod tests {
         );
     }
 
+    /// Tests for validate_activity()
+    #[tokio::test]
```
carlaKC, line 1530: These all look great!

Would like to distill the in-function comments to just one high-level comment on the function that explains what's being covered. I think the tests are readable enough (thanks to nice helper fns) to not require line-by-line comments.

carlaKC (commented):
a test pr is the best pr 🥇 

some comments on the helper functions, while we're here I'd like to get them perfect + a few code style nitpicks

simln-lib/src/test_utils.rs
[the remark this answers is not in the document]
chuksys: What if we add `#![cfg(test)]` at the very top of the file?

simln-lib/Cargo.toml
```diff
@@ -30,10 +30,11 @@ hex = "0.4.3"
 csv = "1.2.2"
 serde_millis = "0.1.1"
 rand_distr = "0.4.3"
-mockall = "0.12.1"
+mockall = "0.13.1"
```
chuksys, line 33: I think it's a good idea adding this to the dev-deps; do we still need it here though?

simln-lib/src/test_utils.rs
```diff
@@ -1,10 +1,19 @@
+#[cfg(test)]
```
chuksys, line 1: I think this should actually be `#![cfg(test)]`.

chuksys (commented):
Thanks for the updates! Just a few observations I thought might be worth considering:

simln-lib/Cargo.toml
jirigrill: You are right, that it does not make sense to have in `[dependencies]` as it is related only for tests, thus it should be only in `[dev-dependencies]`. My bad.

simln-lib/src/test_utils.rs
jirigrill: Good catch, thank you

simln-lib/src/lib.rs
```diff
@@ -1417,19 +1416,19 @@ async fn track_payment_result(
 
 #[cfg(test)]
 mod tests {
+    // use crate::test_utils::TestNetworkBuilder;
```
carlaKC, line 1419: dependency commented out?

simln-lib/src/lib.rs
```diff
@@ -1548,33 +1526,178 @@ mod tests {
         );
     }
 
+    // Tests for validate_activity()
```
carlaKC, line 1529: I don't think that these top level comments add much, let's remove them

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +54,187 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
...
+}
+
+impl LightningTestNodeBuilder {
+    /// Creates test nodes with optional keysend support on specified nodes.
+    /// A convenience method equivalent to `new(node_count).with_keysend_nodes(indices).build_full()`.
+    pub fn setup_test_nodes(node_count: usize, keysend_indices: &[usize]) -> TestNodesResult {
+        Self::new(node_count)
+            .with_keysend_nodes(keysend_indices.to_vec())
+            .build_full()
+    }
```
carlaKC, line 133: This somewhat defeats the point of a generic builder if we're going to add constructors for each custom case?

Let's just use `new().with_keysend_nodes(..)`

Likewise below with `setup_network_test_nodes`

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +54,187 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
...
+        networks: Vec<Network>,
+    ) -> HashMap<PublicKey, Arc<Mutex<dyn LightningNode>>> {
+        Self::new(node_count)
+            .with_networks(networks)
+            .build_clients_only()
+    }
+
+    /// Creates a new builder instance with the specified number of nodes.
+    /// The default configuration includes a balance of 100,000 units per node,
+    /// no keysend support, and no specific networks.
```
carlaKC, line 148: The default should be with keysend support, since that's required for regular operation

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +54,187 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
...
+
+    /// Specifies which nodes should have the keysend feature enabled
+    pub fn with_keysend_nodes(mut self, indices: Vec<usize>) -> Self {
+        self.keysend_indices = indices;
+        self
+    }
+
+    /// Sets specific networks for each node
+    /// Will panic if the number of networks doesn't match node_count
+    /// Returns self for method chaining
```
carlaKC, line 166: nit: here and a few other places - wrap at 120 and end comments with full stops.

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +54,187 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
...
+    pub fn with_networks(mut self, networks: Vec<Network>) -> Self {
+        // Validate that we have the correct number of networks
+        if networks.len() != self.node_count {
+            panic!("Must specify a network for each node");
+        }
```
carlaKC, line 171: Can use `assert!`

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +54,187 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
...
+        (node_infos, clients)
+    }
+}
+
+pub fn create_simulation(
+    clients: HashMap<PublicKey, Arc<Mutex<dyn LightningNode>>>,
+    activity: Vec<ActivityDefinition>,
+) -> Simulation {
+    Simulation::new(
+        SimulationCfg::new(Some(0), 0, 0.0, None, None),
```
carlaKC, line 220: Add a comment that this sets a runtime for the simulation of 0, so `run` will exit immediately - this probably isn't desirable for other tests. 

Happy to leave as-is for now, if we need it we can make it a parameter or add a builder.

simln-lib/src/test_utils.rs
```diff
@@ -1,4 +1,4 @@
-#[cfg(test)]
+#![cfg(test)]
```
carlaKC, line 1: This commit should be fixups for the two that come before it rather than its own commit

simln-lib/src/lib.rs
jirigrill: This was leftover. Will remove it.

simln-lib/src/lib.rs
jirigrill: I left it there on purpose to distinguish groups of test methods related to specific method. But if you think its not valuable, I will remove it.

simln-lib/src/test_utils.rs
jirigrill: Understood, makes sense, I was lazy and kept it workable with previous version. I will update it.

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +54,187 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
...
+
+/// A builder for creating mock Lightning nodes for testing purposes.
+///
+/// This struct provides a flexible way to configure test nodes with features like
+/// keysend support and specific networks. It supports both full node setups (with
+/// node info and clients) and client-only setups for network testing.
+///
+/// # Examples
+/// Basic setup with keysend on specific nodes:
+/// let (nodes, clients) = TestNetworkBuilder::setup_test_nodes(5, &[0, 2]);
```
f3r10, line 108: I think it is necessary to update the `TestNetworkBuilder` reference for `LightningTestNodeBuilder`

simln-lib/src/test_utils.rs
jirigrill: As I am very new to rust, why using assert is better than if + panic?

simln-lib/src/test_utils.rs
jirigrill: As I am using fixup for the first time, now it looks as expected? AFAIK it should be squashed once it is the PR ready.

simln-lib/src/test_utils.rs
carlaKC: It's just shorthand / typical to use in tests. It nicely prints out the left / right values which makes it nice to debug if tests are failing.

simln-lib/Cargo.toml
```diff
@@ -30,7 +30,6 @@ hex = "0.4.3"
 csv = "1.2.2"
 serde_millis = "0.1.1"
 rand_distr = "0.4.3"
-mockall = "0.13.1"
```
carlaKC: This should be a `--fixup a45993d3c8a8a7dff680a62216175723db389b5f`

simln-lib/src/test_utils.rs
carlaKC: Almost there. 

The change to dependencies should fixup the first commit and the rest should fixup the second - right now you've got both of those changes being applied to the second commit in a fixup.

Feel free to remove the fixup and redo it in two separate ones - fixing up a fixup gets a bit messy

simln-lib/src/test_utils.rs
jirigrill: Thanks for your explanation, @carlaKC

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +54,169 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
...
+/// A builder for creating mock Lightning nodes for testing purposes.
+///
+/// This struct provides a flexible way to configure test nodes with features like
+/// keysend support and specific networks. It supports both full node setups (with
+/// node info and clients) and client-only setups for network testing.
+///
+/// # Examples
+///
+/// Using the builder for more control:
+/// let (nodes, clients) = LightningTestNodeBuilder::new(5)
```
carlaKC, line 109: nit: shouldn't this be 3?

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +54,169 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
...
+///
+/// Building clients with specific networks:
+/// let clients = LightningTestNodeBuilder::new(3)
+///     .with_networks(vec![Network::Bitcoin, Network::Testnet, Network::Regtest])
+///     .build_clients_only();
+pub struct LightningTestNodeBuilder {
+    node_count: usize,              // Required - must be provided at creation.
+    initial_balance: u64,           // Always has a value (default: 100,000).
+    keysend_indices: Vec<usize>,    // Always a vector (default: Vec filled with 0..node_count).
+    networks: Option<Vec<Network>>, // Can be None (default) or Some(networks).
```
carlaKC, line 121: I don't think all these comments are necessary (node_count and networks are obvious IMO, looks too chatpgt-ey), and they should be above the field rather than inline when they are.

simln-lib/src/test_utils.rs
```diff
@@ -45,3 +54,169 @@ pub fn create_nodes(n: usize, node_capacity: u64) -> Vec<(NodeInfo, u64)> {
...
+impl LightningTestNodeBuilder {
+    /// Creates a new builder instance with the specified number of nodes.
+    /// The default configuration includes a balance of 100,000 units per node,
+    /// keysend support ON, and Regtest network.
+    pub fn new(node_count: usize) -> Self {
+        Self {
+            node_count,
+            initial_balance: 100_000,                   // Default 100k sats.
+            keysend_indices: (0..node_count).collect(), // Keysend for all nodes ON by default.
+            networks: Some(vec![Network::Regtest; node_count]), // Regtest network for all nodes by default.
```
carlaKC, line 133: Likewise here, comments are redundant

simln-lib/src/lib.rs
```diff
@@ -1548,33 +1525,178 @@ mod tests {
...
+        let simulation = test_utils::create_simulation(clients, vec![]);
+        let result = simulation.validate_node_network().await;
+
+        assert!(result.is_err());
+        assert!(matches!(result, 
+            Err(LightningError::ValidationError(msg)) if msg.contains("mainnet is not supported")));
+    }
+
+    /// Verifies that nodes on Testnet and Regtest fail validation, expecting a
+    /// `ValidationError` with "nodes are not on the same network".
```
carlaKC, line 1663: nit: linewrap looks off

carlaKC (approved):
Still got quite a few style comments, but they're only on comments in tests so going to go ahead with merge. Would be happy to quickly merge a follow up that fixes them up.

Thanks for the effort on the builder, looks great in the end.

PR #238, opened 2025-03-26
sim-cli/refactor: do not call add_node_to_maps in validate_activities

### LINKED ISSUE #237: Refactor: Do not call add_node_to_maps in validate_activities
Create `ActivityValidationParams` struct and pass to `validate_activities` to improve testability and reduce coupling.

              Pass these in rather than calling `add_node_to_maps` here

_Originally posted by @carlaKC in https://github.com/bitcoin-dev-project/sim-ln/pull/232#discussion_r2012566835_

chuksys:
Closes #237

### DIFF
--- a/sim-cli/src/parsing.rs
+++ b/sim-cli/src/parsing.rs
@@ -169,7 +169,10 @@ pub async fn create_simulation(cli: &Cli) -> Result<Simulation, anyhow::Error> {
         ))
     };
 
-    let validated_activities = validate_activities(activity, &clients_info, get_node).await?;
+    let (pk_node_map, alias_node_map) = add_node_to_maps(&clients_info).await?;
+
+    let validated_activities =
+        validate_activities(activity, pk_node_map, alias_node_map, get_node).await?;
     let tasks = TaskTracker::new();
 
     Ok(Simulation::new(cfg, clients, validated_activities, tasks))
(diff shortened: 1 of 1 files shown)


chuksys:
Hi @f3r10 if you have the bandwidth, please can you review this little change in how params are passed into `validate_activities`?

f3r10 (approved):
Tested ACK

sim-cli/src/parsing.rs
```diff
@@ -169,7 +175,15 @@ pub async fn create_simulation(cli: &Cli) -> Result<Simulation, anyhow::Error> {
         ))
     };
 
-    let validated_activities = validate_activities(activity, &clients_info, get_node).await?;
+    let (pk_node_map, alias_node_map) = add_node_to_maps(&clients_info).await?;
+
+    let activity_validation_params = ActivityValidationParams {
```
carlaKC, line 180: I don't feel like constructing this here only to deconstruct it in `validate_activities` adds much to readability - let's just pass the 3 parameters in?

chuksys: The rationale behind my approach is that 4 parameters (the 3 in the struct + get_node_info) will be a bit too much for a high-level understanding of what the function does (especially for someone looking at the code for the first time) and that was why I grouped the related parameters into the struct, to also allow for extension of the function in future, if need be. We could still pass the 4 args in for now and maybe if we need to extend the function in future, we could consider the approach of using a struct?

carlaKC: > The rationale behind my approach is that 4 parameters (the 3 in the struct + get_node_info) will be a bit too much for a high-level understanding of what the function does (especially for someone looking at the code for the first time) and that was why I grouped the related parameters into the struct,

I don't really see how moving the parameters into a struct helps with understanding, especially when it doesn't have comments. It's also a little misleading to me that the `ActivityValidationParams` contains the actual `activity` to be validated - it reads to me that the struct contains some parameters that will define how we validate, rather than the things being validated themselves.

> We could still pass the 4 args in for now and maybe if we need to extend the function in future, we could consider the approach of using a struct?

This seems reasonable to me - I also don't think we're likely to change this struct much in the future (famous last words 😅 ). 

Let's go with passing the arguments and adding a comment explaining what the function does if we're worried that it isn't readable.

chuksys: > Let's go with passing the arguments and adding a comment explaining what the function does if we're worried that it isn't readable.

Okay, I'll make the changes asap!

> I also don't think we're likely to change this struct much in the future (famous last words 😅 )

😂 I'll try to hold you to that!

carlaKC (approved):
lgtm

carlaKC:
You can go ahead and squash the fixups then we'll merge 👍

PR #239, opened 2025-03-28
sim-ln/refactor: Update help, add new cmds in Makefile

jirigrill:


Makefile
```diff
@@ -50,5 +56,9 @@ check: check-code stable-output
 install:
 	cargo install --locked --path sim-cli
 
+fix-all:
```
carlaKC, line 59: `format` is probably a better name for this. Likewise would like a `CLIPPY_CMD` to de-duplicate.

jirigrill:
@carlaKC thank you for your feedback. I have reset the commits to follow your suggestion. Let me know if now it looks you expected.

Issue #215, opened 2025-01-27
Feature: Add ability to run with mocked lightning network


carlaKC:
Depends on #214 

*As is*: We have a fully "mocked" implementation of `LightningNode` in [sim_node](https://github.com/bitcoin-dev-project/sim-ln/blob/main/simln-lib/src/sim_node.rs) that just keeps track of channel state in memory to create very lightweight simulations, but it is not currently surfaced in `sim-cli` so there is no way for people to easily use it.

*Motivation*: This is a great feature to have for larger networks, so we should surface a way for people to use it.

*To Be*:
We need to update our [sim params](https://github.com/bitcoin-dev-project/sim-ln/blob/8834888f141f6c12d582774fe29233a5a2ea32a4/simln-lib/src/lib.rs#L131) to add a new `sim_network` that specifies the channel policies of the network that we want simulated.  This option is mutually exclusive with `nodes`, because we can't run real and fake nodes together. 

I have a [mega-branch](https://github.com/carlaKC/sim-ln/commits/interceptor-latency/) that took a shot at this but I don't like the approach I took (and the branch is a mess). [This](https://github.com/carlaKC/sim-ln/commit/d1afc33d95e473047c7b668fa7b424d10f4bebb5) commit is useful because it provides the json structure that I think that we should use.

Option to run sim-ln with then be:
* `nodes` + `activity`: run defined activity on real nodes
* `nodes`: run random activity on real nodes
* `sim_network` + `activity`: run defined activity on fake nodes
* `sim_network`: run random activity on fake nodes

(all else is invalid)

chuksys:
I would like to get familiar with sim-ln by working on this.

carlaKC:
> I would like to get familiar with sim-ln by working on this.

I don't think that this is a great first issue to pick up, it's going to need a fair amount of refactoring that'll require familiarity with the project.

chuksys:
Ok. I'll pick another issue then come back to this when I'm more familiar with the project.

carlaKC:
A nice follow up for this would be to add a script like [this](https://github.com/carlaKC/attackathon/blob/master/setup/lnd_to_simln.py) to a `utils/` directory so that people can easily convert output from LND's graph to run with sim-ln. 

Safe for other impls!

carlaKC:
Closed in #248, opened #266 to track adding helper scripts for simulation mode.

PR #242, opened 2025-03-28
sim-lib: add simulated clock to speed up simulations

carlaKC:
This PR adds a `Clock` trait and makes `Simulation` generic over this trait so that we have the ability to mock time, as described in #81. The default implementation of this trait just thinly wraps a system clock.

For the version where we want to speed things up, we track the instant that we started our clock and apply a multiplier to speed up time passed. For example, if we're running with a multiplier of 10:
* Track our start instant as zero
* One second of "real" time passes, which is equivalent to 10 seconds of simulated time

When dealing with durations, we can simply multiply/divide to move between simulation and real time periods.
When dealing with timestamps, we use the elapsed duration since start to figure out a sped up clock time.

This PR doesn't surface the implementation anywhere, because it doesn't _really_ make sense to run this speed up on real nodes - things get messy when you're using different clocks; if you want things faster on your real node network, just set the activity to fire more often or the random activity to send more payments. 

It is, however, very useful once we've implemented #215, because we can speed up our sim-node network.

### DIFF
--- a/sim-cli/src/parsing.rs
+++ b/sim-cli/src/parsing.rs
@@ -3,6 +3,7 @@ use bitcoin::secp256k1::PublicKey;
 use clap::{builder::TypedValueParser, Parser};
 use log::LevelFilter;
 use serde::{Deserialize, Serialize};
+use simln_lib::clock::SimulationClock;
 use simln_lib::sim_node::{
     ln_node_from_graph, populate_network_graph, ChannelPolicy, SimGraph, SimulatedChannel,
 };
@@ -198,7 +199,7 @@ pub async fn create_simulation_with_network(
     cli: &Cli,
     sim_params: &SimParams,
     tasks: TaskTracker,
-) -> Result<(Simulation, Vec<ActivityDefinition>), anyhow::Error> {
+) -> Result<(Simulation<SimulationClock>, Vec<ActivityDefinition>), anyhow::Error> {
     let cfg: SimulationCfg = SimulationCfg::try_from(cli)?;
     let SimParams {
         nodes: _,
@@ -241,7 +242,14 @@ pub async fn create_simulation_with_network(
         get_validated_activities(&nodes, nodes_info, sim_params.activity.clone()).await?;
 
     Ok((
-        Simulation::new(cfg, nodes, tasks, shutdown_trigger, shutdown_listener),
+        Simulation::new(
+            cfg,
+            nodes,
+            tasks,
+            Arc::new(SimulationClock::new(1)?),
+            shutdown_trigger,
+            shutdown_listener,
+        ),
         validated_activities,
     ))
 }
@@ -252,7 +260,7 @@ pub async fn create_simulation(
     cli: &Cli,
     sim_params: &SimParams,
     tasks: TaskTracker,
-) -> Result<(Simulation, Vec<ActivityDefinition>), anyhow::Error> {
+) -> Result<(Simulation<SimulationClock>, Vec<ActivityDefinition>), anyhow::Error> {
     let cfg: SimulationCfg = SimulationCfg::try_from(cli)?;
     let SimParams {
         nodes,
@@ -267,7 +275,14 @@ pub async fn create_simulation(
         get_validated_activities(&clients, clients_info, sim_params.activity.clone()).await?;
 
     Ok((
-        Simulation::new(cfg, clients, tasks, shutdown_trigger, shutdown_listener),
+        Simulation::new(
+            cfg,
+            clients,
+            tasks,
+            Arc::new(SimulationClock::new(1)?),
+            shutdown_trigger,
+            shutdown_listener,
+        ),
         validated_activities,
     ))
 }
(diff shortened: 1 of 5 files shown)


carlaKC:
Opening this up for conceptual review, cc @sangbida @f3r10 @elnosh. I'll do a cleanup pass on it before it requires thorough review.

It's the simplest thing that I could come up with (and has worked pretty well when I run simulations using it on a branch), but I'm very open to other opinions!

<details>
  <summary></summary>
  
time

</details>

carlaKC:
Note to self - will update in the next push: 

Right now, SimLN is calling `now` for [every channel policy timestamp](https://github.com/bitcoin-dev-project/sim-ln/blob/main/simln-lib/src/sim_node.rs#L772) that we add to our simulated network. If the clock is sped up fast enough, the timestamp for these channel updates will keep increasing as we add them. LDK only allows updates to be 24H in the future, so it's possible that we can hit an error here if we have a large enough graph and an fast enough simulation clock - I've hit this a few times in jamming simulations.

sangbida:
This looks good to me:) 

I had a few questions about how this would be used in the wild. If the purpose is to only speed up transactions, I think this works great! I was wondering if you may run into issues with clock skew because of how `tokio:sleep` works? This maybe an issue if you wanted to track precise timestamps.

carlaKC:
> I had a question about how this would be used in the wild. If the purpose is to only speed up transactions, I think this works great!

The plan would be to use this alongside #248 to run sped-up transactions on a simulated network. When it comes time to surface this option, we'll only make it available if running with a fake network - trying to speed up transactions on real nodes that are running with regular clock time would indeed be a mess (if that's what you want, just setup your simulation to have more transactions).

> I was wondering if you may run into issues with clock skew because of how tokio:sleep works?

I think that provided we always rely on the simulation clock we should be ok? Eg with a 10x speed up, if we have a sleep for 10 minutes we should just tell `tokio::sleep` to sleep for 1 minute. Maybe I'm missing your point here?

f3r10:
LGTM!!
Looking forward to seeing how it works together with #248

simln-lib/src/lib.rs
```diff
@@ -511,12 +514,13 @@ struct ExecutorKit {
     payment_generator: Box<dyn PaymentGenerator>,
 }
 
-impl Simulation {
+impl <C: Clock + 'static> Simulation <C>{
```
carlaKC, line 517: TODO(carla): can this be replaced with a lifetime?

carlaKC:
Force pushed to:
- Make clock generic rather than using dynamic dispatch (not necessary when we'll just be running with one clock or the other)
- Add a commit to address channel updates issue: https://github.com/bitcoin-dev-project/sim-ln/pull/242#issuecomment-2766176645

Will start with fixups once this is in real review (needs to be rebased on #248 first, waiting till that approach has stabilized).

elnosh:
approach looks good to me (: ran some simulations using `SimulationClock` and seems to be working fine.

sangbida:
I think I misunderstood how `SimulationClock ` is working here. So the case I was worried about was, let's say you have a 100x multiplier, if the tokio::sleep wakes up 5ms late then you have a 500ms discrepancy in the simulation time and this could potentially get compouned over time. But on rereading, the simulation clock isn't actually keeping track of the time it's just using the system clock, so we won't run into the issue with compounding but we may run into issues with the scheduler delay, which I think is okay since the intended use is within a mocked ln network.

carlaKC:
Given that this is being used quite extensively in simulations and working well, I think that we can go ahead with review even though there isn't a way to test this and then base #248 on top of it (instead of the other way around)? Adding it just requires a cli flag and some validation so it shouldn't be much of a lift.

simln-lib/src/clock.rs
```diff
@@ -0,0 +1,136 @@
...
+    async fn sleep(&self, wait: Duration) {
+        time::sleep(wait).await;
+    }
+}
+
+/// Provides an implementation of the Clock trait that speeds up wall clock time by some factor.
+#[derive(Clone)]
+pub struct SimulationClock {
+    // The multiplier that the regular wall clock is sped up by.
+    speedup_multiplier: u32,
```
elnosh, line 33: since this can only be in the [1..1000] range, maybe it could be a `u16` instead?

elnosh (commented):
LGTM

Still need to build more context on the code so I only have some minor comments/questions.

sim-cli/src/parsing.rs
```diff
@@ -79,27 +84,82 @@ pub struct Cli {
     /// Seed to run random activity generator deterministically
     #[clap(long, short)]
     pub fix_seed: Option<u64>,
+    /// A multiplier to wall time to speed up the simulation's clock. Only available when when running on a network of
+    /// simulated nodes.
+    #[clap(long, short)]
+    pub speedup_clock: Option<u16>,
```
f3r10, line 90: There is an issue with this argument 
```bash
Command sim-cli: Short option names must be unique for each argument, but '-s' is in use by both 'sim_file' and 'speedup_clock'
```

carlaKC:
Just squashed and rebased, no [chages](https://github.com/bitcoin-dev-project/sim-ln/compare/5b3671d0768d4b7c290caff447e57c59db39a3fb..1c91815400e92bfdf81a812da50eeb50bc0861f5) in last push.

f3r10 (approved):
Tested ACK

PR #247, opened 2025-03-30
sim-ln/enhance: Add rust-toolchain.toml

jirigrill:
To ensure that everyone is working on same `rustc version` and has installed `rustfmt` and `clippy`


carlaKC (approved):
tACK, conveniently happened that I needed to update my rust version to 1.85 so could see this in all it's glory.

Thanks for the nice quality of life improvement!

Issue #246, opened 2025-03-29
Bug: make docker-build is failing


jirigrill:
**Describe the bug**

I am trying to run `make docker-build` on MacOS on Sonoma 14.5 and getting this error:
```
Dockerfile:32
--------------------
  30 |     RUN rustup target add ${TARGET_RUST_ARCH}
  31 |     RUN rustup component add rustfmt
  32 | >>> RUN cd /sim-ln && RUSTFLAGS='-C target-feature=+crt-static' cargo build --locked --release --target=${TARGET_RUST_ARCH}
  33 |     RUN mv /sim-ln/target/${TARGET_RUST_ARCH}/release/sim-cli /sim-ln/sim-cli
  34 |
--------------------
ERROR: failed to solve: process "/bin/sh -c cd /sim-ln && RUSTFLAGS='-C target-feature=+crt-static' cargo build --locked --release --target=${TARGET_RUST_ARCH}" did not complete successfully: exit code: 101

View build details: docker-desktop://dashboard/build/desktop-linux/desktop-linux/jkoycdlajswpjykz45eut4vxp
make: *** [build-docker] Error 1
```

**To Reproduce**

`make docker-build`

**Config**

MacOS on Sonoma 14.5

jirigrill:
I am trying to resolve the bug meantime.

carlaKC:
Also hitting that on 2b30a3c with docker 24.0.2 / macOS Sequoia 15.3.

Did you also get this openssl error a bit higher up in the build output?
```
12.61 warning: openssl-sys@0.9.106: Could not find directory of OpenSSL installation, and this `-sys` crate cannot proceed without this knowledge. If OpenSSL is installed and this crate had trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the compilation process. See stderr section below for further information.
12.61 error: failed to run custom build command for `openssl-sys v0.9.106`
12.61
12.61 Caused by:
12.61   process didn't exit successfully: `/sim-ln/target/release/build/openssl-sys-f0000663456d59de/build-script-main` (exit status: 101)
12.61   --- stdout
```

jirigrill:
I have prepared bugfix in this PR: https://github.com/bitcoin-dev-project/sim-ln/pull/249

carlaKC:
Closed in #249

PR #249, opened 2025-04-01
sim-ln/bugfix: Set OpenSSL to be built from source 

jirigrill:
As reported [here](https://github.com/bitcoin-dev-project/sim-ln/issues/246), building docker image on MacOS Sonoma 14.5 is failing. 

The solution resoling this error is to build OpenSSL from source during the Rust compilation process, rather than relying on a pre-installed OpenSSL library on the system.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,10 +5,10 @@ FROM rust:latest AS base
 # buildkit will provide this automatically, no need to pass it in
 ARG TARGETARCH
 
-FROM base as builder-amd64
+FROM base AS builder-amd64
 ENV TARGET_RUST_ARCH="x86_64-unknown-linux-musl"
 
-FROM base as builder-arm64
+FROM base AS builder-arm64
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -18,14 +18,22 @@ ENV CC_aarch64_unknown_linux_musl=clang
 ENV AR_aarch64_unknown_linux_musl=llvm-ar
 ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_RUSTFLAGS="-Clink-self-contained=yes -Clinker=rust-lld"
 
-FROM builder-${TARGETARCH} as builder
-RUN echo "Building for architecture:  $TARGET_RUST_ARCH"
+FROM builder-${TARGETARCH} AS builder
```
carlaKC, line 21: nit: generally prefer to have formatting fixes in their own commit

That way, if we ever do have a bug that we need to bisect there's less code in the guilty commit to go through.

carlaKC (approved):
tACK! thanks for the fix

PR #258, opened 2025-04-29
github: bump actions to v4

carlaKC:
Tag of 2.4 failed to build docker images: https://github.com/bitcoin-dev-project/sim-ln/actions/runs/14735190441/job/41375486629

Seems like an upgrade will fix this: https://github.com/orgs/community/discussions/152695

While we're here, I've bumped a whole lot of versions to their latest (w/ help from chatgpt).


carlaKC:
We don't have a great way to test the docker tag flow without mangling these workflows so I'm just going to mergenpray.

PR #259, opened 2025-04-29
 Implement Distinct RNG Sequences 

sangbida:
## Overview
This PR introduces a deterministic yet unique random number generation system for the Lightning Network simulation. Each `NetworkGraphView` now maintains a base seed, while individual nodes receive their own RNG instances derived from this base seed and salted with their public keys. This ensures consistent but varied behavior across the network while preserving reproducibility.

## Changes
- Added seed field to `MutRng` struct
- Implemented salted method that creates node-specific RNGs by combining the original seed with the node's public key
- Updated `random_activity_nodes` to use salted RNGs for each node

### DIFF
--- a/simln-lib/src/lib.rs
+++ b/simln-lib/src/lib.rs
@@ -7,13 +7,13 @@ use bitcoin::Network;
 use csv::WriterBuilder;
 use lightning::ln::features::NodeFeatures;
 use lightning::ln::PaymentHash;
-use rand::rngs::StdRng;
 use rand::{Rng, RngCore, SeedableRng};
 use rand_chacha::ChaCha8Rng;
 use random_activity::RandomActivityError;
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
 use std::fmt::{Display, Formatter};
+use std::hash::{DefaultHasher, Hash, Hasher};
 use std::marker::Send;
 use std::path::PathBuf;
 use std::sync::Mutex as StdMutex;
@@ -527,16 +527,24 @@ type MutRngType = Arc<StdMutex<dyn RngCore + Send>>;
 struct MutRng(MutRngType);
 
 impl MutRng {
-    /// Creates a new MutRng given an optional `u64` argument. If `seed_opt` is `Some`,
-    /// random activity generation in the simulator occurs near-deterministically.
-    /// If it is `None`, activity generation is truly random, and based on a
-    /// non-deterministic source of entropy.
-    pub fn new(seed_opt: Option<u64>) -> Self {
-        if let Some(seed) = seed_opt {
-            Self(Arc::new(StdMutex::new(ChaCha8Rng::seed_from_u64(seed))))
-        } else {
-            Self(Arc::new(StdMutex::new(StdRng::from_entropy())))
-        }
+    /// Creates a new MutRng given an optional `u64` seed and optional pubkey. If `seed_opt` is `Some`, random activity
+    ///  generation in the simulator occurs near-deterministically.
+    /// If it is `None`, activity generation is truly random, and based on a non-deterministic source of entropy.
+    /// If a pubkey is provided, it will be used to salt the seed to ensure each node gets a deterministic but
+    /// different RNG sequence.
+    pub fn new(seed_opt: Option<(u64, Option<&PublicKey>)>) -> Self {
+        let seed = match seed_opt {
+            Some((seed, Some(pubkey))) => {
+                let mut hasher = DefaultHasher::new();
+                let mut combined = pubkey.serialize().to_vec();
+                combined.extend_from_slice(&seed.to_le_bytes());
+                combined.hash(&mut hasher);
+                hasher.finish()
+            },
+            Some((seed, None)) => seed,
+            None => rand::random(),
+        };
+        Self(Arc::new(StdMutex::new(ChaCha8Rng::seed_from_u64(seed))))
     }
 }
 
@@ -552,8 +560,8 @@ pub struct SimulationCfg {
     activity_multiplier: f64,
     /// Configurations for printing results to CSV. Results are not written if this option is None.
     write_results: Option<WriteResults>,
-    /// Random number generator created from fixed seed.
-    seeded_rng: MutRng,
+    /// Optional seed for deterministic random number generation.
+    seed: Option<u64>,
 }
 
 impl SimulationCfg {
@@ -569,7 +577,7 @@ impl SimulationCfg {
             expected_payment_msat,
             activity_multiplier,
             write_results,
-            seeded_rng: MutRng::new(seed),
+            seed,
         }
     }
 }
@@ -985,10 +993,11 @@ impl<C: Clock + 'static> Simulation<C> {
             active_nodes.insert(node_info.pubkey, (node_info, capacity));
         }
 
+        // Create a network generator with a shared RNG for all nodes.
         let network_generator = Arc::new(Mutex::new(
             NetworkGraphView::new(
                 active_nodes.values().cloned().collect(),
-                self.cfg.seeded_rng.clone(),
+                MutRng::new(self.cfg.seed.map(|seed| (seed, None))),
             )
             .map_err(SimulationError::RandomActivityError)?,
         ));
@@ -999,6 +1008,9 @@ impl<C: Clock + 'static> Simulation<C> {
         );
 
         for (node_info, capacity) in active_nodes.values() {
+            // Create a salted RNG for this node based on its pubkey.
+            let seed_opt = self.cfg.seed.map(|seed| (seed, Some(&node_info.pubkey)));
+            let salted_rng = MutRng::new(seed_opt);
             generators.push(ExecutorKit {
                 source_info: node_info.clone(),
                 network_generator: network_generator.clone(),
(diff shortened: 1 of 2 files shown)


simln-lib/src/lib.rs
```diff
@@ -490,18 +489,44 @@ type MutRngType = Arc<StdMutex<dyn RngCore + Send>>;
...
+            seed: seed_opt,
+        }
+    }
+
+    /// Creates a new MutRng that is salted with a pubkey. This ensures that each node
+    /// gets a deterministic but different RNG sequence.
+    pub fn salted(&self, pubkey: &PublicKey) -> Self {
+        // Get the pubkey bytes
+        let pubkey_bytes = pubkey.serialize();
+
```
sangbida, line 518: Would appreciate thoughts on this approach to salting

carlaKC (commented):
Great start on this!

simln-lib/src/lib.rs
```diff
@@ -1522,12 +1549,57 @@ mod tests {
...
+        let mut rng_2 = mut_rng_2.rng.lock().unwrap();
 
         assert_ne!(rng_1.next_u64(), rng_2.next_u64())
     }
 
+    #[test]
+    fn create_salted_mut_rng() {
+        let base_rng = MutRng::new(Some(42));
+
+        let pk1 = PublicKey::from_slice(&[
```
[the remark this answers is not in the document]
sangbida: Ah thanks for this! I was looking for a function like this! I was going to use [this](https://docs.rs/secp256k1/latest/secp256k1/fn.generate_keypair.html) but didn't want to unecessarily add to Cargo.toml.

simln-lib/src/lib.rs
sangbida: Ah okay, we don't care to salt the base seed, right?

simln-lib/src/lib.rs
```diff
@@ -946,6 +971,8 @@ impl Simulation {
         );
 
         for (node_info, capacity) in active_nodes.values() {
+            // Create a salted RNG for this node based on its pubkey
+            let salted_rng = self.cfg.seeded_rng.salted(&node_info.pubkey);
```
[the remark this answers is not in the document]
sangbida: Just to clarify we want to remove `seeded_rng` from the SimulationCfg and replace it with the seed, then generate the rng inside `random_acitivity_nodes`?

simln-lib/src/lib.rs
carlaKC: I think that we do want to include the base seed here just so that there's an option for folks to run with different, fixed seeds? If we just use pubkey, we'll only ever get one set of payments per pubkey, vs if we include the base seed we can change it (while keeping it deterministic)

simln-lib/src/lib.rs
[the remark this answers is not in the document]
carlaKC: Yeah that's what I was thinking - that way we've got the seed available to create our salted RNGs and don't need to store it in the RNG itself.

simln-lib/src/lib.rs
sangbida: Took another stab at this one, in this case I just concatenated the seed with the pubkey and hashed that, which should still give us different but deterministic payments

simln-lib/src/lib.rs
```diff
@@ -946,6 +959,8 @@ impl Simulation {
         );
 
         for (node_info, capacity) in active_nodes.values() {
+            // Create a salted RNG for this node based on its pubkey
+            let salted_rng = base_rng.salted(&node_info.pubkey, self.cfg.seed.unwrap_or(0));
```
carlaKC, line 963: I don't think that this does what we want in the case where `seed` is `None`? In this case we want totally random behavior, and this will always seed with `pubkey + 0` because of the unwrap.

I think that some good options here would be:

1) Move all this abstraction into the `new` function of `MutRng`

`pub fn new(seed_opt: Option<(u64, Option<Pubkey>)>) -> Self`

We'd invoke this differently based on use-case:
```
let base_seed = self.cfg.seed.map(|s| Some((s, None)));
base_rng = new(self.cfg.seed, None);
```

```
let seed_material = self.cfg.seed.map(|s| Some((s, Some(&node_info.pubkey))));
salted_rng = new(seed_material)
```

2) Match on presence of seed 
```
let salted_rng = if let Some(s) = self.cfg.seed {
   MutRng::new(None)
} else {
  MutRng::salted(&node_info.pubkey, s)
}
```

I think that I prefer (1), but open to other approaches.

simln-lib/src/lib.rs
```diff
@@ -501,9 +501,23 @@ impl MutRng {
...
+            Self(Arc::new(StdMutex::new(ChaCha8Rng::from_entropy())))
         }
     }
+
+    /// Creates a new MutRng that is salted with a pubkey. This ensures that each node
+    /// gets a deterministic but different RNG sequence.
+    pub fn salted(&self, pubkey: &PublicKey, seed: u64) -> Self {
+        let mut hasher = DefaultHasher::new();
+
+        // Get pubkey bytes and concatenate with seed bytes
```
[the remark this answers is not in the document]
sangbida: apologies on this one :(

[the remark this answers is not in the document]
carlaKC: absolutely no worries - habits take time!

simln-lib/src/lib.rs
sangbida: I gave this another go by consolidating the salted functionality into new.

simln-lib/src/lib.rs
```diff
@@ -493,16 +493,29 @@ type MutRngType = Arc<StdMutex<dyn RngCore + Send>>;
...
+        let seed = if let Some(seed) = seed_opt {
+            if let Some(pk) = pubkey {
+                let mut hasher = DefaultHasher::new();
+                let mut combined = pk.serialize().to_vec();
+                combined.extend_from_slice(&seed.to_le_bytes());
+                combined.hash(&mut hasher);
+                hasher.finish()
+            } else {
+                seed
+            }
```
carlaKC, line 512: I would prefer to go with the nested options for `seed_opt` / `pubkey` because there is no valid case where you provide a `pubkey` but not a `seed`.

The handling isn't so bad if we get fancy with matching:
```
 pub fn new(seed_opt: Option<(u64, Option<&PublicKey>)>) -> Self {
        let seed = match seed_opt {
            Some(( seed, Some(pubkey) )) => {
                let mut hasher = DefaultHasher::new();
                let mut combined = pk.serialize().to_vec();
                combined.extend_from_slice(&seed.to_le_bytes());
                combined.hash(&mut hasher);
                hasher.finish()
            },
            Some(( seed, None )) => seed,
            None =>  rand::random(),
        }

        Self(Arc::new(StdMutex::new(ChaCha8Rng::seed_from_u64(seed))))
}
```

carlaKC (commented):
One last comment here - really close!

simln-lib/src/lib.rs
```diff
@@ -493,16 +493,25 @@ type MutRngType = Arc<StdMutex<dyn RngCore + Send>>;
...
     /// random activity generation in the simulator occurs near-deterministically.
     /// If it is `None`, activity generation is truly random, and based on a
     /// non-deterministic source of entropy.
-    pub fn new(seed_opt: Option<u64>) -> Self {
-        if let Some(seed) = seed_opt {
-            Self(Arc::new(StdMutex::new(ChaCha8Rng::seed_from_u64(seed))))
-        } else {
-            Self(Arc::new(StdMutex::new(StdRng::from_entropy())))
-        }
+    /// If a pubkey is provided, it will be used to salt the seed to ensure each node gets a deterministic but
```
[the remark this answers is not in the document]
sangbida: I don't think I fully understand the rules about line wrapping here, I thought the cut off was 120 chars and the rest goes to the next line?

simln-lib/src/lib.rs
```diff
@@ -1720,9 +1764,8 @@ mod tests {
         let result = simulation.validate_node_network().await;
 
         assert!(result.is_err());
-        assert!(
-            matches!(result, Err(LightningError::ValidationError(msg)) if msg.contains("we don't control any nodes"))
-        );
+        assert!(matches!(result,
```
carlaKC, line 1767: These three test changes look unnecessary? If it's just a formatting change rather revert since it's unrelated to this pr.

simln-lib/src/lib.rs
```diff
@@ -493,16 +493,24 @@ type MutRngType = Arc<StdMutex<dyn RngCore + Send>>;
...
-    /// random activity generation in the simulator occurs near-deterministically.
-    /// If it is `None`, activity generation is truly random, and based on a
-    /// non-deterministic source of entropy.
-    pub fn new(seed_opt: Option<u64>) -> Self {
-        if let Some(seed) = seed_opt {
-            Self(Arc::new(StdMutex::new(ChaCha8Rng::seed_from_u64(seed))))
-        } else {
-            Self(Arc::new(StdMutex::new(StdRng::from_entropy())))
-        }
+    /// Creates a new MutRng given an optional `u64` seed and optional pubkey. If `seed_opt` is `Some`,random activity
```
carlaKC, line 496: nit: space before `random_activity`

carlaKC (approved):
lgtm, you can go ahead and squash all down to a single commit.

final comments non-blocking

simln-lib/src/lib.rs
sangbida: Yep, the formatter was complaining about these lines after the rebase. I can get rid of it.

Issue #255, opened 2025-04-23
Feature Request: Introduce Forward Interceptors to SimNode


carlaKC:
**Describe desired feature**

Add the ability to "intercept" HTLCs forwarded on `sim-node` mocked lightning nodes, and take custom actions that allow senders to:
- Forward or fail the HTLC
- Attach custom records to the HTLC

We should allow more than one of these interceptors to act on a single HTLC, to allow layering of multiple single-use interceptors - this will require enforcing consistency across interceptors:
* Custom records are merged from all interceptors
* Conflicts cause the simulator to shut down
* HTLC is only forwarded if no single interceptor fails it
* Wait for all interceptors to return before taking action (without blocking simulator)

**Use case for feature**

This allows for very powerful use of sim-ln as a library, where forwarding nodes can take custom actions on traffic.

**Would you like to contribute code for this feature?**

This is (badly) implemented in https://github.com/bitcoin-dev-project/sim-ln/commit/a33e68dd871d7b030a6f87afd17b04d184aeacb6 .. https://github.com/bitcoin-dev-project/sim-ln/commit/2a35bd55b5da2e2cc73b1ed8b03f84bf39613609, I'd be happy to clean it up or have somebody else clean it up and I'll review. If those commits are revived, it's worth having a design discussion first to make sure it's actually an approach we want to go with; they weren't thoroughly designed when written.

elnosh:
thinking of taking a stab at this. From a previous conversation, the intent is that the `intercept_htlc` method should be blocking and create tasks for each of them so that if one of those blocks for long time, it does not block other interceptors too? If that is the case, I think the `intercept_htlc` could look like -> `async fn intercept_htlc(&self, req: InterceptRequest) -> Result<CustomRecords, Forwarding>` or wrap that in another `Result`? instead of sending the response in the channel from `InterceptRequest`

PR #261, opened 2025-05-06
simln-lib: add htlc interceptor for simulated nodes

elnosh:
for https://github.com/bitcoin-dev-project/sim-ln/issues/255

Mostly took changes from these 2 commits to add an `Interceptor` trait:
- https://github.com/carlaKC/sim-ln/commit/c5d7649d0969f524dc1076760128338f38853ed8 
- https://github.com/carlaKC/sim-ln/commit/228706ca3ef70d6111535ae75ef09af724dea448

 Some of the things I changed from those commits:
- `intercept_htlc` now returns a `Result` to communicate the result of the intercepted htlc instead of sending it through a channel.
- spawns a task in a `JoinSet` for each interceptor's `intercept_htlc` so that if any of those holds the htlc for a long time it does not block the other interceptors. It then waits for the completion of the tasks in the joinset. If any of them returns a result to fail the htlc, it drops the `JoinSet` since there is no need to wait for completion of other tasks because the htlc will fail anyways.

### DIFF
--- a/simln-lib/src/sim_node.rs
+++ b/simln-lib/src/sim_node.rs
@@ -9,8 +9,10 @@ use bitcoin::{Network, ScriptBuf, TxOut};
 use lightning::ln::chan_utils::make_funding_redeemscript;
 use serde::{Deserialize, Serialize};
 use std::collections::{hash_map::Entry, HashMap};
+use std::fmt::Display;
 use std::sync::Arc;
 use std::time::UNIX_EPOCH;
+use tokio::task::JoinSet;
 use tokio_util::task::TaskTracker;
 
 use lightning::ln::features::{ChannelFeatures, NodeFeatures};
@@ -61,6 +63,9 @@ pub enum ForwardingError {
     /// The forwarded htlc has insufficient fee for the channel's policy (fee / expected fee / base fee / prop fee).
     #[error("InsufficientFee: offered fee: {0} (base: {1}, prop: {2}) < expected: {3}")]
     InsufficientFee(u64, u64, u64, u64),
+    /// Custom error for interceptors if encountered a forwarding error during htlc interception.
+    #[error("InterceptorError: {0}")]
+    InterceptorError(String),
 }
 
 /// CriticalError represents an error while propagating a payment in a simulated network that
@@ -89,6 +94,13 @@ pub enum CriticalError {
     /// Sanity check on channel balances failed (node balances / channel capacity).
     #[error("SanityCheckFailed: node balance: {0} != capacity: {1}")]
     SanityCheckFailed(u64, u64),
+    /// Intercepted HTLCs have duplicated custom records attached.
+    #[error("DuplicateCustomRecord: key {0}")]
+    DuplicateCustomRecord(u64),
+    /// Custom error for interceptors if they encountered a critical error during htlc
+    /// interception.
+    #[error("InterceptorError: {0}")]
+    InterceptorError(String),
 }
 
 type ForwardResult = Result<Result<(), ForwardingError>, CriticalError>;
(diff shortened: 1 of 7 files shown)


carlaKC:
Driveby comment: could we split up the logic adding a unique index and the interceptor into separate commits?
For the sake of breaking up into smaller logical chunks / reviewability.

simln-lib/src/sim_node.rs
```diff
@@ -930,16 +1074,99 @@ async fn add_htlcs(
...
+                    outgoing_cltv,
+                );
+
+                log::trace!(
+                    "Sending HTLC to intercepor: {} {request}",
+                    interceptor.name()
+                );
+
+                let interceptor_clone = Arc::clone(interceptor);
+                intercepts.spawn(async move { interceptor_clone.intercept_htlc(request).await });
```
carlaKC, line 1123: This might be a good spot for `futures::future::join_all`?

Rather than needing to spawn a new task here, we can just call (but not await) the `intercept_htlc` method and then await them in whatever order they arrive after that?

Author's choice. - I don't see us running with _so many_ interceptors that creating these tasks starts to give us a performanc hit.

simln-lib/src/sim_node.rs
```diff
@@ -930,16 +1074,99 @@ async fn add_htlcs(
...
+                                    }
+                                },
+                                Entry::Vacant(e) => {
+                                    e.insert(v);
+                                },
+                            };
+                        }
+                    },
+                    Ok(Err(e)) => {
+                        drop(intercepts);
```
carlaKC, line 1153: I think that we may need a more gentle shutdown than dropping the task (which afaik will force abort the `intercept_htlc` call).

I can see it being difficult to write interceptor code that works with another system when your code may be run half way and then killed - could end up with funny states. My instinct is to provides a `triggered` pair and instruct interceptors to listen on it for shutdown signals? Open to other approaches as well.

I do think that the ability to shut down all the other interceptors once the first error is reached is a really, really nice feature.

simln-lib/src/sim_node.rs
```diff
@@ -655,6 +668,126 @@ impl<T: SimNetwork> LightningNode for SimNode<'_, T> {
     }
 }
 
+#[async_trait]
+pub trait Interceptor: Send + Sync {
+    /// Implemented by HTLC interceptors that provide input on the resolution of HTLCs forwarded in the simulation.
+    async fn intercept_htlc(&self, req: InterceptRequest)
+        -> Result<CustomRecords, ForwardingError>;
```
carlaKC, line 675: If we only have one layer of `Result` here, how does the `intercept_htlc` call tell the simulator that it's hit a ciritical error and wants to shut down (rather than a forwarding error for this payment specifically)?

carlaKC (commented):
Main comments are around how we handle shutdown in a reaonsable way. Specifically:
* An interceptor has had a critical failure, how does it tell the simulator to shut down
* The simultor needs to terminate the interceptors, how can we cleanly do this so that interception code isn't left in a bad state

simln-lib/src/sim_node.rs
elnosh: looked into `join_all` but if iiuc it will wait until all futures resolve. I think we'd prefer to get the responses as the tasks finish rather than waiting for all to finish? that way we can fail the other interceptors if one failed the htlc.

simln-lib/src/sim_node.rs
elnosh: sounds good :thumbsup: provide that trigger in the `InterceptRequest` and send on it if we need to fail other interceptors? 
just to note, although `intercept_htlc` may force-aborted, interceptor should still get notified about htlc resolution through `notify_resolution` but I think we'd still like to shutdown gently through `intercept_htlc`?

simln-lib/src/sim_node.rs
elnosh: I wasn't sure if that's something we'd want. I didn't like the idea of nested `Result`s. It could instead be communicated through specific variants in `ForwardingError`? Simulation could check with `is_critical` if the `ForwardingError` returned warrants a shutdown.

elnosh:
> * An interceptor has had a critical failure, how does it tell the simulator to shut down
> 

I wasn't sure if letting the interceptor shutdown the simulation was something we'd want but it could be done through specific variants in the `ForwardingError` returned? Simulation could check with `is_critical` if the ForwardingError returned warrants a shutdown. Although `Interceptor`s should be aware of this to trigger a shutdown by providing specific variants. 

> * The simultor needs to terminate the interceptors, how can we cleanly do this so that interception code isn't left in a bad state

should we include a channel or shutdown signal in `InterceptRequest`?

simln-lib/src/sim_node.rs
carlaKC: > I think we'd prefer to get the responses as the tasks finish rather than waiting for all to finish? that way we can fail the other interceptors if one failed the htlc.

Yeah indeed, we want to hear back from the first completing one so that we can cancel the rest. Could use `FuturesUnordered` but I think that's an over-optimization (it's for lots of futures), so happy to leave as-is 👍

simln-lib/src/sim_node.rs
carlaKC: > It could instead be communicated through specific variants in ForwardingError

That seems like a bit of a layering violation to me. An`InterceptorError` for interceptors that might want to fail forwards with their own `ForwardingFalure` reason that isn't defined in simln (eg, you don't have enough reputation) makes sense to me, but we're stretching its definition a bit to cover things like unexpected internal state errors.

Nested results are definitely ugly, but I think that it more clearly represents the two types of failures that we could have here? One for "fail this HTLC", one for "something has gone wrong with my interception". If we define a type for the inner result the function signatures won't be too ugly at least.

simln-lib/src/sim_node.rs
carlaKC: > provide that trigger in the InterceptRequest and send on it if we need to fail other interceptors?

I think that we should handle the triggering? 
- Give each interceptor a listener
- On first interceptor error, trigger shutdown

Then interceptors are responsible for listening on that shutdown listener and getting their state in order before they exit. We still wait for each to finish, but that should be relatively quick because we've signaled that it's shutdown time.

> Interceptor should still get notified about htlc resolution through notify_resolution

As-is I don't think we'd notify if the HTLC is failed by the interceptor? It'll never be "fully" forwarded by the node, so we don't notify it being resolved.

elnosh: > * Give each interceptor a listener 
> * On first interceptor error, trigger shutdown
> 
> 
> Then interceptors are responsible for listening on that shutdown listener and getting their state in order before they exit. We still wait for each to finish, but that should be relatively quick because we've signaled that it's shutdown time.

got it, will do. 

> As-is I don't think we'd notify if the HTLC is failed by the interceptor? It'll never be "fully" forwarded by the node, so we don't notify it being resolved.

I thought we would. Because if an interceptor returns an error, we then return it in `add_htlcs` and then call `remove_htlcs` here: https://github.com/elnosh/sim-ln/blob/2de405fdcb37c8ffe4bd8d2cc0077ef7099cc3ed/simln-lib/src/sim_node.rs#L1274 which internally calls `notify_resolution` for each interceptors

simln-lib/src/sim_node.rs
elnosh: fair! will change 👍

simln-lib/src/sim_node.rs
carlaKC: > I thought we would.

Ok cool, forgot when that is/isn't called bu sgtm!

carlaKC:
WDYT about adapting [this latency interceptor](https://github.com/carlaKC/sim-ln/commit/2a35bd55b5da2e2cc73b1ed8b03f84bf39613609) and [surfacing](https://github.com/carlaKC/sim-ln/commit/71bd12e29c029f19e987ae91c818e645e373b309) it on the CI? 

Nice to have a user of the interceptor that we can run + test in master rather than merging in some unused code and having to test it externally.

elnosh:
sgtm, as I was doing exactly that and testing externally heh.

simln-lib/src/sim_node.rs
```diff
@@ -643,6 +668,126 @@ impl<T: SimNetwork> LightningNode for SimNode<'_, T> {
...
+
+    /// Notifies the interceptor that a previously intercepted htlc has been resolved. Default implementation is a no-op
+    /// for cases where the interceptor only cares about interception, not resolution of htlcs.
+    async fn notify_resolution(
+        &self,
+        _res: InterceptResolution,
+    ) -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
+        Ok(())
+    }
```
elnosh, line 684: rethinking this, I'm wondering what type of error should this be or if this method should return a `Result` at all? Leaning towards this should not return a `Result` since the simulation is just notifying the interceptor of a resolution.

carlaKC: Perhaps the interceptor wants to signal shutdown on the `notify_resolution` because it got information it wasn't expecting? Or a database became unavailable, etc.

elnosh:
Added changes addressing comments. Using nested results now to differentiate between normal and critical errors that could be returned. If a `CriticalError` is received, it returns and triggers a shutdown. If a `ForwardingError` happens during interception, we send a signal to other interceptors to let them know they should shut down. 

Also took your commit https://github.com/carlaKC/sim-ln/commit/2a35bd55b5da2e2cc73b1ed8b03f84bf39613609 as suggested to add the latency interceptor. 

If the simulation runs with the `LatencyInterceptor` it can log some error messages during shutdown 
```
2025-05-14T15:59:40.049Z ERROR [simln_lib] Track payment failed for 7a56bab4c9b27d3cc09438dd6c32255eb4ecb4559337e9125c8920f8b071ab61: Track payment error: shutdown during payment tracking.
```
because we triggered a shutdown while an intercepted payment (with the added latency) hasn't resolved.

carlaKC (commented):
This is looking really good! Only two major comments on shutdown + cli option for latency.

Been a bit nitty about docs because this feature really will only be used when people use simln as a library so it's important to get the docs readable.

simln-lib/src/sim_node.rs
```diff
@@ -980,15 +1102,104 @@ async fn add_htlcs(
...
+
+                let interceptor_clone = Arc::clone(interceptor);
+                intercepts.spawn(async move { interceptor_clone.intercept_htlc(request).await });
+            }
+
+            // Read results from the interceptors and check whether any of them returned an instruction to fail
+            // the HTLC. If any of the interceptors did return an error, we send a shutdown signal
+            // to the other interceptors that may have not returned yet.
+            let mut interceptor_failure = None;
+            while let Some(res) = intercepts.join_next().await {
```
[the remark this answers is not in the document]
elnosh: > I think that we can do this by selecting on `intercepts.join_next()` and pass in `SimGraph`'s `shutdown_trigger` and shut down the interceptors if we get the high level signal that it's time to shut down.

I think we'll need to pass in a shutdown listener from the upstream `Simulation` since currently we don't have a way to listen for a shutdown signal in `SimGraph` afaik. `shutdown_trigger` is the one we use to trigger on critical errors.


> We're somewhat running into a flaw with `triggered` here - that we can't create a child trigger that would shut down with the parent, but maybe that's a feature not a bug. Either way, I think it's outside of the scope of this PR to rework all that so I think a `select` is our best option (even if ugly).

maybe for another PR but could look into [CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html#method.child_token) looks like it lets you create "child tokens" that can get cancelled along with the parent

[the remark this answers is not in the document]
carlaKC: > I think we'll need to pass in a shutdown listener from the upstream Simulation since currently we don't have a way to listen for a shutdown signal in SimGraph afaik.

Right, we need `shutdown_listener` 🤦‍♀️   SGTM, we can add it to `SimGraph` and then pass it along.

> maybe for another PR but could look into [CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html#method.child_token) looks like it lets you create "child tokens" that can get cancelled along with the parent

Yeah I like the idea of investigating this - took a look when we updated the leaky joinset and liked it.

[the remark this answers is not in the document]
elnosh: moved the htlc interception logic to a separate method. I didn't end up needing to use `select` here since we are passing the shutdown listener to `SimGraph` now so it can be passed directly in the `InterceptRequest` to the interceptors.

simln-lib/src/sim_node.rs
```diff
@@ -980,15 +1102,104 @@ async fn add_htlcs(
...
+                                    let existing_value = e.get();
+                                    if *existing_value == v {
+                                        return Err(CriticalError::DuplicateCustomRecord(k));
+                                    }
+                                },
+                                Entry::Vacant(e) => {
+                                    e.insert(v);
+                                },
+                            };
+                        }
```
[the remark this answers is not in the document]
elnosh: Since `CustomRecords` is just a alias for `HashMap`, I ended up leaving as is instead of changing because I would need to wrap the `HashMap` in a struct to define a method on it so I thought it would be ok to just leave it. Lmk if you'd prefer to create type for and I can change it.

simln-lib/src/sim_node.rs
```diff
@@ -692,6 +704,178 @@ impl<T: SimNetwork> LightningNode for SimNode<'_, T> {
...
+                    // Interceptor call succeeded and indicated that we should proceed with the forward. Merge
+                    // any custom records provided, failing if interceptors provide duplicate values for the
+                    // same key.
+                    for (k, v) in records {
+                        match attached_custom_records.entry(k) {
+                            Entry::Occupied(e) => {
+                                let existing_value = e.get();
+                                if *existing_value == v {
+                                    return Err(CriticalError::DuplicateCustomRecord(k));
+                                }
```
carlaKC, line 850: Shouldn't we fail if `existing_value != v`? The same value means we can merge with no conflicts, different values means we've got an invalid result that we don't know how to handle.

simln-lib/src/sim_node.rs
```diff
@@ -692,6 +704,178 @@ impl<T: SimNetwork> LightningNode for SimNode<'_, T> {
...
+            intercepts.spawn(async move { interceptor_clone.intercept_htlc(request).await });
+        }
+
+        // Read results from the interceptors and check whether any of them returned an instruction to fail
+        // the HTLC. If any of the interceptors did return an error, we send a shutdown signal
+        // to the other interceptors that may have not returned yet.
+        let mut interceptor_failure = None;
+        while let Some(res) = intercepts.join_next().await {
+            let res =
+                res.map_err(|join_err| CriticalError::InterceptorError(join_err.to_string()))?;
```
carlaKC, line 837: Wondering about how best to handle `CriticalError` but here and below for `DuplicateCustomRecords`.

When we error out, will `shutdown_trigger` be called somewhere up the stack? Otherwise the other interceptors don't get the signal to shut down. 

I think that the strictly cleanest way to do this would be to:
1. Track that we ran into a critical error
2. Trigger shutdown
3. Loop through remaining tasks to make sure they completed
4. Return the first critical error we hit

Doesn't need to be in this PR, but maybe let's write this up in an issue just so that we don't forget?

simln-lib/src/sim_node.rs
```diff
@@ -980,16 +1189,48 @@ async fn add_htlcs(
...
+            forwarding_node: hop.pubkey,
+            payment_hash,
+            incoming_htlc: incoming_htlc.clone(),
+            incoming_custom_records: incoming_custom_records.clone(),
+            outgoing_channel_id: next_scid,
+            incoming_amount_msat: outgoing_amount,
+            outgoing_amount_msat: outgoing_amount - hop.fee_msat,
+            incoming_expiry_height: outgoing_cltv,
+            outgoing_expiry_height: outgoing_cltv - hop.cltv_expiry_delta,
+            shutdown_listener: shutdown_signal_clone.1.clone(),
```
carlaKC, line 1216: If we use the top level `shutdown_signal` for intercepts and trigger on `FowardingErrors` then won't the simulation shut down any time an interceptor fails a payment? I think that it needs:
- A shutdown trigger that is used by only the interceptors
- A shutdown listener that is used by only the interceptors
- A shutdown listener that is used by the simulation

It's pretty annoying for the interceptor to have two shutdown signals, so we could look at an approach where we monitor the simulation listener and trigger the interceptor one if it's triggered (this is where child triggers would have been great).

simln-lib/src/sim_node.rs
```diff
@@ -1053,9 +1313,20 @@ async fn propagate_payment(
     route: Path,
     payment_hash: PaymentHash,
     sender: Sender<Result<PaymentResult, LightningError>>,
-    shutdown: Trigger,
+    interceptors: Vec<Arc<dyn Interceptor>>,
+    shutdown_signal: (Trigger, Listener),
 ) {
-    let notify_result = match add_htlcs(nodes.clone(), source, route.clone(), payment_hash).await {
+    let shutdown_clone = (shutdown_signal.0.clone(), shutdown_signal.1.clone());
```
carlaKC, line 1319: nit: simplify to `let shutdown_clone = shutdown_signal.clone();`?

simln-lib/src/sim_node.rs
```diff
@@ -1896,4 +2198,168 @@ mod tests {
...
+        // htlc.
+        let (_, result) = test_kit
+            .send_test_payment(test_kit.nodes[0], test_kit.nodes[3], amt)
+            .await;
+
+        assert!(matches!(
+            result.unwrap().payment_outcome,
+            PaymentOutcome::Unknown
+        ));
+
```
carlaKC, line 2272: Right now I can make this test fail by adding a `listener` to the test kit and checking that we've triggered top level shutdown of the simulation (because a failed interceptor shuts everything down).

```
		select!{
			_ = test_kit.listener => {assert!(false, "simulation shut down by interceptor")},
			_ = async {} => {},
		}
```

Would be good to add this in above here to assert that interceptor failures aren't killing the whole simulation.

carlaKC (commented):
Sorry for the slow turnaround on this one!

A few thoughts on tricksey shutdown stuff.

simln-lib/src/sim_node.rs
elnosh: ah, I misunderstood the "conflict" case we wanted to fail. Thought that if they were exactly the same we'd want to fail. But that doesn't make much sense :man_facepalming: will change

simln-lib/src/sim_node.rs
carlaKC: Yeah this is very nitpickey - in the case of critical shutdown a few interceptors might get aborted because the joinset is dropped (this will actually happen before shutdown is triggered, so my previous comment about it getting triggered up the stack doesn't really matter).

Let's leave as-is and make an issue noting that we don't guarantee clean shutdown of interceptors when we hit a `CriticalError`. I don't see it being much of a problem in practice, but good to have an issue so it's known at least.

simln-lib/src/sim_node.rs
carlaKC: TBH I've never found an intuitive way to do this type of merge API - a world where each interceptor is only allowed to set one value also _sort of_ makes sense, but it's just more restrictive.

But yeah, let's go the more permissive "error on conflict" approach 👍

elnosh:
Thanks for the patience! Hopefully closer now :fingers_crossed:

From latest feedback, made following changes:
- Created trigger to be used by interceptors only and passing this listener to the `InterceptRequest`.
- `handle_intercepted_htlc` gets a `shutdown_listener` and if we get a signal to shutdown it will trigger the interceptors to shutdown too. 
- If we get a forwarding error we trigger the interceptors to shutdown but since triggers (simulation shutdown and interceptor shutdown) are separated, this should not cause a simulation to shutdown.

simln-lib/src/sim_node.rs
```diff
@@ -684,6 +704,205 @@ impl<T: SimNetwork> LightningNode for SimNode<'_, T> {
...
+                        },
+                    }
+                }
+                _ = shutdown_listener.clone() => {
+                    // If we get a simulation shutdown signal, we trigger the interceptors to
+                    // shutdown and wait for them to finish before returning.
+                    interceptor_trigger.trigger();
+                    while intercepts.join_next().await.is_some() {}
+                    break 'get_resp
+                }
```
elnosh, line 894: one thing I wasn't completely sure was here. If we get simulation shutdown, it triggers the signal to the interceptors and it just waits for them to finish and then break

elnosh:
thanks! added debug statement and `is_empty` check

PR #262, opened 2025-05-06
docker: fix base image to 1.85

carlaKC:
Changes to later versions of rust have broken our dockerfile. This change fixes our base image to the version of rust we're currently using, where the break is not present.

We'll need to fix this in future when we bump versions, but for now we make a minimum change to be able to cut the 2.4 release.

Testing this on my personal fork: https://github.com/carlaKC/sim-ln/actions/runs/14882930754
Which will push to my dockerhub account: https://hub.docker.com/r/carlakirkcohen/simln/tags

### DIFF
--- a/.github/workflows/push-docker-image.yml
+++ b/.github/workflows/push-docker-image.yml
@@ -67,7 +67,7 @@ jobs:
       - name: Upload digest
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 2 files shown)


carlaKC:
Successfully tested this on my own fork, so going to go ahead and merge + tag.

PR #265, opened 2025-05-12
simln-lib: refactor ForwardingError to nested results

elnosh:
doing as a prefactor for https://github.com/bitcoin-dev-project/sim-ln/pull/261 

Previously we were checking for critical errors with `is_critical` on the `ForwardingError` to trigger a simulation shutdown. Added a `CriticalError` type that includes the errors we considered critical and now returning them throughout the payment propagation in nested results to better differentiate between the 2 types of errors.

### DIFF
--- a/simln-lib/src/sim_node.rs
+++ b/simln-lib/src/sim_node.rs
@@ -36,22 +36,6 @@ use crate::ShortChannelID;
 /// node and used for analysis.
 #[derive(Debug, Error)]
 pub enum ForwardingError {
-    /// Zero amount htlcs are invalid in the protocol.
-    #[error("ZeroAmountHtlc")]
-    ZeroAmountHtlc,
-    /// The outgoing channel id was not found in the network graph.
-    #[error("ChannelNotFound: {0}")]
-    ChannelNotFound(ShortChannelID),
-    /// The node pubkey provided was not associated with the channel in the network graph.
-    #[error("NodeNotFound: {0:?}")]
-    NodeNotFound(PublicKey),
-    /// The channel has already forwarded an HTLC with the payment hash provided.
-    /// TODO: remove if MPP support is added.
-    #[error("PaymentHashExists: {0:?}")]
-    PaymentHashExists(PaymentHash),
-    /// An htlc with the payment hash provided could not be found to resolve.
-    #[error("PaymentHashNotFound: {0:?}")]
-    PaymentHashNotFound(PaymentHash),
     /// The forwarding node did not have sufficient outgoing balance to forward the htlc (htlc amount / balance).
     #[error("InsufficientBalance: amount: {0} > balance: {1}")]
     InsufficientBalance(u64, u64),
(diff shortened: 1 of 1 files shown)


simln-lib/src/sim_node.rs
```diff
@@ -85,21 +91,7 @@ pub enum ForwardingError {
...
-                | ForwardingError::ChannelNotFound(_)
-                | ForwardingError::NodeNotFound(_)
-                | ForwardingError::PaymentHashExists(_)
-                | ForwardingError::PaymentHashNotFound(_)
-                | ForwardingError::SanityCheckFailed(_, _)
-                | ForwardingError::FeeOverflow(_, _, _)
-        )
-    }
-}
+type ForwardResult = Result<Result<(), ForwardingError>, CriticalError>;
```
carlaKC, line 94: Will the plan here to be to add a variant for interceptors in both `CriticalError` and `ForwardingError` which external implementations would use?

An alternative _could_ be to just make this a `BoxError` instead of `CriticalError` so that callers don't have to wrap their errors. There's no getting  round adding a variant for `ForwardingError` so seems reasonable to just do this consistently. 

Mostly musing out loud, think we could go either way here.

simln-lib/src/sim_node.rs
```diff
@@ -384,12 +395,13 @@ impl SimulatedChannel {
...
         success: bool,
-    ) -> Result<(), ForwardingError> {
+    ) -> Result<(), CriticalError> {
         let htlc = self
             .get_node_mut(sending_node)?
             .remove_outgoing_htlc(hash)?;
         self.settle_htlc(sending_node, htlc.amount_msat, success)?;
-        self.sanity_check()
+        self.sanity_check()?;
+        Ok(())
```
carlaKC, line 404: Can just return `self.sanity_check()`?

simln-lib/src/sim_node.rs
```diff
@@ -992,7 +1004,7 @@ async fn remove_htlcs(
     route: Path,
     payment_hash: PaymentHash,
     success: bool,
-) -> Result<(), ForwardingError> {
+) -> ForwardResult {
```
carlaKC, line 1007: I think that this can just be `Result<(), CriticalError>` - we only hit critical failures here.

carlaKC (approved):
One vague musing about error handling + some nits, nice PR 🌵

simln-lib/src/sim_node.rs
elnosh: > Will the plan here to be to add a variant for interceptors in both `CriticalError` and `ForwardingError` which external implementations would use?

yeah. If they encountered some regular error, provide it in something like `ForwardingError::InterceptError(error)` and if they encountered an error that warranted a simulation shutdown `CriticalError::InterceptError(critical_error_msg)`

> There's no getting round adding a variant for `ForwardingError` so seems reasonable to just do this consistently.

sorry, not quite sure I understood this part. Could you expand here a bit more :sweat_smile:

carlaKC: > sorry, not quite sure I understood this part. Could you expand here a bit more 😅

Just meaning that we're definitely going to have to add `ForwardingError::InterceptorError(...)` so that interceptor errors can provide "inner" errors - basically what you said above!

PR #273, opened 2025-05-26
Add pathfinder trait

sangbida:


sangbida:
I'm away this week and intermittently through June, so this PR might take some time to complete.

simln-lib/src/sim_node.rs
```diff
@@ -484,46 +494,111 @@ pub trait SimNetwork: Send + Sync {
...
+        dest: PublicKey,
+        amount_msat: u64,
+        pathfinding_graph: &NetworkGraph<&'static WrappedLog>,
+    ) -> Result<Route, SimulationError> {
+        let scorer_graph = NetworkGraph::new(bitcoin::Network::Regtest, &WrappedLog {});
+        let scorer = ProbabilisticScorer::new(
+            ProbabilisticScoringDecayParameters::default(),
+            Arc::new(scorer_graph),
+            &WrappedLog {},
+        );
```
carlaKC, line 547: I think that this takes us a step backwards from the goal of having one scorer per `SimNode`? I would have thought that we'd have a `HashMap<Pubkey, Scorer> ` here so that we can look up the source node's scorer and use it per `find_route` call.

Having said that, this PR has made me realize that we're not actually reporting any of our outcomes to the scorers (which means we're basically using an empty one every time).  Wrote up details in #279 - sadly I think this PR will need to wait for that, as it'll directly impact this API (we now also need updating as part of `Pathfinder`, though it may be as simple as just including the `ScorerUpdate` trait).

sangbida: Aah! That's interesting. When I started on this issue I thought the primary objective was to have generic pathfinding. 

I did add a hashmap with the pubkey-scorer but ran into a bunch of lifetime issues so decided to simplify it with a fresh scorer each time :(

Also, this issue only pertains to scoring with the DefaultPathfinder, right? Because the onus of maintaining scorer feedback would be on the user maintaining their custom Pathfinder. 

So if you add a feedback mechanism for the scorer it would make sense for it to live inside DefaultPathFinder?

carlaKC: > Because the onus of maintaining scorer feedback would be on the user maintaining their custom Pathfinder.

True, but we'd need to have to provide a way for the simulator to report success/failure to the custom pathfinder - they _could_ use `track_payment` but I think it makes sense for us to require that they implement `report_success/failure` as part of the trait so that we can handle that all automatically?

> I could create a PR on top of this branch to solve for the scorer feedback mechanism in https://github.com/bitcoin-dev-project/sim-ln/issues/279 or I could close this PR for now until the scorer feedback issue is resolved and refactor this code so it works with the feedback mechanism.

I think it makes sense to get feedback in and then come back to this PR? If you'd be happy to take stab at #279 that would be great! I think we should start with that, leave this PR open and then rebase this on top of it?

carlaKC: So thinking that the final pathfinder api would be:
```
find_route
report_payment_success
report_payment_failure
``

Very similar to what LDK has right now (in `ScorerUpdate`). We then set the simulator up to report success/failure to whatever pathfinder is plugged in.

sangbida: I like your Pathfinder API approach! That makes sense to me! I've updated the description to reflect that the development is paused until we have more feedback and we have closed out #279. 

I'm happy to take a stab at #279 but I will be going on holiday for around 2 weeks in a few days so it might just sit there for a while. I'm happy for someone else to pick it up in the meantime so there can be progress and I'll rebase this PR against it, but if no one has picked it up by then I'll definitely pick it up :)

carlaKC:
@sangbida if you've got some time for this, it's ready for a rebase now that #291 if in!

@chuksys perhaps you could review or pick this up if @sangbida doesn't have bandwidth for it rn.

f3r10:
@sangbida @chuksys let me know if you need any help. I have been using this PR as a base for https://github.com/bitcoin-dev-project/sim-ln/issues/289

PR #275, opened 2025-05-27
simln-lib: remove unnecessary async in SimNode trait

### LINKED ISSUE #267: simln-lib: `SimNetwork` async trait
does the `SimNetwork` trait need to be async? Its method `lookup_node` and `list_nodes` are marked async but it does not seem necessary. The only impl of it `SimGraph` implements these without any async calls: 

https://github.com/bitcoin-dev-project/sim-ln/blob/8cf204263c3b191c06235e92da7d49cc2afc80a6/simln-lib/src/sim_node.rs#L874-L879

carlaKC:
Fixes #267


elnosh (approved):
LGTM

Issue #243, opened 2025-03-28
Feature Request: Fully Deterministic produce_events


carlaKC:
⚠ This will be a non-trivial refactor ⚠ 

**Describe desired feature**

SimLN can currently be run with a fixed seed for its random activity setting, but payments produced by the simulator will not be fully deterministic.

This is because we have a single `DestinationGenerator` which is used across many `produce_events` tasks*. We can hit the following sequence of events:
- We have an RNG that produces values 1, 2, 3 then 4
- We have two `produce_events` tasks A and B that will both sleep for 5 seconds in between payments**
- Both A and B are competing for the same lock on our `Arc<Mutex<DestinationGeneartor>>`

First run:
- A gets lock first: A gets value 1, B gets value 2
- A gets lock first: A gets value 3, B gets value 4

Second run:
- A gets lock first: A gets value 1, B gets value 2
- B gets lock first: A gets value 4, B gets value 3

Across the two runs, the nodes get different destinations based on the order in which they acquire the lock. This averages out across multiple runs, but isn't properly deterministic. We should fix that.

\* this was intentional, we don't want to have to store our lightning graph many times over
\** the sleep time is random, so it's not _that_ likely that they sleep for exactly the same time, but not impossible (especially if we start to speed up the clock)

**Use case for feature**

Fully deterministic runs are important for reproducible simulations.

**Design Considerations**

Part of this issue will be proposing a solution for this issue and discussing it here before opening up a PR, because it'll likely be a larger change to the codebase.

Incomplete thoughts that I have had about this are:
1) Bite the memory bullet and copy the network graph for each node so that they have a distinct `DestinationGenerator`

This gets rid of the locking issue, as each will lock their own destination and always get the same fixed values.

2) Have a single `produce_events` loop that deterministically queues events for each node

Rather than spinning up a task for each node to produce its own events, have one master task that's responsible for dispatching events for each of the nodes. This would look something like:
- Get the next payment time for each node
- Drop them in a sorted heap
- Pull the next item off the heap, get the destination from the shared `DestinationGenerator`
- Once dispatched, get the next payment time for the node and push it onto the heap

3) There are probably better ways to do this, but my potato-brain hasn't thought of them.

**Would you like to contribute code for this feature?**

I'll collaborate on design + contribute review, but won't be able to champion it myself.

Onyekachukwu-Nweke:
I would like to pick this up

carlaKC:
> I would like to pick this up

[...]

PR #277, opened 2025-06-03
simln-lib/refactor: fully deterministic produce events

f3r10:
Addresses #243

### DIFF
--- a/simln-lib/src/lib.rs
+++ b/simln-lib/src/lib.rs
@@ -1,3 +1,4 @@
+// use anyhow::Ok;
 use async_trait::async_trait;
 use bitcoin::secp256k1::PublicKey;
 use bitcoin::Network;
@@ -9,7 +10,8 @@ use rand::{Rng, RngCore, SeedableRng};
 use rand_chacha::ChaCha8Rng;
 use random_activity::RandomActivityError;
 use serde::{Deserialize, Serialize};
-use std::collections::HashSet;
+use std::cmp::Reverse;
+use std::collections::{BTreeMap, BinaryHeap, HashSet};
 use std::fmt::{Display, Formatter};
 use std::marker::Send;
 use std::path::PathBuf;
@@ -476,7 +478,7 @@ pub struct Simulation {
     /// Config for the simulation itself.
     cfg: SimulationCfg,
     /// The lightning node that is being simulated.
-    nodes: HashMap<PublicKey, Arc<Mutex<dyn LightningNode>>>,
+    nodes: BTreeMap<PublicKey, Arc<Mutex<dyn LightningNode>>>,
     /// The activity that are to be executed on the node.
     activity: Vec<ActivityDefinition>,
     /// Results logger that holds the simulation statistics.
@@ -503,14 +505,14 @@ struct ExecutorKit {
     source_info: NodeInfo,
     /// We use an arc mutex here because some implementations of the trait will be very expensive to clone.
     /// See [NetworkGraphView] for details.
-    network_generator: Arc<Mutex<dyn DestinationGenerator>>,
+    network_generator: Arc<dyn DestinationGenerator>,
     payment_generator: Box<dyn PaymentGenerator>,
 }
 
 impl Simulation {
     pub fn new(
         cfg: SimulationCfg,
-        nodes: HashMap<PublicKey, Arc<Mutex<dyn LightningNode>>>,
+        nodes: BTreeMap<PublicKey, Arc<Mutex<dyn LightningNode>>>,
         activity: Vec<ActivityDefinition>,
         tasks: TaskTracker,
     ) -> Self {
@@ -819,7 +821,7 @@ impl Simulation {
                     source_info: description.source.clone(),
                     // Defined activities have very simple generators, so the traits required are implemented on
                     // a single struct which we just cheaply clone.
-                    network_generator: Arc::new(Mutex::new(activity_generator.clone())),
+                    network_generator: Arc::new(activity_generator.clone()),
                     payment_generator: Box::new(activity_generator),
                 });
             }
@@ -836,7 +838,7 @@ impl Simulation {
         // Collect capacity of each node from its view of its own channels. Total capacity is divided by two to
         // avoid double counting capacity (as each node has a counterparty in the channel).
         let mut generators = Vec::new();
-        let mut active_nodes = HashMap::new();
+        let mut active_nodes = BTreeMap::new();
 
         // Do a first pass to get the capacity of each node which we need to be able to create a network generator.
         // While we're at it, we get the node info and store it with capacity to create activity generators in our
@@ -859,17 +861,12 @@ impl Simulation {
             active_nodes.insert(node_info.pubkey, (node_info, capacity));
         }
 
-        let network_generator = Arc::new(Mutex::new(
+        let network_generator = Arc::new(
             NetworkGraphView::new(
                 active_nodes.values().cloned().collect(),
                 self.cfg.seeded_rng.clone(),
             )
             .map_err(SimulationError::RandomActivityError)?,
-        ));
-
-        log::info!(
-            "Created network generator: {}.",
-            network_generator.lock().await
         );
 
         for (node_info, capacity) in active_nodes.values() {
@@ -936,6 +933,33 @@ impl Simulation {
         channels
     }
 
+    fn generate_payments<A: PaymentGenerator + ?Sized>(
+        &self,
+        heap: &mut BinaryHeap<Reverse<(Duration, usize)>>,
+        source: &NodeInfo,
+        node_generator: &A,
+        index: usize,
+    ) -> Result<(), SimulationError> {
+        let mut current_count = 0;
+        loop {
+            let wait = get_payment_delay(current_count, source, node_generator)?;
+            heap.push(Reverse((wait, index)));
+            if node_generator.payment_count().is_none() {
+                return Ok(());
+            }
+            if let Some(c) = node_generator.payment_count() {
+                if c == current_count {
+                    log::info!(
+                        "Payment count has been met for {source}: {c} payments. Stopping the activity."
+                    );
+                    return Ok(());
+                }
+            }
+            current_count += 1;
+
+        }
+    }
+
     /// Responsible for spinning up producers for a set of activities. Requires that a consumer channel is present
     /// for every source node in the set of executors.
     async fn dispatch_producers(
@@ -944,33 +968,63 @@ impl Simulation {
         producer_channels: HashMap<PublicKey, Sender<SimulationEvent>>,
         tasks: &TaskTracker,
     ) -> Result<(), SimulationError> {
-        for executor in executors {
+        let mut heap: BinaryHeap<Reverse<(Duration, usize)>> = BinaryHeap::new();
+        for (index, executor) in executors.iter().enumerate() {
+            self.generate_payments(
+                &mut heap,
+                &executor.source_info,
+                executor.payment_generator.as_ref(),
+                index,
+            )?;
+
+        }
+        while let Some(Reverse((wait, index))) = heap.pop() {
+            let executor = executors
+                .get(index)
+                .ok_or(SimulationError::MissingNodeError("Missing executor".to_string()))?;
             let sender = producer_channels.get(&executor.source_info.pubkey).ok_or(
                 SimulationError::RandomActivityError(RandomActivityError::ValueError(format!(
                     "Activity producer for: {} not found.",
                     executor.source_info.pubkey,
                 ))),
             )?;
+            let (destination, capacity) = executor
+                .network_generator
+                .choose_destination(executor.source_info.pubkey)
+                .map_err(SimulationError::DestinationGenerationError)?;
 
             // pe: produce events
             let pe_shutdown = self.shutdown_trigger.clone();
             let pe_listener = self.shutdown_listener.clone();
             let pe_sender = sender.clone();
+            let source = executor.source_info.clone();
+            // Only proceed with a payment if the amount is non-zero, otherwise skip this round. If we can't get
+            // a payment amount something has gone wrong (because we should have validated that we can always
+            // generate amounts), so we exit.
+            let amount = match executor.payment_generator.payment_amount(capacity) {
+                Ok(amt) => {
+                    if amt == 0 {
+                        log::debug!("Skipping zero amount payment for {source} -> {destination}.");
+                    }
+                    amt
+                },
+                Err(e) => {
+                    return Err(SimulationError::PaymentGenerationError(e));
+                },
+            };
             tasks.spawn(async move {
-                let source = executor.source_info.clone();
-
                 log::info!(
-                    "Starting activity producer for {}: {}.",
-                    source,
-                    executor.payment_generator
+                    "Starting activity producer for {}.",
+                    source
                 );
 
                 if let Err(e) = produce_events(
-                    executor.source_info,
-                    executor.network_generator,
-                    executor.payment_generator,
+                    source.clone(),
                     pe_sender,
                     pe_listener,
+                    wait,
+                    destination,
+                    amount,
                 )
                 .await
                 {
@@ -981,7 +1035,6 @@ impl Simulation {
                 }
             });
         }
-
         Ok(())
     }
 }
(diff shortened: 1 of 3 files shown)


f3r10:
Opening in draft still need to fix some issues.
But I think it is ready for a first review pass @carlaKC

carlaKC (commented):
Direction looking good here! Main comment is that I think we need to have a step where we replenish our heap by calling `generate_payments` again?
- If `payment_count().is_none()` we return a single payment from `generate_payments`
- For `RandomPaymentActivity`, this means we'll do one payment per node and then shut down?

Related to this is that we _possibly_ don't want to queue up tons of events for when `payment_count` is defined (say, we want a million payments, we'll queue up a million items which is a bit of a memory waste). This _probably_ isn't much of a big deal, because I'd imagine this use case is primarily for smaller numbers but something to keep in mind as we address the above requirement.

Also would be good to rebase this early on to get to a more current base 🙏

f3r10:
> Direction looking good here! Main comment is that I think we need to have a step where we replenish our heap by calling `generate_payments` again?

The idea would be to generate all the payments at once, so the master task would dispatch the events. 

> If payment_count().is_none() we return a single payment from generate_payments

Yes, in this case, only one payment is generated

> For RandomPaymentActivity, this means we'll do one payment per node and then shut down?

Yes, right now it is working in this mode :thinking: 

> Related to this is that we possibly don't want to queue up tons of events for when payment_count is defined (say, we want a million payments, we'll queue up a million items which is a bit of a memory waste). This probably isn't much of a big deal, because I'd imagine this use case is primarily for smaller numbers but something to keep in mind as we address the above requirement.

Yes, you are right, maybe it would be better to create some batches of payments. I am going to try to come up with an alternative to reduce the memory waste. :thinking:

f3r10:
Hello Carla, I have been working on a new approach.

1. Central manager task spawned on a unique threat in charge of getting the random destination and the random wait time.
2. For each executor, spawn a new thread with a channel listening for payment events that will finally send the event to the consumers.
3. For each executor, spawn a new thread in charge of generating the payment events. (Each one of these threads using a channel ask to the manager for the wait time. After waiting the required wait time, another event is sent over another channel to the manager in order to calculate the random destination. Once the random destination is calculated, another event is sent over yet another channel (step 2).

Using this approach, the wait_time and the final destination are totally deterministic. But there is still one issue. 
In step_3 it is not possible to control how the threads are spawned.
Example:
One execution
ex1 -> wait_time 0 -> destination node_3
ex2 -> wait_time 1 -> destination node_4

Another execution
ex2 -> wait_time 0 -> destination node_3

carlaKC:
I think I'm missing something, but I'm not sure how this addresses the issues of different threads firing in different orders?

Say we have two Payment Event Generators, A and B, and they both send payments every 5 seconds.
Say we have:
* Two Payment Event Generators, A and B, which each send payments every 5 seconds
* A single RNG, which returns the values destination 1, 2, 3, 4

Each event generator will need to ask the central manager for a destination every 5 seconds, and the order that this happens will depend on which request propagates first? Whether this is by channel or mutex, we can't control it - meaning that we can get different destinations, as is the case now. 

I think that the key is to deterministically generate a queue of events in once place, and then dispatch them in-order so that we always get the same order. This PR gets pretty close to that, it just needs a way to replenish it's loop which I don't think it too terrible of a change.

carlaKC:
Deleted previous comment - it had some misunderstandings.

Why can't we keep the current approach of generating a queue of events and then replenish the queue when we run out of events? By generating all of our payment data in one place, we don't need to worry about thread execution order. 

I think that this can be as simple as pushing a new event to the queue every time we pop one? We might need to track some state for payment count (because we'll need to remember how many we've had), but for random activity it should be reasonable.

carlaKC (commented):
Some relative timestamp issues that we need to address - didn't occur to me earlier in design phase.

I noticed this when testing out against a toy network, would be good to have some unit tests to assert that we get the payment sequence that we're expecting (should be reasonable to do with defined activities with set wait times / counts).

simln-lib/src/lib.rs
```diff
@@ -619,14 +620,42 @@ struct ExecutorKit {
     source_info: NodeInfo,
     /// We use an arc mutex here because some implementations of the trait will be very expensive to clone.
     /// See [NetworkGraphView] for details.
-    network_generator: Arc<Mutex<dyn DestinationGenerator>>,
+    network_generator: Arc<dyn DestinationGenerator>,
     payment_generator: Box<dyn PaymentGenerator>,
 }
 
+struct PaymentEvent {
+    wait_time: Duration,
```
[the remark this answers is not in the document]
f3r10: Oh!! right good catch

simln-lib/src/lib.rs
```diff
@@ -1084,47 +1107,108 @@ impl<C: Clock + 'static> Simulation<C> {
...
+                                    amt
+                                },
+                                Err(e) => {
+                                    return Err(SimulationError::PaymentGenerationError(e));
+                                },
+                            };
+                            // Wait until our time to next payment has elapsed then execute a random amount payment to a random
+                            // destination.
+                            pe_clock.sleep(wait_time).await;
+                            t.spawn(async move {
```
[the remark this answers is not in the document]
f3r10: Without the loop, the consumer channel is closed, and it is not possible to send more consume_events :thinking:

simln-lib/src/lib.rs
```diff
@@ -591,7 +592,7 @@ pub struct Simulation<C: Clock + 'static> {
     /// Config for the simulation itself.
     cfg: SimulationCfg,
     /// The lightning node that is being simulated.
-    nodes: HashMap<PublicKey, Arc<Mutex<dyn LightningNode>>>,
+    nodes: BTreeMap<PublicKey, Arc<Mutex<dyn LightningNode>>>,
```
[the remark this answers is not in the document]
f3r10: Because the iteration of the values of a HashMap is not deterministic. For example, in one run the nodes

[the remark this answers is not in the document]
f3r10: Because the iteration of the values of a HashMap is not deterministic. 
For example, if there are 5 nodes N1, N2, N3, N4, N5
In one run, the iteration could be:
- N5 -> N1 / 1sat
- N3 -> N2  / 1sat
- N1 -> N5 / 10 sat

On a second run, could it be
N1 -> N1 / 1 sat - this is not valid, and it is necessary to choose another destination
N1 -> N2 / 1 sat 
N5 -> N5 / 10 sat - also not valid ....
and so on

f3r10:
Still thinking about how to add tests to assert that we get the payment sequence that we're expecting :thinking: 
Right now, there is only one test checking the number of payments and the expected elapsed time

simln-lib/src/lib.rs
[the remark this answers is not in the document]
carlaKC: Looking at the latest version - isn't relative time still an issue here?

simln-lib/src/lib.rs
```diff
@@ -1085,47 +1109,116 @@ impl<C: Clock + 'static> Simulation<C> {
...
+                                .lock()
+                                .await
+                                .get(&executor.source_info.pubkey)
+                                .ok_or(SimulationError::PaymentGenerationError(PaymentGenerationError(format!("executor {} not found", executor.source_info))))?
+                                .clone();
+                            if let Some(c) = executor.payment_generator.payment_count() {
+                                if c == payload.current_count {
+                                    log::info!(
+                                        "Payment count has been met for {}: {c} payments. Stopping the activity."
+                                    , executor.source_info);
```
carlaKC, line 1157: nit: just to be aware of, `cargo fmt` breaks in selects for some ungodly reason, so it's worth trying to do a little manual formatting

simln-lib/src/lib.rs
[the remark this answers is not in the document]
carlaKC: Can't we also kill the channel on `consume_events` completely? And just directly pass it a `SimulationEvent` that it dispatches to the node? If we do that in a task, it's okay if the event takes a while to dispatch.

simln-lib/src/lib.rs
```diff
@@ -1085,47 +1109,116 @@ impl<C: Clock + 'static> Simulation<C> {
...
+                                },
+                                Err(e) => {
+                                    return Err(SimulationError::PaymentGenerationError(e));
+                                },
+                            };
+                            generate_payment(&mut heap, executor, payload.current_count + 1, payment_event_payloads.clone()).await?;
+
+                            // Wait until our time to next payment has elapsed then execute a random amount payment to a random
+                            // destination.
+                            pe_clock.sleep(payload.wait_time).await;
```
carlaKC, line 1198: Still need to select on shutdown?

simln-lib/src/lib.rs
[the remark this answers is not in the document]
f3r10: No really, in the latest version, `absolute_time` is used for heap ordering, whereas the `wait_time` is for sleeping. I tried to use `absolute_time` for sleeping, but it is not exact.

I added two unit tests, one for the defined activities with a defined payment count and the other for testing the random Rng, and the behaviour is the expected :thinking:

simln-lib/src/lib.rs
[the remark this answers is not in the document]
f3r10: you are right, hashmap lookups are O(1)

simln-lib/src/lib.rs
f3r10: yes, otherwise the task that is processing the payment's head does not stop.

simln-lib/src/lib.rs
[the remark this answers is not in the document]
carlaKC: > No, really. In the latest version, absolute_time is used for heap ordering, whereas wait_time is only for sleeping.

Seems like unnecessary complexity to duplicate these values when we could just track `absolute_time` and sleep for `abs time - now`?

simln-lib/src/lib.rs
```diff
@@ -1031,257 +1056,194 @@ impl<C: Clock + 'static> Simulation<C> {
...
-                    clock,
-                    pe_sender,
-                    pe_listener,
-                )
+        // in order to choose a deterministic destination is necessary to sort the nodes by its public key
+        executors.sort_by(|ex1, ex2| ex1.source_info.pubkey.cmp(&ex2.source_info.pubkey));
+        for executor in executors {
+            let payload = PaymentEventPayload {
+                wait_time: Duration::from_secs(0),
+                destination: executor.source_info.clone(),
```
carlaKC, line 1076: Setting a field called destination to the source value and then overwriting it later is some serious code smell to me.

carlaKC (commented):
I'm finding the design decisions in this PR quite difficult to follow. To me the obvious approach is to queue cheap-to-copy payment events and track node stats/executors in one big hashmap. Happy to hear arguments for this way, but currently not convinced.

simln-lib/src/lib.rs
f3r10: yes you are right, it is necessary to add a `select` in this part

f3r10:
> I'm finding the design decisions in this PR quite difficult to follow. To me the obvious approach is to queue cheap-to-copy payment events and track node stats/executors in one big hashmap. Happy to hear arguments for this way, but currently not convinced.

Thanks for the feedback, using one big hashmap for the executors is a much better design.

simln-lib/src/lib.rs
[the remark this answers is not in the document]
f3r10: you are right @carlaKC I think I found the problem. Right now, the next payment is generated before the `sleep`; therefore, the difference between  `abs time` and `now` is, on some occasions, negative. If the next payment is generated after the `sleep` part, the difference is always positive.

Using the new approach, there are still problems with negative `times`, but in the order of `847.822µs`, which should be zero.

.. quite difficult dealing with time :sweat_smile:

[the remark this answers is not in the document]
carlaKC: > .. quite difficult dealing with time 😅

<details>
  <summary>indeed 💀</summary>
  time
</details>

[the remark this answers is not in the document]
carlaKC: > Using the new approach, there are still problems with negative times, but in the order of -847.822µs, which should be zero.

We might want to just add some sort of quality check in here that this isn't drifting too far? It we're lagging too far behind then the expected number of payments we simulate will be off. We can just pick a magic number and error out if it's too big. 

Happy to leave this for a follow up!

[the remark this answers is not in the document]
f3r10: This is the part that is dealing with it https://github.com/bitcoin-dev-project/sim-ln/pull/277/files#diff-887dd37ec76113778ea02ac50b44aa1ee99a96b73e917f2fa1d3e42358792f43R1104 
For now, if the times are negative, it will always return 0s.

I also found another issue related to how time measurements are done in Rust. For example, instead of getting 4s, the result will be `3.99324`, and at the end, summing up those differences would result in breaking the deterministic aspect that we want to achieve.

simln-lib/src/lib.rs
```diff
@@ -620,10 +621,42 @@ struct ExecutorKit {
     source_info: NodeInfo,
     /// We use an arc mutex here because some implementations of the trait will be very expensive to clone.
     /// See [NetworkGraphView] for details.
-    network_generator: Arc<Mutex<dyn DestinationGenerator>>,
+    network_generator: Arc<dyn DestinationGenerator>,
     payment_generator: Box<dyn PaymentGenerator>,
 }
 
+struct PaymentEvent {
+    pubkey: PublicKey,
```
carlaKC, line 629: nit: pubkey -> source?

simln-lib/src/lib.rs
```diff
@@ -620,10 +621,42 @@ struct ExecutorKit {
...
+    absolute_time: SystemTime,
+    destination: NodeInfo,
+    capacity: Option<u64>,
+}
+
+struct PaymentEventPayload {
```
carlaKC, line 635: This isn't every intuitively named.

simln-lib/src/lib.rs
```diff
@@ -780,20 +813,12 @@ impl<C: Clock + 'static> Simulation<C> {
...
-            event_sender.clone(),
-            &self.tasks,
-        );
 
         // Next, we'll spin up our actual producers that will be responsible for triggering the configured activity.
         // The producers will use their own TaskTracker so that the simulation can be shutdown if they all finish.
         let producer_tasks = TaskTracker::new();
         match self
-            .dispatch_producers(activities, consumer_channels, &producer_tasks)
+            .dispatch_producers(activities, event_sender.clone(), &producer_tasks)
```
carlaKC, line 821: unnecessary clone

simln-lib/src/lib.rs
```diff
@@ -995,18 +1020,17 @@ impl<C: Clock + 'static> Simulation<C> {
             active_nodes.insert(node_info.pubkey, (node_info, capacity));
         }
 
-        // Create a network generator with a shared RNG for all nodes.
-        let network_generator = Arc::new(Mutex::new(
+        // in order to choose a deterministic destination is necessary to sort the nodes by its public key
```
carlaKC, line 1023: nit: Comments should start with capital and end with `.`

simln-lib/src/lib.rs
```diff
@@ -995,18 +1020,17 @@ impl<C: Clock + 'static> Simulation<C> {
...
+        let mut sorted_actives_nodes: Vec<(NodeInfo, u64)> =
+            active_nodes.values().cloned().collect();
+        sorted_actives_nodes.sort_by(|n1, n2| n1.0.pubkey.cmp(&n2.0.pubkey));
+
+        let network_generator = Arc::new(
             NetworkGraphView::new(
-                active_nodes.values().cloned().collect(),
+                sorted_actives_nodes,
```
carlaKC, line 1030: Here the caller somewhat assumes understanding of the implementation of the network generator. 

I think that we should move this sort into `NetworkGraphView` and then add a note on the trait that implementations should be deterministic if called with the same seed.

simln-lib/src/lib.rs
```diff
@@ -1031,257 +1055,191 @@ impl<C: Clock + 'static> Simulation<C> {
...
-                {
-                    pe_shutdown.trigger();
-                    log::debug!("Activity producer for {source} exited with error {e}.");
-                } else {
-                    log::debug!("Activity producer for {source} completed successfully.");
-                }
-            });
+        // in order to choose a deterministic destination is necessary to sort the nodes by its public key
+        executors.sort_by(|ex1, ex2| ex1.source_info.pubkey.cmp(&ex2.source_info.pubkey));
+        let now = SystemTime::now();
```
carlaKC, line 1071: Should be `self.clock.now()`

simln-lib/src/lib.rs
```diff
@@ -620,10 +621,42 @@ struct ExecutorKit {
...
+struct PaymentEvent {
+    pubkey: PublicKey,
+    absolute_time: SystemTime,
+    destination: NodeInfo,
+    capacity: Option<u64>,
+}
+
+struct PaymentEventPayload {
+    executor: ExecutorKit,
+    current_count: u64,
```
carlaKC, line 637: Thoughts on making this an `Option`? Since we don't need it for payments that aren't count-bounded.

simln-lib/src/lib.rs
```diff
@@ -1031,257 +1055,191 @@ impl<C: Clock + 'static> Simulation<C> {
...
+
+                                            log::debug!("Send event payment for {source} completed successfully.");
+                                        }
+                                    });
                                 }
                             }
-                        }
+                        },
+                        None => {
+                            break
```
carlaKC, line 1170: Let's add a log here - can see it being useful in debugging

simln-lib/src/lib.rs
```diff
@@ -1031,257 +1055,191 @@ impl<C: Clock + 'static> Simulation<C> {
...
+                            let pe_output_sender = output_sender.clone();
+                            // Only proceed with a payment if the amount is non-zero, otherwise skip this round. If we can't get
+                            // a payment amount something has gone wrong (because we should have validated that we can always
+                            // generate amounts), so we exit.
+                            let amount = payload.executor.payment_generator.payment_amount(capacity).map_err(SimulationError::PaymentGenerationError)?;
+                            if amount == 0 {
+                                log::debug!(
+                                    "Skipping zero amount payment for {source} -> {}.", destination
+                                );
+                                generate_payment(&mut heap, pubkey, payload.current_count, &mut payment_event_payloads, SystemTime::now()).await?;
```
carlaKC, line 1139: None of the `?` errors will be caught now, because we're spawning this in a task? I believe the loop will exit but we won't see the error.

I think that we should pull this out into a function and then the spawn callsite is responsible for checking whether it exists with an error and triggering shutdown appropriately if it does (like the tasks spawned in `internal_run`).

simln-lib/src/lib.rs
```diff
@@ -1031,257 +1055,191 @@ impl<C: Clock + 'static> Simulation<C> {
...
+
+                            // pe: produce events
+                            let pe_shutdown = shutdown.clone();
+                            let pe_clock = clock.clone();
+                            let source = payload.executor.source_info.clone();
+                            let pe_output_sender = output_sender.clone();
+                            // Only proceed with a payment if the amount is non-zero, otherwise skip this round. If we can't get
+                            // a payment amount something has gone wrong (because we should have validated that we can always
+                            // generate amounts), so we exit.
+                            let amount = payload.executor.payment_generator.payment_amount(capacity).map_err(SimulationError::PaymentGenerationError)?;
```
carlaKC, line 1134: Why not generate this along with the rest of the details we push onto the heap?

Doesn't necessarily have to change, just interested in motivation.

carlaKC (commented):
Haven't reviewed the tests yet, but looking semantically correct now 👍 

No real need for fixups here since this is one commit - feel free to squash what you have and just force push changes, there isn't a commit structure to preserve in this case.

simln-lib/src/lib.rs
f3r10: The `generate_payment` is not part of the `task.spawn`; it is located just before. 
As I understand, the `select!` macro does not spawn tasks :thinking:

simln-lib/src/lib.rs
```diff
@@ -1031,257 +1055,191 @@ impl<C: Clock + 'static> Simulation<C> {
...
+                            destination,
+                            capacity
+                        })) => {
+                            let internal_now = SystemTime::now();
+                            let wait_time = match absolute_time.duration_since(internal_now) {
+                                Ok(elapsed) => {
+                                    // the output of duration_since is not perfectly round affecting the waiting time
+                                    // and as consequence the results are not deterministic; for this reason
+                                    // it is necessary to round the output.
+                                    Duration::from_secs(elapsed.as_secs_f64().round() as u64)
```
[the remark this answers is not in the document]
f3r10: it is also available `as_millis_f64()` but is not stable yet
```rust
    #[unstable(feature = "duration_millis_float", issue = "122451")]
    #[must_use]
    #[inline]
    #[rustc_const_unstable(feature = "duration_consts_float", issue = "72440")]
    pub const fn as_millis_f64(&self) -> f64 {
    ...
```

simln-lib/src/lib.rs
carlaKC: I mean that we spawn all of this code in a task on line 1088. Any `?` that we use in the block of code that is spawned in the task is just swallowed, so we can silently fail rather than printing out the error.

To try this yourself, add a `return(Err())` anywhere in this select - the program shuts down but we can't see the error.

simln-lib/src/lib.rs
f3r10: I think it is not necessary. It will make the code unnecessarily more verbose. Right now, the counter is only used here. 
```rust
if let Some(c) = payment_tracker.executor.payment_generator.payment_count() {
    if c == payment_tracker.payments_completed {
        log::info!( "Payment count has been met for {}: {c} payments. Stopping the activity.", payment_tracker.executor.source_info);
        continue
    }
}
```
And also in `get_payment_delay`, which does not accept an  `option`

simln-lib/src/lib.rs
f3r10: Oh right, I see it now :sweat_smile:

simln-lib/src/lib.rs
```diff
@@ -1550,20 +1444,183 @@ async fn track_payment_result(
...
+
+    let amount = payment_tracker
+        .executor
+        .payment_generator
+        .payment_amount(capacity)
+        .map_err(SimulationError::PaymentGenerationError)?;
+
+    payments_tracker
+        .entry(payment_tracker.executor.source_info.pubkey)
+        .and_modify(|p| p.payments_completed = current_count);
```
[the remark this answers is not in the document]
f3r10: I agree, I will make the change

simln-lib/src/lib.rs
```diff
@@ -780,20 +819,12 @@ impl<C: Clock + 'static> Simulation<C> {
...
-            event_sender.clone(),
-            &self.tasks,
-        );
 
         // Next, we'll spin up our actual producers that will be responsible for triggering the configured activity.
         // The producers will use their own TaskTracker so that the simulation can be shutdown if they all finish.
         let producer_tasks = TaskTracker::new();
         match self
-            .dispatch_producers(activities, consumer_channels, &producer_tasks)
+            .dispatch_producers(activities, event_sender, &producer_tasks)
```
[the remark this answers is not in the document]
f3r10: Yes, you are right, it is not necessary anymore.

Issue #269, opened 2025-05-16
Feature Request: Ability to exclude sim_node from random activity


carlaKC:
**Describe desired feature**

When running `sim-ln` in simulated network mode, you have to specify all the channels in the fake network that you're setting up. We infer our list of nodes from this set of channels, since each includes a pubkey.

If you're running with random activity (which is implied by the lack of defined activity), this means that each of the nodes in your network will be sending/receiving. We should have a way to be able to list a node in our network but not send/receive payments for it.

Note: this doesn't apply in the case where we're running on real nodes. If you want to exclude a node, you can just not list it in your `nodes` (because we fetch the graph from the node itself). To do this for a simulated network, you'd need to remove the node from your graph completely by removing all of its channels.

**Use case for feature**

This is useful for me in channel jamming simulations, but more generally helpful for users looking to simulate networks that have some nodes that only route.

I think that the cleanest way to express this would be to add an `exclude` list to `SimParams`:
* Only valid when `sim_network` is also set
* Provides a `Vec<Pubkey>` of node that shouldn't send/receive
* Update `Simulation.run` to accept `[ActivityDefinition] or [Exclude List]` (probably represent with an enum?)

**Would you like to contribute code for this feature?**

I will review!

PR #278, opened 2025-06-04
sim-lib: exclude nodes from random activity

elnosh:
Closes https://github.com/bitcoin-dev-project/sim-ln/issues/269

Added an `ActivityOpts` enum to pass `DefinedActivity` or `Random` and optionally specify nodes to be excluded in the random variant.

### DIFF
--- a/sim-cli/src/parsing.rs
+++ b/sim-cli/src/parsing.rs
@@ -13,7 +13,7 @@ use simln_lib::{
     ActivityDefinition, Amount, Interval, LightningError, LightningNode, NodeId, NodeInfo,
     Simulation, SimulationCfg, WriteResults,
 };
-use simln_lib::{ShortChannelID, SimulationError};
+use simln_lib::{ActivityOpts, ShortChannelID, SimulationError};
 use std::collections::HashMap;
 use std::fs;
 use std::path::PathBuf;
@@ -123,6 +123,13 @@ impl Cli {
             ));
         }
 
+        if sim_params.sim_network.is_empty() && !sim_params.exclude.is_empty() {
+            return Err(anyhow!(
+                "List of nodes to exclude from sending/receiving
+                    in random activity is only allowed on a simulated network"
+            ));
+        }
+
         Ok(())
     }
 }
@@ -135,6 +142,8 @@ pub struct SimParams {
     pub sim_network: Vec<NetworkParser>,
     #[serde(default)]
     pub activity: Vec<ActivityParser>,
+    #[serde(default)]
+    pub exclude: Vec<PublicKey>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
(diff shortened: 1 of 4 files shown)


simln-lib/src/lib.rs
```diff
@@ -173,6 +173,17 @@ pub type Amount = ValueOrRange<u64>;
...
+/// Payment activity options for running the simulation.
+pub enum ActivityOpts {
+    /// Defined payment activity provided by user that will be executed in the simulation.
+    DefinedActivity(Vec<ActivityDefinition>),
+    /// Generate random activity. Can optionally exclude nodes provided in ExcludeList.
+    Random(Option<ExcludeList>),
+}
+
+/// Nodes to exclude from sending/receiving in random activity.
+pub type ExcludeList = Vec<PublicKey>;
```
carlaKC, line 185: I realize this is written up in the original issue, but wondering whether we want to have this `ExcludeList` in the library? Could be simpler to just have this enum and require that random provides a list of pubkeys that should be included.

The caller can just do that filtering by itself and pass the list through to the simulator. WDYT?

carlaKC: I like the alias, was more wondering about whether an include/exclude list is a good API.

>  I'd prefer an optional exclusion list rather than inclusion.

👍

sim-cli/src/parsing.rs
```diff
@@ -123,6 +123,13 @@ impl Cli {
             ));
         }
 
+        if sim_params.sim_network.is_empty() && !sim_params.exclude.is_empty() {
+            return Err(anyhow!(
+                "List of nodes to exclude from sending/receiving
+                    in random activity is only allowed on a simulated network"
```
carlaKC, line 129: nit: note that user can just remove a node from `nodes` if they don't want it included for regular operation

sim-cli/src/parsing.rs
```diff
@@ -554,7 +567,11 @@ pub async fn get_validated_activities(
     clients: &HashMap<PublicKey, Arc<Mutex<dyn LightningNode>>>,
     nodes_info: HashMap<PublicKey, NodeInfo>,
     activity: Vec<ActivityParser>,
-) -> Result<Vec<ActivityDefinition>, LightningError> {
+) -> Result<ActivityOpts, LightningError> {
+    if activity.is_empty() {
```
carlaKC, line 571: It's odd to me to split random activity `ActivityOpts` between here and the call site - either this function is responsible for assembling all of the possible `ActivityOpts` or it's just for defined (non-random) activities, now sure what the motivation to split it like this is?

sim-cli/src/parsing.rs
```diff
@@ -123,6 +123,13 @@ impl Cli {
             ));
         }
 
+        if sim_params.sim_network.is_empty() && !sim_params.exclude.is_empty() {
```
carlaKC, line 126: Should also check that `activity` + `exclude` aren't specified at the same time?

carlaKC (commented):
Taking a closer look at this, I don't think that `Simualtion` should be aware of an `ExcludeList`:

It seems like a clean division that `Simulation` knows about about a list of nodes that it should be running activity on, and those nodes have a view of the full graph.

It's our parsing module's job to take our simulated network and filter out the ones we don't want to run random activity on (as that's really just a consequence of how we've chosen to express the graph). It can do this by filtering the `ln_node_from_graph` node list, and passing those nodes to the simulator. Our `SimGraph` will still work in the background, and the `Simulator` doesn't need to know about this detail at all.

My bad writing that direction up in the original issue, doesn't feel like the right way to do it now that I've seen it in code. Sorry it took seeing the code to realize this!

elnosh:
closing in favor of https://github.com/bitcoin-dev-project/sim-ln/pull/280. Thanks for the suggestion! @carlaKC

PR #282, opened 2025-06-11
add alias to simulated network

elnosh:
adds an alias to a simulated network that can be optionally set in the `sim_network`. If not set, it defaults to an empty string.


simln-lib/src/sim_node.rs
```diff
@@ -1007,17 +1004,18 @@ impl SimGraph {
 }
 
 /// Produces a map of node public key to lightning node implementation to be used for simulations.
-pub async fn ln_node_from_graph(
+pub async fn ln_node_from_graph<'a>(
+    nodes_info: &HashMap<PublicKey, NodeInfo>,
```
carlaKC, line 1008: Do we need to pass `nodes_info` in here? Could use `graph.lock().await.list_nodes()`

carlaKC (approved):
✅

PR #283, opened 2025-06-11
pass cfg to `create_simulation_with_network` 

elnosh:
I think it'd be easier to use `create_simulation_with_network` outside of `sim-cli` if it took a `SimulationCfg` instead of the `cli`.


carlaKC (approved):
LGTM!

PR #284, opened 2025-06-13
Fix: Remove Unnecessary &mut self Requirements from LightningNode Trait

### LINKED ISSUE #264: simln-lib: refactor `LightningNode` trait to not use mutable references
Currently, the methods in the `LightningNode` trait require mutable references, including ones that only retrieve information e.g `async fn get_network(&mut self) -> Result<Network, LightningError>;` This is because internally the implementations (lnd, cln, etc) have `Client`s that require mutable references to make their calls. However this looks like leaking from the implementation to the trait as a method to list something should not need a mutable reference. 

Internally, `Client`s could instead be wrapped in a `Mutex` and then methods in trait can just have a reference.

mubarak23:
closes #264 
This PR refactors the LightningNode trait and its implementations to remove unnecessary &mut self references for methods that do not mutate internal state.

### Motivation
Previously, several trait methods like get_network and list_channels required &mut self, even though they only retrieved information. This requirement leaked internal implementation details—specifically, that the underlying client field required a mutable reference to make gRPC or HTTP calls.

This design was misleading and unnecessarily restricted usage of the trait, e.g. preventing multiple read-only operations from being performed concurrently on the same node.

### Changes
Wrapped client fields (e.g. NodeClient, EclairClient, etc.) inside tokio::sync::Mutex.

Updated all client usages to acquire a mutable lock via let mut client = self.client.lock().await;.

Updated trait method signatures to require &self instead of &mut self where applicable.

Ensured thread-safe, concurrent-friendly behavior without sacrificing correctness or clarity.

### DIFF
--- a/simln-lib/src/lib.rs
+++ b/simln-lib/src/lib.rs
@@ -341,7 +341,7 @@ pub trait LightningNode: Send {
     async fn get_node_info(&self, node_id: &PublicKey) -> Result<NodeInfo, LightningError>;
     /// Lists all channels, at present only returns a vector of channel capacities in msat because no further
     /// information is required.
-    async fn list_channels(&mut self) -> Result<Vec<u64>, LightningError>;
+    async fn list_channels(&self) -> Result<Vec<u64>, LightningError>;
     /// Get the network graph from the point of view of a given node.
     async fn get_graph(&self) -> Result<Graph, LightningError>;
 }
--- a/simln-lib/src/cln.rs
+++ b/simln-lib/src/cln.rs
@@ -112,7 +112,7 @@ impl ClnNode {
     /// Fetch channels belonging to the local node, initiated locally if is_source is true, and initiated remotely if
     /// is_source is false. Introduced as a helper function because CLN doesn't have a single API to list all of our
     /// node's channels.
-    async fn node_channels(&mut self, is_source: bool) -> Result<Vec<u64>, LightningError> {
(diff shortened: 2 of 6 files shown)


simln-lib/src/lib.rs
```diff
@@ -324,26 +324,26 @@ pub trait LightningNode: Send {
...
+        &self,
         hash: &PaymentHash,
         shutdown: Listener,
     ) -> Result<PaymentResult, LightningError>;
     /// Gets information on a specific node.
-    async fn get_node_info(&mut self, node_id: &PublicKey) -> Result<NodeInfo, LightningError>;
+    async fn get_node_info(&self, node_id: &PublicKey) -> Result<NodeInfo, LightningError>;
     /// Lists all channels, at present only returns a vector of channel capacities in msat because no further
     /// information is required.
     async fn list_channels(&mut self) -> Result<Vec<u64>, LightningError>;
```
elnosh, line 344: should also be able to make this just take `&self`

elnosh (approved):
think should be good to squash

carlaKC:
Small nit for when you do squash: take a look at [commit structure](https://github.com/bitcoin-dev-project/sim-ln/blob/main/docs/CONTRIBUTING.md#commit-history) in contributing guide - we use an odd format (historical reasons), would be nice to align with that while you're at it.

carlaKC (approved):
tACK, nice cleanup 🥦

PR #287, opened 2025-06-24
sim-lib: add custom records to send_to_route

carlaKC:
The `send_to_route` method is an advanced-user feature that allows sending custom payments inside of simln. This PR adds the ability to attach your own `custom_records` to the first hop of the payment, useful for cases where you don't always want to use the default ones configured in `SimGraph`. 

For example, this is useful in channel jamming simulations where an attacker may want to set their own values for accountability (formerly endorsement).


simln-lib/src/sim_node.rs
```diff
@@ -529,14 +530,18 @@ impl<'a, T: SimNetwork> SimNode<'a, T> {
         }
     }
 
-    /// Dispatches a payment to a specified route.
+    /// Dispatches a payment to a specified route. If `custom_records` is `Some`, they will be attached to the outgoing
+    /// update_add_htlc, otherwise [`SimGraph::default_custom_records`] will be used if `None`. If default custom records
+    /// are configured, but you want to send a payment without them specify `Some(CustomRecords::default()`.
```
elnosh, line 535: missing closing `)` in `Some(CustomRecords::default()`

elnosh (commented):
LGTM

simln-lib/src/sim_node.rs
carlaKC: [added](https://github.com/bitcoin-dev-project/sim-ln/compare/51fa6fe7b8c3ec885a716448799df592a86488ae..c9799f8c1bd87340094f7bc318ce128fcb7eaca7) 👍

PR #290, opened 2025-06-24
return `SimNode`s when creating simulated network

elnosh:
return `SimNode`s when creating a simulated network. 

This required changing `SimNode` to not use any lifetimes because `create_simulation_with_network` will return a hashmap with the concrete `SimNode`s but internally pass them as `dyn LightningNode` for the `Simulation`.


simln-lib/src/sim_node.rs
```diff
@@ -487,38 +487,40 @@ pub trait SimNetwork: Send + Sync {
     fn list_nodes(&self) -> Vec<NodeInfo>;
 }
 
+type LdkNetworkGraph = NetworkGraph<Arc<WrappedLog>>;
```
carlaKC, line 490: nitty nit: would have moved this refactor into its own commit for the change - single responsibility commits nice where we can get them

carlaKC (approved):
🏁

PR #294, opened 2025-07-01
Fix Critical Error handling in HTLC interception

### LINKED ISSUE #272: simln-lib: handle graceful shutdown during htlc interception
as a follow-up from https://github.com/bitcoin-dev-project/sim-ln/pull/261, if a `CriticalError` occurs during htlc interception it immediately returns and triggers a simulation shutdown. 

Instead of immediately returning, we can trigger a shutdown signal to other interceptors, if any, and wait for them to finish.

mubarak23:
closes #272 

TODO: This needed to be done.

Currently, when a CriticalError occurs during HTLC interception in [handle_intercepted_htlc](https://github.com/bitcoin-dev-project/sim-ln/blob/c4059cd30353b03c0d7c495121ef7ce9e698af47/simln-lib/src/sim_node.rs#L854), the function immediately returns without waiting for other running interceptors to complete. This can leave interceptors in an inconsistent state and prevent proper cleanup.

## Changes
- Store critical errors instead of immediately returning, When a CriticalError occurs, store it in a local variable rather than returning immediately.
- Send shutdown signal to all other running interceptors when a critical error is encountered
- Added critical_error variable to store any CriticalError that occurs
- Modified the Err(e) match arm to store the error and trigger shutdown instead of immediate return

### DIFF
--- a/simln-lib/src/sim_node.rs
+++ b/simln-lib/src/sim_node.rs
@@ -880,6 +880,8 @@ async fn handle_intercepted_htlc(
     // the HTLC. If any of the interceptors did return an error, we send a shutdown signal
     // to the other interceptors that may have not returned yet.
     let mut interceptor_failure = None;
+    let mut critical_error = None;
+
     'get_resp: loop {
         tokio::select! {
             res = intercepts.join_next() => {
@@ -917,8 +919,11 @@ async fn handle_intercepted_htlc(
                         interceptor_failure = Some(fwd_error);
                         interceptor_trigger.trigger();
                     },
+                    // Interceptor returned a CriticalError, Trigger a shutdown and store the error
+                    // to return after all interceptors have finished.
                     Err(e) => {
-                        return Err(e);
+                        critical_error = Some(e);
+                        interceptor_trigger.trigger();
                     },
                 }
             }
@@ -931,6 +936,11 @@ async fn handle_intercepted_htlc(
                 break 'get_resp
             }
         }
+    }
+
+    // if we have a critical error, returned it after all interceptors have finished.
+    if let Some(e) = critical_error {
+        return Err(e);
     }
 
     if let Some(e) = interceptor_failure {
--- a/simln-lib/src/sim_node.rs
+++ b/simln-lib/src/sim_node.rs
@@ -2480,6 +2480,38 @@ mod tests {
         assert!(response.unwrap().unwrap() == CustomRecords::from([(1000, vec![1])]));
     }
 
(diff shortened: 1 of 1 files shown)


elnosh (commented):
LGTM. Thanks for the change.
Could you update the TODO comment at the top saying this needed to be done.

carlaKC:
Change looks good - would be nice to add test coverage for this shutdown case

simln-lib/src/sim_node.rs
```diff
@@ -2470,6 +2480,38 @@ mod tests {
...
+            &interceptors,
+            interceptor_trigger,
+            shutdown_listener,
+        )
+        .await;
+
+        assert!(result.is_err());
+        let err_string = format!("{:?}", result.unwrap_err());
+        assert!(err_string.contains("critical failure"));
+    }
```
elnosh, line 2513: hmm this checks that the returned error is the one returned from interceptor but not sure if @carlaKC meant to add a case where other interceptors received a shutdown signal if a critical error was returned another one. Although we are just mocking interceptors so I'm babbling here.

elnosh (commented):
heh sorry, with this ` Could you update the TODO comment at the top saying this needed to be done.` I meant update the TODO comment in `handle_intercepted_htlc`

simln-lib/src/sim_node.rs
carlaKC: > to add a case where other interceptors received a shutdown signal if a critical error was returned from another one

This is what I meant. 

Sadly I don't think that we'll be able to mock the interceptors to check that each of them trigger shutdown. Was thinking of a very simple interceptor that just blocks until we get shutdown and then we add a timeout on the test to make sure that nothgin is blocking.

mubarak23: so there is no way to get an interceptor to blocked until we got a shutdown

carlaKC: I'm not sure what you mean? You can implement an interceptor that listens on the shutdown listener for the purpose of the test and that should block?

We wait for all the interceptors to exit when we get the shutdown signal, so they should block if the trigger isn't hit by a critical error.

mubarak23: ok, 
i will probably look into how mock interceptor are done i.e 

`create_intercept_request`

then i will implement an interceptor that listens on the shutdown listener and mock it

mubarak23:
> @mubarak23 any updates here? Will be closing due to inactivity if review is not addressed soon.

i was not able to fully implement interceptor, but i will share what i have and get pointers in the right direction

PR #295, opened 2025-07-02
sim-cli: exclude nodes from simulation but not from returned map

carlaKC:
When we want to exclude nodes from random activity generation, we may still want to include them in our map of network nodes so that the caller can use any node in the network.


elnosh (approved):
:thumbsup:

Issue #111, opened 2023-09-21
Design Doc: Generate Random Activity


carlaKC:
This issue is the proposed design document for #72.

## Random Activity Generation

When we're looking to generate random activity for some topology, there are a few values that we need to pick: 
- How often do we fire payments?
- How do we choose source and destination nodes?
- How do we choose payment amounts?

## Goals
- Use the underlying graph topology to inform the types of payment flows generated for the network. 
- Introduce randomness to payment timing, amount and destination. 
- Scale payment activity by node size, meaning that nodes with more liquidity deployed will send payments more frequently. 

## User Configuration
Goals for user configuration of this functionality: 
1. Quickstart: just specify nodes and generate random activity for all nodes. 
2. Combo: allow combinations of `activity_description` and randomly generated activity. 
3. Subset: limit random activity to a subset of `nodes`.

This covers a reasonably large range of use cases without blowing up configuration complexity too much.

Proposed handling: 
- If no `activity_description` is present: run random activity with all nodes.
- Add a `random_activity` section that is just a list of pubkeys.
  - May be present alongside a `activity_description`
  - If specified, only run random activity on the nodes listed.

So, for our original goal use cases we'd have: 
1. Quickstart: `nodes` -> runs random activity on all nodes.
2. Combo: `nodes`, `activity_description`, `random_activity` = [`A`, `B`] -> runs activity descriptions *and* random activity on `A` and `B`.
3. Subset: `nodes`, `random_activity` = [`B`, `C`] -> runs random activity on a subset of nodes.

## How often do we fire payments?

We can conceptually think about the level of network activity as a question of capital efficiency: "In a month of activity, how many times over do nodes in the network send its total deployed capital?"

- If it's 0.5x, the network sends half its total capacity in payments in a month.
- If it's 1x, the network sends exactly its total capacity in payments in a month. 
- If it's 3x, the network sends three times its total capacity in payments in a month.

For our simulation, we will consider the following: 
- `channel_capacity`: the sum of the liquidity of a node’s announced channels / 2 (so that we don’t double count capacity for channel counterparties).
- `multiplier`: the multiplier on `channel_capacity` that determines the total sent in a month.
- `expected_payment_size`: 100_000 satoshis, the expected payment size in the network (further specified below).

For each node in the network we calculate the total number of events that we expect in a month: 
`total_events` = `(channel_capacity * multiplier) / expected_payment_size` 

[...]

PR #301, opened 2025-08-14
Decrease Variance of LogNormal to Converge on Expected Value Sooner

carlaKC:
See rationale [here](https://github.com/bitcoin-dev-project/sim-ln/issues/111#issuecomment-3185472086).

This PR makes a fix to the way we pick our `LogNormal` distribution's parameters which results in our getting to an average of our `expected_payment_amount` much sooner.

As this changes the way we pick payments, the hard-coded ordering of some of our tests has to change.
While I'm here, I also make some changes to long-running unit tests to speed them up a bit.

### DIFF
--- a/simln-lib/src/lib.rs
+++ b/simln-lib/src/lib.rs
@@ -1592,7 +1592,7 @@ async fn track_payment_result(
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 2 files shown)


f3r10 (approved):
Tested ACK

carlaKC:
Not actively working on this, warrants further investigation IMO.

PR #302, opened 2025-08-18
tests: speed up ordering tests by using simulation clock

carlaKC:
Pulling this out of #301 to speed up the tests so we don't have to wait ~1min for each run, this change gets full test suite back down to ~5s.

### DIFF
--- a/simln-lib/src/lib.rs
+++ b/simln-lib/src/lib.rs
@@ -1592,7 +1592,7 @@ async fn track_payment_result(
 
 #[cfg(test)]
 mod tests {
-    use crate::clock::SystemClock;
+    use crate::clock::{Clock, SimulationClock, SystemClock};
     use crate::test_utils::{MockLightningNode, TestNodesResult};
     use crate::{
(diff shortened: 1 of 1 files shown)


f3r10:
I found the issue.
The problem is with the rounding of the wait_time.
There are times (after 1000 times, probably) that instead of getting  `1.543701036s`, the wait_time is `1.495485804s` 
Rounding `1.543701036s` is `2`, and rounding `1.495485804s` is  `1`. Becuase of this reason the determined order fails.

I think one possible solution would be to round to the first decimal place.

f3r10:
the solution that I found was to round the wait_time to one decimal.
```rust
fn round_duration_to_one_decimal(duration: Duration) -> f64 {
    let millis = duration.as_millis();
    let rounded_millis = (millis + 50) / 100 * 100;
    rounded_millis as f64 / 1000.0
}
```
image

carlaKC:
I'm not actively working on this - would be great if somebody wanted to pick it up!

Issue #304, opened 2025-09-15
Bug: Excluded channels contribute to random activity generation


carlaKC:
**Describe the bug**

In `create_simulation_with_network` we use our `exclude` list to remove all nodes that we don't want sending/receiving payments in random activity generation from the list of nodes that we pass the simulator. We still include their channels in the graph, because we want them to be able to _forward_ payments.

In our random activity generation, we use the channel balance of a node to determine how much it will send and receive. Although we're excluding the nodes from the simulation, their channel balances are still being considered when we calculate how much their peers should send/receive.

We should exclude these channels from their peers balance calculations somehow (possibly just with a marker field on `SimulatedChannel`?).

**To Reproduce**

Given some network `A-B-C` where all channels are the same size. Add `A` to the exclude list, and note in the logs that B's total channel capacity is more than `C`. If we were properly ignoring `A`, then `B` and `C` would have the same capacity.

PR #305, opened 2025-09-18
simln-lib: exclude capacity from excluded nodes list

elnosh:
fixes https://github.com/bitcoin-dev-project/sim-ln/issues/304

added an `exclude_capacity` field to `SimulatedChannel` to then exclude that channel capacity for the node when creating the `SimGraph`

### DIFF
--- a/sim-cli/src/parsing.rs
+++ b/sim-cli/src/parsing.rs
@@ -172,17 +172,6 @@ pub struct NetworkParser {
     pub node_2: ChannelPolicy,
 }
 
-impl From<NetworkParser> for SimulatedChannel {
-    fn from(network_parser: NetworkParser) -> Self {
-        SimulatedChannel::new(
-            network_parser.capacity_msat,
-            network_parser.scid,
-            network_parser.node_1,
-            network_parser.node_2,
-        )
-    }
-}
-
 /// Data structure used to parse information from the simulation file. It allows source and destination to be
 /// [NodeId], which enables the use of public keys and aliases in the simulation description.
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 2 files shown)


sim-cli/src/parsing.rs
```diff
@@ -277,7 +278,14 @@ pub async fn create_simulation_with_network(
...
         .clone()
         .into_iter()
-        .map(SimulatedChannel::from)
+        .map(|channel| {
+            let exclude_capacity = exclude.contains(&channel.node_1.pubkey)
+                || exclude.contains(&channel.node_2.pubkey);
+
+            let mut sim_channel = SimulatedChannel::from(channel);
+            sim_channel.exclude_capacity = exclude_capacity;
+            sim_channel
```
carlaKC, line 287: nit: We could give up on the strict API of `From` and just use `new` directly to shorten this a bit?

simln-lib/src/sim_node.rs
```diff
@@ -319,6 +319,7 @@ pub struct SimulatedChannel {
     short_channel_id: ShortChannelID,
     node_1: ChannelState,
     node_2: ChannelState,
+    pub exclude_capacity: bool,
```
carlaKC, line 322: Another point for setting this in `new`, wouldn't need to make one field `pub`.

Also probably needs a comment to explain what we're excluding capacity from.

simln-lib/src/sim_node.rs
```diff
@@ -1062,16 +1065,18 @@ impl SimGraph {
...
-            for info in [&channel.node_1.policy, &channel.node_2.policy] {
-                match nodes.entry(info.pubkey) {
-                    Entry::Occupied(o) => o.into_mut().1.push(channel.capacity_msat),
-                    Entry::Vacant(v) => {
-                        v.insert((
-                            node_info(info.pubkey, info.alias.clone()),
-                            vec![channel.capacity_msat],
-                        ));
-                    },
+            if !channel.exclude_capacity {
```
carlaKC, line 1068: Could we add a very simple test that:
- Creates a graph with an excluded channel
- Creates `ln_node_from_graph`
- Asserts that capacities don't include the excluded channel

Would be nice to have the whole flow covered.

carlaKC (commented):
Just nits and a test-beg 🙏

sim-cli/src/parsing.rs
elnosh: he, cool. I did the ugly `pub exclude_capacity` as you pointed below to keep the `From`. Changed to use `new` :thumbsup:

simln-lib/src/sim_node.rs
elnosh: added test for the the simple case.

> Would be nice to have the whole flow covered.

For this, do you mean flow from creating the simulation with `create_simulation_with_network` and nodes balances returned from that?

PR #306, opened 2025-09-23
return all `SimNode`s from `ln_node_from_graph`

elnosh:
Changes:
- repurpose `list_channels` to `channel_capacities` in the `LightningNode` trait since this call was only used to determine how much the node will send per month.
- fixes bug introduced in https://github.com/bitcoin-dev-project/sim-ln/commit/39257da6cc077c34992c7962c4c3909b8200e247 where it would not return all the nodes (the ones with excluded channels). Even if we are excluding channels from nodes, we still want to return `SimNode`s so that a user could use them.

### DIFF
--- a/simln-lib/src/sim_node.rs
+++ b/simln-lib/src/sim_node.rs
@@ -477,6 +477,7 @@ impl SimulatedChannel {
 
 /// SimNetwork represents a high level network coordinator that is responsible for the task of actually propagating
 /// payments through the simulated network.
+#[async_trait]
 pub trait SimNetwork: Send + Sync {
     /// Sends payments over the route provided through the network, reporting the final payment outcome to the sender
     /// channel provided.
@@ -490,7 +491,7 @@ pub trait SimNetwork: Send + Sync {
     );
 
     /// Looks up a node in the simulated network and a list of its channel capacities.
-    fn lookup_node(&self, node: &PublicKey) -> Result<(NodeInfo, Vec<u64>), LightningError>;
+    async fn lookup_node(&self, node: &PublicKey) -> Result<(NodeInfo, Vec<u64>), LightningError>;
     /// Lists all nodes in the simulated network.
     fn list_nodes(&self) -> Vec<NodeInfo>;
 }
@@ -794,11 +795,16 @@ impl<T: SimNetwork, C: Clock> LightningNode for SimNode<T, C> {
     }
 
     async fn get_node_info(&self, node_id: &PublicKey) -> Result<NodeInfo, LightningError> {
-        Ok(self.network.lock().await.lookup_node(node_id)?.0)
+        Ok(self.network.lock().await.lookup_node(node_id).await?.0)
     }
 
     async fn channel_capacities(&self) -> Result<u64, LightningError> {
-        let channels = self.network.lock().await.lookup_node(&self.info.pubkey)?;
+        let channels = self
+            .network
+            .lock()
+            .await
+            .lookup_node(&self.info.pubkey)
+            .await?;
         Ok(channels.1.iter().sum())
     }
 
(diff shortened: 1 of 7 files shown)


carlaKC (commented):
Feel like we haven't quite hit the right API for this, but stuggling to think of a better approach - let's chat about it irl.

simln-lib/src/sim_node.rs
```diff
@@ -797,8 +797,9 @@ impl<T: SimNetwork, C: Clock> LightningNode for SimNode<T, C> {
         Ok(self.network.lock().await.lookup_node(node_id)?.0)
     }
 
-    async fn list_channels(&self) -> Result<Vec<u64>, LightningError> {
-        Ok(self.network.lock().await.lookup_node(&self.info.pubkey)?.1)
+    async fn channel_capacities(&self) -> Result<u64, LightningError> {
+        let channels = self.network.lock().await.lookup_node(&self.info.pubkey)?;
+        Ok(channels.1.iter().sum())
```
carlaKC: Here we should throw a node not found error if one is not in the list.

We can filter the channels by exclude to get the balance, so that we get a zero balance.

simln-lib/src/sim_node.rs
```diff
@@ -798,7 +798,16 @@ impl<T: SimNetwork, C: Clock> LightningNode for SimNode<T, C> {
...
     async fn channel_capacities(&self) -> Result<u64, LightningError> {
-        let channels = self.network.lock().await.lookup_node(&self.info.pubkey)?;
+        let channels = match self.network.lock().await.lookup_node(&self.info.pubkey) {
+            Ok(channels) => channels,
+            Err(e) => match e {
+                // This is a bit weird in that a user could have a SimNode but if it was excluded
+                // from sending payments, it won't be returned from the lookup_node call. In that
+                // case, we return a capacity of 0 instead of returning a node not found error.
+                LightningError::GetNodeInfoError(_) => return Ok(0),
+                _ => return Err(e),
```
carlaKC, line 808: No implicit zero, fail if we don't find the node like in the previous commit.

simln-lib/src/sim_node.rs
```diff
@@ -1102,13 +1111,46 @@ pub async fn ln_node_from_graph<C: Clock>(
...
     clock: Arc<C>,
 ) -> Result<HashMap<PublicKey, Arc<Mutex<SimNode<SimGraph, C>>>>, LightningError> {
-    let mut nodes: HashMap<PublicKey, Arc<Mutex<SimNode<SimGraph, C>>>> = HashMap::new();
+    let network_graph = routing_graph.read_only();
+    let mut nodes: HashMap<PublicKey, Arc<Mutex<SimNode<SimGraph, C>>>> =
+        HashMap::with_capacity(network_graph.nodes().len());
+
+    // We want to return all the nodes even if they were excluded in the SimGraph. So we iterate
+    // over the routing_graph that should include all nodes.
+    for node in network_graph.nodes().unordered_iter() {
```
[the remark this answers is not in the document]
carlaKC: This whole function can probably go back to what it was.

carlaKC (commented):
tl;dr:

- SimGraph: has all of the nodes in the network, some have channels that are marked as exclude. We filter our capacity by the exclude field.
- RoutingGraph: has all of the nodes, that's it.
- `ln_node_from_graph`: returns the full list of nodes, so we can use any one we want for misc purposes.
- `Simulation::new`: delete any nodes that we want to exclude because we don't want them sending/receiving. 

Removing from the list of nodes that we give the simulator stops them from being eligible for sending/receiving payments.  Having the exclude field stop their peers from over estimating the amount they should send/receive, while still allowing us to forward payments over the excluded channels.

simln-lib/src/sim_node.rs
```diff
@@ -477,6 +477,7 @@ impl SimulatedChannel {
 
 /// SimNetwork represents a high level network coordinator that is responsible for the task of actually propagating
 /// payments through the simulated network.
+#[async_trait]
```
elnosh, line 480: had to make this an `async_trait` because of tokio mutex :( un-yay

elnosh:
addressed feedback - storing all nodes in `SimGraph` but excluding capacity based on exclude field in `SimulatedChannel`

simln-lib/src/sim_node.rs
```diff
@@ -1232,13 +1239,26 @@ impl SimNetwork for SimGraph {
...
-            None => Err(LightningError::GetNodeInfoError(
-                "Node not found".to_string(),
-            )),
-        }
+    async fn lookup_node(&self, node: &PublicKey) -> Result<(NodeInfo, Vec<u64>), LightningError> {
+        let node_info = match self.nodes.get(node) {
+            Some(node) => node.clone(),
+            None => {
+                return Err(LightningError::GetNodeInfoError(
+                    "Node not found".to_string(),
```
carlaKC, line 1247: nit: include pubkey of node for debugging

carlaKC (approved):

PR #307, opened 2026-04-08
Add support of ldk-node

### LINKED ISSUE #26: Feature: Multi-Implementation Support

randomlogin:
This PR adds support of ldk-node to sim-ln.

Closes: https://github.com/bitcoin-dev-project/sim-ln/issues/26

As ldk-node is not a daemon, before we start we parse the simulation config file and extract all ldk-nodes, build them and start. For that an additional chain source (bitcoind/esplora/electrum) is needed to start.

Before this PR can be merged, an update of used lighting libraries should be merged, as the current ldk-node has newer versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ lightning network development. It may be useful to you if you are:
 * LND ✅
 * CLN ✅
 * Eclair ✅️
-* LDK-node 🏗️
+* LDK-Server ✅
 
 See our [tracking issue](https://github.com/bitcoin-dev-project/sim-ln/issues/26)
 for updates on implementation support (contributions welcome!).
@@ -36,6 +36,7 @@ of the simulator uses keysend to execute payments, which must be enabled as foll
 * LND: `--accept-keysend`
 * CLN: enabled by default
 * Eclair: `-Declair.features.keysend=optional` (or `--features.keysend=optional` if you're using Polar)
+* LDK-Server: enabled by default via `spontaneous_send`
 
 NOTE: for CLN `keysend` to work with eclair, you need to add additional config to eclair:
 ```
@@ -98,6 +99,23 @@ The required access details will depend on the node implementation.
   "api_password": <password_to_authorize>
 }
 ```
+* LDK-Server:
+```
+{
+  "address": <ip:port or domain:port>,
+  "api_key": <hex_encoded_api_key>,
+  "cert": <path_to_tls_cert>
+}
+```
+The `api_key` is the raw bytes of `~/.ldk-server/<network>/api_key` hex-encoded.
+Unlike other backends, ldk-server does not require an `id` field — the node's
+public key and network are fetched automatically on startup.
+
+Note: ldk-server channels are **unannounced by default**. Pass `--announce-channel`
+to `open-channel` (and set `announcement_addresses` in the ldk-server config) to
+make channels public and visible to other nodes for routing. Also ldk-server enforces a minimum final CLTV expiry delta
+of **144 blocks** on inbound keysend payments. Ensure that sending nodes are configured with a sufficiently high final
+CLTV delta.
 
 Payment activity can be simulated in two different ways:
 * [Random activity](#setup---random-activity): generate random activity on the `nodes` provided, 
@@ -135,6 +153,11 @@ to send and receive payments when running with random activity.
       "base_url": "127.0.0.1:8286",
       "api_username": "",
       "api_password": "eclairpw"
+    },
+    {
+      "address": "localhost:3536",
+      "api_key": "ldk_server_hex_encoded_api_key",
+      "cert": "/path/tls.crt"
     }
   ]
 }
--- a/README.md
+++ b/README.md
@@ -104,13 +104,12 @@ The required access details will depend on the node implementation.
 {
   "address": <ip:port or domain:port>,
   "api_key": <hex_encoded_api_key>,
-  "cert": <path_to_tls_cert>,
-  "network": <bitcoin|testnet|signet|regtest>
+  "cert": <path_to_tls_cert>
 }
 ```
 The `api_key` is the raw bytes of `~/.ldk-server/<network>/api_key` hex-encoded.
 Unlike other backends, ldk-server does not require an `id` field — the node's
-public key is fetched automatically on startup.
+public key and network are fetched automatically on startup.
 
 Note: ldk-server channels are **unannounced by default**. Pass `--announce-channel`
 to `open-channel` (and set `announcement_addresses` in the ldk-server config) to
@@ -158,8 +157,7 @@ to send and receive payments when running with random activity.
     {
       "address": "localhost:3536",
       "api_key": "ldk_server_hex_encoded_api_key",
-      "cert": "/path/tls.crt",
-      "network": "signet"
+      "cert": "/path/tls.crt"
     }
   ]
 }
--- a/README.md
+++ b/README.md
@@ -45,11 +45,6 @@ NOTE: for CLN `keysend` to work with eclair, you need to add additional config t
 ```
 where N and M are numbers, and N must be larger than M (N must be 22, CLN's default, or more)
 
-NOTE: ldk-server enforces a minimum final CLTV expiry of 144 blocks on inbound keysend payments.
-If you are sending from LND or CLN, set their final CLTV delta to at least 150:
-* LND: `bitcoin.timelockdelta=150` in `lnd.conf`
-* CLN: `cltv-final=150` in `config`
-
 ## Getting Started
 
 Once you have all the pre-requisites installed, clone the repo:
@@ -116,6 +111,12 @@ The required access details will depend on the node implementation.
 The `api_key` is the raw bytes of `~/.ldk-server/<network>/api_key` hex-encoded.
 Unlike other backends, ldk-server does not require an `id` field — the node's
 public key is fetched automatically on startup.
+
+Note: ldk-server channels are **unannounced by default**. Pass `--announce-channel`
+to `open-channel` (and set `announcement_addresses` in the ldk-server config) to
+make channels public and visible to other nodes for routing. Also ldk-server enforces a minimum final CLTV expiry delta
+of **144 blocks** on inbound keysend payments. Ensure that sending nodes are configured with a sufficiently high final
+CLTV delta.
 
 Payment activity can be simulated in two different ways:
 * [Random activity](#setup---random-activity): generate random activity on the `nodes` provided, 
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ lightning network development. It may be useful to you if you are:
 * LND ✅
 * CLN ✅
 * Eclair ✅️
-* LDK-node 🏗️
+* LDK-server ✅
 
 See our [tracking issue](https://github.com/bitcoin-dev-project/sim-ln/issues/26)
 for updates on implementation support (contributions welcome!).
@@ -36,6 +36,7 @@ of the simulator uses keysend to execute payments, which must be enabled as foll
 * LND: `--accept-keysend`
 * CLN: enabled by default
 * Eclair: `-Declair.features.keysend=optional` (or `--features.keysend=optional` if you're using Polar)
+* LDK-server: enabled by default via `spontaneous_send`
 
 NOTE: for CLN `keysend` to work with eclair, you need to add additional config to eclair:
 ```
@@ -43,6 +44,11 @@ NOTE: for CLN `keysend` to work with eclair, you need to add additional config t
 -Declair.channel.fulfill-safety-before-timeout-blocks=M
 ```
 where N and M are numbers, and N must be larger than M (N must be 22, CLN's default, or more)
+
+NOTE: ldk-server enforces a minimum final CLTV expiry of 144 blocks on inbound keysend payments.
+If you are sending from LND or CLN, set their final CLTV delta to at least 150:
+* LND: `bitcoin.timelockdelta=150` in `lnd.conf`
+* CLN: `cltv-final=150` in `config`
 
 ## Getting Started
 
@@ -98,6 +104,18 @@ The required access details will depend on the node implementation.
   "api_password": <password_to_authorize>
 }
 ```
+* LDK-server:
+```
+{
+  "address": <ip:port or domain:port>,
+  "api_key": <hex_encoded_api_key>,
+  "cert": <path_to_tls_cert>,
+  "network": <bitcoin|testnet|signet|regtest>
+}
+```
+The `api_key` is the raw bytes of `~/.ldk-server/<network>/api_key` hex-encoded.
+Unlike other backends, ldk-server does not require an `id` field — the node's
+public key is fetched automatically on startup.
 
 Payment activity can be simulated in two different ways:
 * [Random activity](#setup---random-activity): generate random activity on the `nodes` provided, 
@@ -135,6 +153,12 @@ to send and receive payments when running with random activity.
       "base_url": "127.0.0.1:8286",
       "api_username": "",
       "api_password": "eclairpw"
+    },
+    {
+      "address": "localhost:3536",
+      "api_key": "ldk_server_hex_encoded_api_key",
+      "cert": "/path/tls.crt",
+      "network": "signet"
     }
   ]
 }
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ lightning network development. It may be useful to you if you are:
 * LND ✅
 * CLN ✅
 * Eclair ✅️
-* LDK-server ✅
+* LDK-Server ✅
 
 See our [tracking issue](https://github.com/bitcoin-dev-project/sim-ln/issues/26)
 for updates on implementation support (contributions welcome!).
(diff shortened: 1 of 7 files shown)


carlaKC:
> Before this PR can be merged, an update of used lighting libraries should be merged, as the current ldk-node has newer versions.

Feel free to bump this in a preparatory commit in this PR!

randomlogin:
Initially I was trying to wire up ldk-node, which is not a daemon and I thought sim-ln could orchestrate ldk-node building, initialization. However it turned out that we want already running and funded nodes with opened channels, so this wasn't an option. 

Added support of https://github.com/lightningdevkit/ldk-server, which is a reference daemon implementation with external API, it uses ldk-node internally. 

Please note, that ldk-server has not yet been released, therefore a git reference and a revision is used in Cargo.toml

The configuration for ldk-server looks in the following way:

```
{
  "address": <ip:port or domain:port>,
  "api_key": <hex_encoded_api_key>,
  "cert": <path_to_tls_cert>,
  "network": <bitcoin|testnet|signet|regtest>
}
```

Please note that currently it's not possible to guess the network via RPC calls, that's why we need to provide it in the config. This might change in future.

carlaKC:
> Added support of https://github.com/lightningdevkit/ldk-server, which is a reference daemon implementation with external API, it uses ldk-node internally.
Please note, that ldk-server has not yet been released, therefore a git reference and a revision is used in Cargo.toml

Happy to go with server, I believe a release is coming soon ™️ !

carlaKC:
Promise to get to review this week - sorry I've been a bit slow but it's on my list!

simln-lib/src/ldk.rs
```diff
@@ -0,0 +1,286 @@
...
+use tokio::time::{self, Duration};
+use triggered::Listener;
+
+use std::collections::HashMap;
+
+use crate::{
+    serializers, Graph, LightningError, LightningNode, NodeInfo, PaymentOutcome, PaymentResult,
+};
+
+pub struct LdkNode {
```
chuksys: I think `LdkServer` might be a better name for this struct.

simln-lib/src/ldk.rs
```diff
@@ -0,0 +1,286 @@
...
+    client: LdkServerClient,
+    info: NodeInfo,
+    network: Network,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct LdkConnection {
```
chuksys: Should we call this `LdkServerConnection`?

simln-lib/src/ldk.rs
```diff
@@ -0,0 +1,286 @@
...
+    pub address: String,
+    pub api_key: String,
+    #[serde(deserialize_with = "serializers::deserialize_path")]
+    pub cert: String,
+}
+
+impl LdkNode {
```
chuksys: Same as above

simln-lib/src/ldk.rs
```diff
@@ -0,0 +1,286 @@
...
+                features,
+                alias,
+            },
+            network,
+        })
+    }
+}
+
+#[async_trait]
+impl LightningNode for LdkNode {
```
chuksys: Same as above

simln-lib/src/ldk_server.rs
```diff
@@ -0,0 +1,286 @@
...
+        Ok(Graph { nodes_by_pk })
+    }
+}
+
+/// Convert the `types.Network` proto enum (encoded as i32) returned by
+/// ldk-server's GetNodeInfo into a `bitcoin::Network`.  
+fn network_from_proto(value: i32) -> Result<Network, LightningError> {
+    match value {
+        0 => Ok(Network::Bitcoin),
+        1 => Ok(Network::Testnet),
```
chuksys, line 257: Should we include the mapping for `Testnet4` just so it's easier to extend this in future if we start supporting `Testnet4`? For now, we could error out with a `Testnet4 not supported` message.

simln-lib/src/lib.rs
```diff
@@ -36,6 +36,7 @@ pub mod clock;
 mod defined_activity;
 pub mod eclair;
 pub mod latency_interceptor;
+pub mod ldk;
```
chuksys: `ldk_server` might be a better name?

README.md
```diff
@@ -16,7 +16,7 @@ lightning network development. It may be useful to you if you are:
 * LND ✅
 * CLN ✅
 * Eclair ✅️
-* LDK-node 🏗️
+* LDK-server ✅
```
chuksys: nit: Let's write this as `LDK-Server` with the S capitalized.

README.md
```diff
@@ -36,6 +36,7 @@ of the simulator uses keysend to execute payments, which must be enabled as foll
 * LND: `--accept-keysend`
 * CLN: enabled by default
 * Eclair: `-Declair.features.keysend=optional` (or `--features.keysend=optional` if you're using Polar)
+* LDK-server: enabled by default via `spontaneous_send`
```
chuksys: Same as above.

README.md
```diff
@@ -98,6 +99,23 @@ The required access details will depend on the node implementation.
   "api_password": <password_to_authorize>
 }
 ```
+* LDK-server:
```
chuksys: Same as above

chuksys (commented):
Thanks for working on this! I've done a quick initial code/design review. It looks good so far, but I still need to run some local tests. I'll do that ASAP and get back to you with final feedback.

simln-lib/src/ldk.rs
randomlogin: Changed

simln-lib/src/ldk.rs
randomlogin: Changed

simln-lib/src/ldk.rs
randomlogin: Changed

simln-lib/src/ldk.rs
randomlogin: Changed

simln-lib/src/ldk_server.rs
randomlogin: Changed

simln-lib/src/lib.rs
randomlogin: Changed to ldk_server

README.md
randomlogin: Capitalized

README.md
randomlogin: Capitalized

README.md
randomlogin: Capitalized

simln-lib/src/ldk_server.rs
```diff
@@ -22,22 +22,22 @@ use crate::{
     serializers, Graph, LightningError, LightningNode, NodeInfo, PaymentOutcome, PaymentResult,
 };
 
-pub struct LdkNode {
+pub struct LdkServer {
```
chuksys: To align with the naming conventions of the other impls, let's go with `LdkServerNode` here & elsewhere. Sorry for the back-and-forth!

chuksys (changes requested):

simln-lib/src/ldk_server.rs
randomlogin: Renamed. Force pushed without a new commit.

chuksys:
Just tested this using the following config:

```
{
    "nodes": [
      {
        "id": "alice",
        "address": "172.23.48.1:10002",
        "macaroon": "/mnt/c/Users/User/.polar/networks/1/volumes/lnd/alice/data/chain/bitcoin/regtest/admin.macaroon",
        "cert": "/mnt/c/Users/User/.polar/networks/1/volumes/lnd/alice/tls.cert"
      },
      {
        "id": "carol",
        "address": "172.23.48.1:10003",
        "macaroon": "/mnt/c/Users/User/.polar/networks/1/volumes/lnd/carol/data/chain/bitcoin/regtest/admin.macaroon",
        "cert": "/mnt/c/Users/User/.polar/networks/1/volumes/lnd/carol/tls.cert"      
      },
      {
        "address": "https://localhost:3536",
        "api_key": "b91adc35b0d407dab92f5b821d51e343a19497cadd0b7fd518b846c156db6df3",
        "cert": "/tmp/ldk-server/tls.crt"
      }
    ],
    "activity": [
      {
        "source": "alice",
        "destination": "carol",
        "interval_secs": 1,
        "amount_msat": 2000
      },
      {
        "source": "alice",
        "destination": "02df5ad45ae0bfc2b9fef546640ea89cd482274e92b7bb38e1d394822c8368d173",
        "interval_secs": 5,
        "amount_msat": 5000
      }
    ]
}
```

These are the logs:

sim-ln-test-with-ldk-server

sim-cli/src/parsing.rs
```diff
@@ -160,6 +160,7 @@ pub enum NodeConnection {
     Lnd(lnd::LndConnection),
     Cln(cln::ClnConnection),
     Eclair(eclair::EclairConnection),
+    Ldk(ldk_server::LdkServerConnection),
```
chuksys: This should be `LdkServer`  not `Ldk`.

sim-cli/src/parsing.rs
```diff
@@ -396,6 +397,7 @@ async fn get_clients(
             NodeConnection::Lnd(c) => Arc::new(Mutex::new(LndNode::new(c).await?)),
             NodeConnection::Cln(c) => Arc::new(Mutex::new(ClnNode::new(c).await?)),
             NodeConnection::Eclair(c) => Arc::new(Mutex::new(EclairNode::new(c).await?)),
+            NodeConnection::Ldk(c) => Arc::new(Mutex::new(LdkServerNode::new(c).await?)),
```
chuksys: This should be `LdkServer` not `Ldk`.

sim-cli/src/parsing.rs
randomlogin: Renamed

sim-cli/src/parsing.rs
randomlogin: Renamed

chuksys (approved):
If CI is happy, this LGTM.

randomlogin:
Should I ping someone, for example @carlaKC, to merge?

chuksys:
> Should I ping someone, for example @carlaKC, to merge?

Since this is a major feature, @carlaKC will need to do a final review before merging. Please give her a bit of time to take a thorough look at it!

simln-lib/src/ldk_server.rs
```diff
@@ -0,0 +1,289 @@
...
+
+        loop {
+            tokio::select! {
+                biased;
+                _ = shutdown.clone() => {
+                    return Err(LightningError::TrackPaymentError(
+                        "Shutdown before tracking results".to_string(),
+                    ));
+                },
+                _ = time::sleep(Duration::from_millis(500)) => {
```
carlaKC, line 126: We should use `SimulationClock`'s `sleep` here rather than hardcoding time.

We've got our clock behind a trait so that we can speed it up in specific simulation modes. While we never use this in practice with real nodes (like a LDK server), I think it's better to keep our abstraction in place.

simln-lib/src/ldk_server.rs
```diff
@@ -0,0 +1,289 @@
...
+            .ok_or_else(|| LightningError::GetNodeInfoError("Node not found".to_string()))?;
+
+        // ldk-server's graph RPC only surfaces the `node_announcement` fields
+        // (alias, rgb, addresses) and no feature bits, so fall back to empty.
+        let alias = node.announcement_info.map(|a| a.alias).unwrap_or_default();
+
+        Ok(NodeInfo {
+            pubkey: *node_id,
+            alias,
+            features: NodeFeatures::empty(),
```
carlaKC: Do we need to set keysend here?

I think that running this with random activity would fail - did you test this out?

README.md
```diff
@@ -98,6 +99,23 @@ The required access details will depend on the node implementation.
   "api_password": <password_to_authorize>
 }
 ```
+* LDK-Server:
```
carlaKC, line 102: Nice thanks for docs 🫶

carlaKC (commented):
So sorry that this took so long!

simln-lib/src/ldk_server.rs
randomlogin: In simulated networks we don't call any real API methods to get information about payments. And this particular usage of `sleep` is connected to real wall clock (i.e. we're bounded by the real time a node needs to process a request, not by our 'simulated time').

All other backends use the same pattern with `time::sleep`, none of the node backend implementations has a `Clock`. If you insist, I can add it to all the backends we have.

simln-lib/src/ldk_server.rs
```diff
@@ -0,0 +1,289 @@
...
+
+        let pubkey = PublicKey::from_str(&info.node_id)
+            .map_err(|err| LightningError::GetInfoError(err.to_string()))?;
+        let alias = info.node_alias.unwrap_or_default();
+        let network = network_from_proto(info.network)?;
+
+        // ldk-server doesn't expose feature bits, but it always supports keysend
+        // via `spontaneous_send`, so advertise it so sim-ln's keysend checks pass.
+        let mut features = NodeFeatures::empty();
+        features.set_keysend_optional();
```
enigbe: Feature bit exposure is now supported by `ldk-server` following https://github.com/lightningdevkit/ldk-server/pull/231, so we should be able to query the node’s actual features instead of creating an empty feature set and manually setting `keysend`.

Manually setting it assumes the node advertises `keysend` support. That matches default LDK behavior today, but it can be wrong for a fork that changes feature support/requirements, in which case `sim-ln` would incorrectly treat the node as `keysend`-capable.

simln-lib/src/ldk_server.rs
enigbe: I think this should follow the same pattern as the LND and CLN implementations: query the backend for the requested node and populate `NodeInfo.features` from the returned feature set.

Returning `NodeFeatures::empty()` here means sim-ln will treat the node as not supporting `keysend`, even if it does. Now that `ldk-server` exposes feature bits through `GetNodeInfo`, we should use that API and construct `NodeInfo` from the actual returned features instead of manual default feature support.

enigbe (changes requested):

simln-lib/src/ldk_server.rs
enigbe: Clippy is failing here because format! is only wrapping a static string, so it’s considered a useless allocation pattern. This should be:
```rust
2 => Err(LightningError::GetInfoError(
    "testnet4 network is not supported".to_string(),
)),
```
format! is only useful if we’re interpolating values into the message.

simln-lib/src/ldk_server.rs
randomlogin: Unfortunately ldk-server does not show feature bits for other nodes, it only shows them for itself. In fact there are two methods `get-node-info` about itself and `graph-get-node` which returns information about other nodes, its example response is the following:
```
{
  "node": {
    "channels": [
      3173352185970884608,
      3173346688412745728,
      3173354384994140160,
      3336995800067670017
    ],
    "announcement_info": {
      "last_update": 1782471184,
      "alias": "lnd-signet-node",
      "rgb": "ff0000",
      "addresses": []
    }
  }
}
```

I've tested random activity. it works fine. Also when we call `get_node_info` it checks whether we ask about this node, or another one. For this node we provide correct `NodeInfo` that was populated in `new()`.

simln-lib/src/ldk_server.rs
randomlogin: Fixed, now we return real feature list.

simln-lib/src/ldk_server.rs
randomlogin: Fixed

simln-lib/src/ldk_server.rs
enigbe: > Unfortunately ldk-server does not show feature bits for other nodes, it only shows them for itself.

This is surprising. I've open a PR https://github.com/lightningdevkit/ldk-server/pull/237 to address this. If merged, we should get access to feature bits from other nodes like shown below.

```json
{
  "node": {
    "channels": [
      468391953498113
    ],
    "announcement_info": {
      "last_update": 1782481522,
      "alias": "ldk_server_2",
      "rgb": "000000",
      "addresses": [
        "54.3.7.81:9745"
      ],
      "features": {
        "0": {
          "name": "DataLossProtect",
          "is_required": true
        },
        ...
      }
    }
  }
}
```

carlaKC: Not putting keysend in other nodes features will mean that the simulation will fail if we are sending payments to a node that we _don't_ have control over.

Eg for network `A - B`:
```
nodes: [A]
activity: [A -> B]
```

This is kinda niche-ey, but it is something people wanted to regularly shoot payments out to strangeres on signet, so I'd like to preserve it. Happy to hard-code if they're not surfaced by LDK-server, but if it's newly added let's use their API.

> I've tested random activity. it works fine.

(my bad, thought this hit is this case)

randomlogin: Fixed. Previously I tested random activity with all owned nodes, so I think your original comment is valid.

simln-lib/src/ldk_server.rs
carlaKC: Fosho, was more a comment on consistency but if we have it on the other backends let's leave as-is.

carlaKC (approved):
Typically would want to fixup and squash things like clippy fixes and node feature setting, but I left this hanging unacceptably long so happy to go in as-is. Will merge once CI is clean - thanks for the patience here and the cool new feature!

randomlogin:
I'll squash commits.

carlaKC:
> I'll squash commits.

Ready and waiting with the merge button 🫡

carlaKC (approved):

PR #309, opened 2026-06-23
Use time-paused tokio for zero sleeps

carlaKC:
Our current solution to waits is to use a sped up wall clock, which has all sorts of limitations.

Using a paused, single-threaded tokio runtime we can get discrete-event-like functionality without doing a large refactor.

### DIFF
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -22,7 +22,7 @@ async fn main() -> anyhow::Result<()> {
     }
 
     let cli = Cli::parse();
-    let sim_params = parse_sim_params(&cli).await?;
+    let sim_params = parse_sim_params(&cli)?;
 
     SimpleLogger::new()
         .with_level(LevelFilter::Warn)
--- a/sim-cli/src/main.rs
+++ b/sim-cli/src/main.rs
@@ -4,17 +4,16 @@ use std::time::SystemTime;
 use clap::Parser;
 use log::LevelFilter;
 use sim_cli::parsing::{create_simulation, create_simulation_with_network, parse_sim_params, Cli};
+use simln_lib::latency_interceptor::LatencyIntercepor;
+use simln_lib::sim_node::Interceptor;
 use simln_lib::{
-    clock::SimulationClock,
-    latency_interceptor::LatencyIntercepor,
-    sim_node::{CustomRecords, Interceptor},
-    SimulationCfg,
+    clock::SimulationClock, runtime::block_on_virtual_time, sim_node::CustomRecords,
+    ActivityDefinition, Simulation, SimulationCfg,
 };
 use simple_logger::SimpleLogger;
 use tokio_util::task::TaskTracker;
 
-#[tokio::main]
-async fn main() -> anyhow::Result<()> {
+fn main() -> anyhow::Result<()> {
     // Enable tracing if building in developer mode.
     #[cfg(feature = "dev")]
     {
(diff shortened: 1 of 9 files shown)


elnosh (commented):
ran a couple of simulations and it is indeed faster - a 6-month simulation that previously took hours now ran in 27 mins. However, I did run into an issue explained here https://github.com/carlaKC/jam-ln/pull/133#pullrequestreview-4612400628 which is that any polling loop could cause issues if the comments on `clock.rs` are accurate 

```
If the system never reaches the state where all tasks are waiting, the clock will not
advance. This may happen if there is an always-pollable loop, for example.
```

simln-lib/src/sim_node.rs
```diff
@@ -1164,9 +1164,8 @@ pub fn populate_network_graph<C: Clock>(
         };
 
         graph.update_channel_from_unsigned_announcement(&announcement, &Some(&utxo_validator))?;
-
-        // LDK only allows channel announcements up to 24h in the future. Use a fixed timestamp so that even if we've
-        // sped up our clock dramatically, we won't hit that limit.
+        // LDK only allows channel announcements up to 24h in the future. Use a fixed timestamp so that even if virtual
+        // time has advanced dramatically, we won't hit that limit.
```
elnosh: if I'm understanding the behavior of `SimulationClock` correctly, then this comment is inaccurate (?) Line below calls `clocks.now()` which if running with virtual time runtime then this could return something far in the future. I guess it works fine because this method is called at the beginning when setting up the network so virtual time prob has not advanced much but still comment seems false.

simln-lib/Cargo.toml
```diff
@@ -34,8 +34,17 @@ rand_chacha = "0.3.1"
...
+# in tests, so it is gated behind an opt-in feature rather than always enabled — consumers that don't run virtual-time
+# simulations don't pull `test-util` into their production build.
+virtual-time = ["tokio/test-util"]
+
 [dev-dependencies]
 ntest = "0.9.0"
 mockall = "0.13.1"
 futures = "0.3.31"
 tempfile = "3"
+# Features used only by the test suite: macros for #[tokio::test] and rt-multi-thread for multi-threaded test runtimes.
```
elnosh: nit: i don't love these self-explanatory claude comments. Feel free to ignore tho

sim-cli/src/parsing.rs
```diff
@@ -518,7 +518,7 @@ async fn validate_activities(
     Ok(validated_activities)
 }
 
-async fn read_sim_path(data_dir: PathBuf, sim_file: PathBuf) -> anyhow::Result<PathBuf> {
```
elnosh, line 534: cool

sim-cli/src/main.rs
```diff
@@ -33,35 +32,52 @@ async fn main() -> anyhow::Result<()> {
...
-
-    let (sim, validated_activities) = if sim_params.sim_network.is_empty() {
-        create_simulation(&cli, &sim_params, tasks.clone()).await?
+    let cfg = SimulationCfg::try_from(&cli)?;
+    let latency = cli.latency_ms.unwrap_or(0);
+    let interceptors = if latency > 0 {
+        vec![Arc::new(LatencyIntercepor::new_poisson(
+            latency as f32,
+            cli.fix_seed,
+        )?) as Arc<dyn Interceptor>]
```
elnosh: this could be in the branch where we call `create_simulation_with_network` since we only use it there.

sim-cli/src/parsing.rs
```diff
@@ -114,6 +118,19 @@ impl Cli {
             ));
         }
 
+        if self.virtual_time {
+            if !sim_params.nodes.is_empty() {
+                return Err(anyhow!(
+                    "Virtual time is only allowed when running on a simulated network; real nodes run on wall time"
+                ));
+            }
+            if self.total_time.is_none() {
```
elnosh, line 127: nit: can fix this comment https://github.com/bitcoin-dev-project/sim-ln/blob/96c8f6dc3e8bf9d54e60f56f344f1e8e2a11fd00/simln-lib/src/lib.rs#L567

elnosh (commented):
Discussed offline and runtime introduced here doesn't seem to be the culprit for that specific simulation taking longer.

LGTM - some small nits

simln-lib/Cargo.toml
carlaKC: Fair, I was on the fence about them - will remove.

simln-lib/src/sim_node.rs
carlaKC: Yeah comment needs to go. 

> I guess it works fine because this method is called at the beginning when setting up the network so virtual time prob has not advanced much

We don't have any sleeps between clock creation and calling this, so the clock will be exactly at the wall clock time we started it with.

carlaKC:
Sorry for the slop! [here](https://github.com/bitcoin-dev-project/sim-ln/compare/96c8f6dc3e8bf9d54e60f56f344f1e8e2a11fd00..728393e64325a15a6b7e6063cd9950ea22b06b13) and [here](https://github.com/bitcoin-dev-project/sim-ln/compare/8283ba5473cecaa96398fe57efe1b82a293253c2..6d9a610b56d77638873bb28151a436f9de04e611)

elnosh (approved):

PR #314, opened 2026-08-04
Add layered integration test framework with simulated and real-node tiers

carlaKC:
Adds an integration test framework covering sim-ln's node backends, payment modalities and configuration surface, structured as independent layers:

- **Environment layer** (`integration-tests/src/env/`): provisions a network and emits its part of the sim.json config. Two providers: a deterministic in-process simulated graph, and a dockerized heterogeneous regtest network (bitcoind + one node each of LND, CLN, Eclair and ldk-server in a ring of announced channels, via testcontainers). Knows nothing about payments.
- **Scenario layer** (`src/scenario.rs`): describes activity (defined/random) and config style (alias vs pubkey references, scalar vs `[min, max]` values). Knows nothing about provisioning.
- **Runner + assertions** (`src/runner.rs`, `src/asserts.rs`): assembles real config files, drives them through the same public entry points the sim-cli binary uses, and asserts over parsing results, dispatched payments and the results CSV.

### Test tiers

**Simulated** (`tests/sim_matrix.rs`, runs in the existing test CI job, ~0.2s on virtual time): the full cross-product of node reference styles and value shapes for defined activity, random activity with/without exclusions, seeded determinism, count- and time-bounded runs, untagged connector inference, and negative validation cases asserting specific errors.

**Real nodes** (`tests/real_nodes.rs`, `#[ignore]`, new CI job / `make integration-real`, ~8 min): one shared network, scenarios run sequentially — defined keysends across every directed ring edge (each connector proves it sends *and* receives cross-implementation, with receipts checked against each destination node's own books), a multi-hop route, alias-referenced activities, and random activity.

### Error resistance

Every real-node startup step polls with capped exponential backoff and names the node/step in its failure. Container logs are dumped on startup or scenario failure. Readiness keeps mining while waiting for channels because some implementations (ldk-node) broadcast funding transactions asynchronously after the open call returns.

### Notes

- ldk-server has no published image: the harness (and CI, with a docker layer cache) builds one from the upstream repo at the rev the `ldk-server-client` dependency already pins; `SIMLN_LDK_SERVER_IMAGE` overrides.
- ACINQ ships no versioned eclair tags past 0.8.0, so eclair is pinned by digest; the image is built from the post-release dev commit and needs `-Declair.allow-unsafe-startup=true` (a static dev-build guard, nothing dynamic).
- `testcontainers` is pinned to `=0.25.0`, the last release whose MSRV fits the repo's Rust 1.85 toolchain; `serde_with`/`time` are pinned in Cargo.lock for the same reason.
- Count-bounded assertions tolerate the loss of the final payment record per activity: meeting a payment count triggers shutdown in the same instant as the last dispatch, and the results consumer's `select!` prefers the shutdown listener over draining pending results. Possibly worth an upstream fix (drain the channel before exiting) — happy to file separately.

Both tiers run green locally on macOS (Apple Silicon, Docker Desktop; eclair runs under Rosetta since its image is amd64-only).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/integration-tests/src/asserts.rs
+++ b/integration-tests/src/asserts.rs
@@ -1,0 +1,166 @@
+//! Shared assertions over simulation output, independent of which environment produced it.
+
+use bitcoin::secp256k1::PublicKey;
+use std::collections::HashMap;
+
+use crate::env::NodeHandle;
+use crate::runner::SimOutput;
+use crate::scenario::ValueSpec;
+
+/// Asserts that the validated activities resolved to the expected source/destination handles, in
+/// order. This checks alias/pubkey resolution end to end: however the config referenced the node,
+/// validation must land on the right pubkey.
+pub fn assert_activities_resolved(out: &SimOutput, expected: &[(&NodeHandle, &NodeHandle)]) {
+    assert_eq!(
+        out.activities.len(),
+        expected.len(),
+        "expected {} validated activities, got {}",
+        expected.len(),
+        out.activities.len()
+    );
+
+    for (activity, (source, dest)) in out.activities.iter().zip(expected) {
+        assert_eq!(
+            activity.source.pubkey, source.pubkey,
+            "activity source resolved to {} instead of {} ({})",
+            activity.source.pubkey, source.pubkey, source.alias
+        );
+        assert_eq!(
+            activity.destination.pubkey, dest.pubkey,
+            "activity destination resolved to {} instead of {} ({})",
+            activity.destination.pubkey, dest.pubkey, dest.alias
+        );
+    }
+}
+
+/// Asserts that exactly `expected` payments were dispatched, and that the results CSV recorded
+/// every one of them.
+pub fn assert_total_payments(out: &SimOutput, expected: u64) {
+    assert_eq!(
+        out.total_payments, expected,
+        "expected {expected} payments, simulation reported {}",
+        out.total_payments
+    );
+    assert_eq!(
+        out.records.len() as u64,
+        expected,
+        "expected {expected} payment records in results CSV, found {}",
+        out.records.len()
+    );
+}
+
+/// Asserts that at least one payment was dispatched and that the CSV agrees with the simulation's
+/// own count. Used for random activity, where exact counts depend on the generator.
+pub fn assert_payments_dispatched(out: &SimOutput) {
+    assert!(
+        out.total_payments > 0,
+        "expected the simulation to dispatch payments, got none"
+    );
+    assert_eq!(
+        out.records.len() as u64,
+        out.total_payments,
+        "results CSV has {} records but simulation reported {} payments",
+        out.records.len(),
+        out.total_payments
+    );
+}
+
+/// Asserts that every recorded payment succeeded.
+pub fn assert_all_success(out: &SimOutput) {
+    let failures: Vec<_> = out.records.iter().filter(|r| !r.is_success()).collect();
+    assert!(
+        failures.is_empty(),
+        "expected all payments to succeed, {} failed: {failures:?} (success rate {:.2}%)",
+        failures.len(),
+        out.success_rate
+    );
+}
+
+/// Asserts that every recorded payment amount could have been produced by the given spec.
+pub fn assert_amounts_within(out: &SimOutput, spec: ValueSpec) {
+    for record in &out.records {
+        assert!(
+            spec.contains(record.amount_msat),
+            "payment of {} msat outside configured amount {spec:?}",
+            record.amount_msat
+        );
+    }
+}
+
+/// Asserts the results of count-bounded defined activity: payments flow only between the
+/// configured pairs, every recorded payment succeeded, and each pair recorded either `count` or
+/// `count - 1` payments.
+///
+/// The final payment of a run is allowed to be missing because meeting a payment count shuts the
+/// simulation down in the same instant as the last dispatch, and the results consumer prefers the
+/// shutdown signal over draining pending results — so the last payment's record can be dropped.
+pub fn assert_defined_payments(out: &SimOutput, pairs: &[(&NodeHandle, &NodeHandle)], count: u64) {
+    assert_all_success(out);
+    let counts = assert_payments_between(out, pairs);
+
+    for (source, dest) in pairs {
+        let recorded = counts
+            .get(&(source.pubkey, dest.pubkey))
+            .copied()
+            .unwrap_or(0);
+        assert!(
+            (count - 1..=count).contains(&recorded),
+            "expected {count} (or {} with the trailing record lost to shutdown) payments from {} \
+             to {}, recorded {recorded}",
+            count - 1,
+            source.alias,
+            dest.alias
+        );
+    }
+}
+
+/// Asserts that recorded payments flow only between the expected (source, destination) pairs, and
+/// returns the per-pair counts for further assertions.
+pub fn assert_payments_between(
+    out: &SimOutput,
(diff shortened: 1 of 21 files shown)


integration-tests/src/env/containers/cln.rs
```diff
@@ -0,0 +1,287 @@
...
+            .ok_or_else(|| anyhow!("cln newaddr returned no bech32 address"))
+    }
+
+    pub async fn open_channel(
+        &self,
+        peer: &PublicKey,
+        peer_host: &str,
+        peer_port: u16,
+    ) -> anyhow::Result<()> {
+        with_backoff("cln open channel", Backoff::slow(), || async {
```
only1dreamgene, line 204: `open_channel` wraps the whole `connect_peer` + `fund_channel` sequence in `with_backoff`, but channel-open RPCs aren't idempotent. If the response is lost or times out after the node already accepted the request (funding tx broadcast), a retry issues a second channel-open to the same peer. Same pattern in `lnd.rs`'s `open_channel_sync`, `eclair.rs`'s `open`, and `ldk_server.rs`'s `open_channel`. Might be worth splitting the connect step (safe to retry) from the actual open call, or checking for an existing/pending channel before retrying.

integration-tests/src/asserts.rs
```diff
@@ -0,0 +1,166 @@
...
+    assert_all_success(out);
+    let counts = assert_payments_between(out, pairs);
+
+    for (source, dest) in pairs {
+        let recorded = counts
+            .get(&(source.pubkey, dest.pubkey))
+            .copied()
+            .unwrap_or(0);
+        assert!(
+            (count - 1..=count).contains(&recorded),
```
only1dreamgene, line 107: `count - 1` underflows if `count == 0` (u64). Not hit by current call sites, but a future caller asserting zero payments between excluded pairs would panic here instead of getting a clear assertion failure. `count.saturating_sub(1)` would be safer.

integration-tests/tests/real_nodes.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    let nodes = network.nodes();
+    let pairs: Vec<_> = RING.iter().map(|(s, d)| (&nodes[*s], &nodes[*d])).collect();
+    assert_activities_resolved(&out, &pairs);
+    assert_defined_payments(&out, &pairs, count);
+
+    // Each ring destination's own node reports the keysends it received. The final payment of
+    // the run may still be settling (or its record lost to shutdown), hence the range.
+    for (_, dest) in RING {
+        let received = network.settled_keysend_count(dest).await? - receipts_before[dest];
```
only1dreamgene, line 145: `settled_keysend_count(dest).await? - receipts_before[dest]` can underflow-panic if the post-scenario count is ever lower than the pre-scenario snapshot (e.g. a paginated `list_payments` race, or a restart mid-test). A `saturating_sub` plus a descriptive message would fail more usefully than a raw overflow panic.

integration-tests/src/retry.rs
```diff
@@ -0,0 +1,117 @@
...
+    Fut: Future<Output = Result<T, E>>,
+{
+    let start = Instant::now();
+    let mut delay = backoff.initial;
+
+    loop {
+        match op().await {
+            Ok(t) => return Ok(t),
+            Err(e) => {
+                if start.elapsed() + delay > backoff.timeout {
```
only1dreamgene, line 63: This check aborts before making the next attempt whenever `elapsed + delay > timeout`, even when the retry itself would still land within budget (e.g. elapsed=59s of a 60s timeout with `delay=5s` — aborts here even though the next poll would happen well before 60s). Could produce flaky 'not ready' failures for nodes that would've answered on the very next poll. Might be worth checking `elapsed >= timeout` after attempting, or clamping the final delay to the remaining budget instead of bailing early.

integration-tests/src/env/containers/ldk_server.rs
```diff
@@ -0,0 +1,317 @@
...
+
+use super::{
+    dump_logs, exec_stdout, extract_file, BTC_RPC_PASS, BTC_RPC_USER, CHANNEL_SIZE_SAT, P2P_PORT,
+};
+use crate::env::{NodeHandle, NodeImpl};
+use crate::retry::{with_backoff, Backoff};
+
+/// The rev the workspace's ldk-server-client dependency pins; the server is built from the same
+/// rev so client and server always match.
+const LDK_SERVER_REV: &str = "8163f4fe139368613959bf4f10b19ee6a5b9b4ab";
```
only1dreamgene, line 32: This is a second hardcoded copy of the ldk-server rev, separate from the git rev pinned for `ldk-server-client` in `integration-tests/Cargo.toml`. The CI workflow already derives its rev by grepping `Cargo.lock` rather than hardcoding it again — might be worth doing the same here so a future dependency bump can't silently build a server image at a different commit than the client expects.

integration-tests/src/env/containers/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// credentials (certs, macaroons) are written under `creds_dir`, which must outlive the
+    /// simulations run against this network.
+    pub async fn start(creds_dir: &Path) -> anyhow::Result<Self> {
+        let docker_network = format!("simln-itest-{}", std::process::id());
+
+        let bitcoind = bitcoind::Bitcoind::start(&docker_network).await?;
+
+        // The four nodes are independent of each other until channels open: start them in
+        // parallel since (particularly with cold image pulls) startup dominates runtime.
+        let (lnd, cln, eclair, ldk) = tokio::try_join!(
```
only1dreamgene, line 128: If one of these four futures errors while the others are still mid-initialization (containers already up, credentials being extracted), `try_join!` drops the still-running futures immediately. Cleanup then relies entirely on `Drop`/the testcontainers reaper rather than each harness's own `dump_logs`/error-context path, so a fast-failing node can leave the still-starting ones without a diagnostic log dump. A `join_all` with explicit error aggregation would let every harness reach its own failure/cleanup path.

integration-tests/tests/real_nodes.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/// One defined activity per directed ring edge: proves every connector both dispatches and
+/// receives keysend cross-implementation, with receipt confirmed against each destination node's
+/// own books rather than sim-ln's records alone.
+async fn scenario_ring_defined(network: &RealNetwork) -> anyhow::Result<()> {
+    let count = 3u64;
+
+    let mut receipts_before = [0u64; 4];
+    for (i, receipts) in receipts_before.iter_mut().enumerate() {
```
only1dreamgene, line 104: These four `settled_keysend_count` calls are awaited sequentially. They're independent RPCs to four different nodes (and Eclair's JVM-under-emulation responses are already the slow one) — a `tokio::try_join!` here would cut this baseline-read step down to the slowest single call instead of the sum of all four.

only1dreamgene (commented):
A few notes from going through the new integration-tests crate: two unchecked u64 subtractions that can panic on edge-case inputs, a non-idempotent retry pattern around channel-open RPCs that could double-open channels on a lost response, and some smaller flakiness/efficiency items inline.

carlaKC:
Thanks for the review @only1dreamgene ! 

This is an entirely-clauded idea that I haven't even read the code for, just thought it would be a good idea to throw up and deal with ✨ sometime ✨ but I never got around to it.

Would you be interested in taking over this PR and addressing your own review? It's definitely something I'd like to get over the finish line.

PR #315, opened 2026-08-12
simln-lib: enforce HTLC policy limits against the correct party

elnosh:
`check_outgoing_addition` previously checked every limit against the sending node's own policy. Split each limit to the party that owns it:

* htlc_minimum_msat / htlc_maximum_msat are forwarder-advertised fields of the sender's own gossiped channel_update (populated into the graph from `self.policy`), so they are enforced against `self.policy`.
* max_accepted_htlcs / max_htlc_value_in_flight_msat are receiver- negotiated (BOLT-2 channel open, never gossiped), so they are enforced against the counterparty's policy. Keeping this at add time means the final recipient's inbound limits are enforced too, since it never adds an outgoing HTLC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,7 +304,7 @@ If you want to run the cli in a containerized environment, see the docker set up
 
 ## Advanced Usage - Network Simulation
 
-If you are looking to simulate payments are large lightning networks 
+If you are looking to simulate payments on large lightning networks 
 without the resource consumption of setting up a large cluster of nodes,
 you may be interested in dispatching payments on a simulated network.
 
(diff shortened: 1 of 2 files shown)


carlaKC (commented):
Makes sense. Could you update the readme's Advanced Usage section to make note of how these will be used?

elnosh:
> Could you update the readme's Advanced Usage section to make note of how these will be used?

done

carlaKC:
thanks!

carlaKC (approved):

PR #316, opened 2026-08-31
Use LDK Server node features

### LINKED ISSUE #310: Followup: Lookup real features for LDK-server
**Describe the followup**

From #307, blocked on ldk-server change.

We [currently hardcode](https://github.com/bitcoin-dev-project/sim-ln/pull/307/changes#diff-55302799864abb244abbbeb3ef2bbc3d9155ce8b7102bae46a4782e3e0d53a95R182) nodes features in LDK-server because they're not surfaced. Once https://github.com/lightningdevkit/ldk-server/pull/237 is in, we can update sim-ln to use the actual node's features.

Really shouldn't be an issue unless somebody turns off keysend, but good to be correct.

enigbe:
## What this PR does
`LdkServerNode ` assumed every remote graph node supports `keysend` because LDK Server node announcement features were not previously exposed by its API. This allowed nodes that do not advertise `keysend` support to pass activity validation, resulting in avoidable payment failures.

Here we:
- Update the `ldk-server-client` revision to include graph node announcement features.
- Use remote nodes’ advertised features instead of assuming `keysend` support. Nodes without announcement information are treated as having no advertised features.

Closes #310


carlaKC (approved):
Thanks for the PR!

PR #319, opened 2026-09-15
Parallel pathfinding and other things

elnosh:
I was running with a simulated network of ~26k channels and it was very slow. From looking at perf results it seems majority of time is spent on pathfinding. Not that pathfinding is slow but in a network with ~26k channels and default simulation values (send multiplier), the simulation will generate *a lot* of payments making majority of time spent on pathfinding. The payment volume increases a lot because each node is set to send some multiplier of its capacity per certain period. With high channel count, this will spike the payment generation.

Some things done here:
- I didn't know this but even on a `current_thread` runtime, `spawn_blocking` will spawn additional threads for blocking operations. So use that to parallelize pathfinding.
- Some mutex changes and unnecessary locks taken for the scorer and `LightningNode`.
- Bumped ldk to `0.2`. We were on `0.0.123` which was a bit old. Pathfinder is a bit faster since then so bumped that dep. 

And just reducing the multiplier to generate less payments helped.


carlaKC (approved):
Nice clean up! Haven't looked deeply into the paused runtime workings but will take your/claude's word for it!

PR #320, opened 2026-09-16
payment retries on simulated network

elnosh:
This adds a payment retries when running in a simulated network. It just sets it to at max try 5.

Note that there was a bug in the reporting to the scorer. While re-working the outcomes to also report the channels where a route failed, (claude) noticed that [this](https://github.com/bitcoin-dev-project/sim-ln/blob/82da8d83b3fdf6e9074d2e890c208f0ff6f07d2a/simln-lib/src/sim_node.rs#L790) was reporting the failure to the scorer with just an index casted as a u64 where LDK is expecting a `scid`. This is fixed in https://github.com/bitcoin-dev-project/sim-ln/commit/85b213dfb3bfb256212ebd8bc8ae20641efd1342

The retries approach just takes the channel at which the payment failed and passes it to LDK to ignore when calculating the route for the next retry


elnosh:
needed a rebase after https://github.com/bitcoin-dev-project/sim-ln/pull/319

